PR #1, opened 2014-06-30
[IMP] Add class CarrierAccount, so the specifics carriers modules use al...

florian-dacosta:
Add a new class to configure the specific Carrier Accounts.
The purpose of this class is to be inherited by specific carrier modules so they add their specificities.

You can find an example of use here (the module is still work in progress)

http://bazaar.launchpad.net/~florian-dacosta/+junk/delivery_carrier_chronopost/view/head:/delivery_carrier_chronopost/config.py

Original proposal on Launchpad : https://code.launchpad.net/~akretion-team/carriers-deliveries/7-add-base-class-for-carrier-configs/+merge/224598

### DIFF
--- a/base_delivery_carrier_label/delivery.py
+++ b/base_delivery_carrier_label/delivery.py
@@ -96,3 +96,41 @@ class DeliveryCarrier(orm.Model):
             'delivery.carrier.option',
             'carrier_id', 'Option'),
     }
+
+
+class CarrierAccount(orm.Model):
+    _name = 'carrier.account'
+    _description = 'Base account datas'
+
+    def _get_carrier_type(self, cr, uid, context=None):
+        """ To inherit to add carrier type like Chronopost, Postlogistics..."""
+        return []
+
+    def __get_carrier_type(self, cr, uid, context=None):
+        """ Wrapper to preserve inheritance for selection field """
+        return self._get_carrier_type(cr, uid, context=context)
+
+    def _get_file_format(self, cr, uid, context=None):
+        """ To inherit to add label file types"""
+        return [('PDF', 'PDF'),
+                ('SPD', 'SPD'),
+                ('PPR', 'PPR'),
+                ('THE', 'THE'),
+                ('ZPL', 'ZPL'),
+                ('XML', 'XML')]
+
+    def __get_file_format(self, cr, uid, context=None):
+        """ Wrapper to preserve inheritance for selection field """
+        return self._get_file_format(cr, uid, context=context)
+
+    _columns = {
+        'name': fields.char('Name', size=64, required=True),
+        'account': fields.char('Account Number', size=32, required=True),
+        'password': fields.char('Account Password', size=32, required=True),
+        'file_format': fields.selection(__get_file_format, 'File Format',
+            help="Default format of the carrier's label you want to print"),
+        'type': fields.selection(__get_carrier_type, 'Type', required=True,
+            help="In case of several carriers, help to know which account belong to which carrier"),
+    }
+
+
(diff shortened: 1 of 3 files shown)


bealdav:
Approved

guewen:
Is it something generic? On what criteria do you think that?

base_delivery_carrier_label/delivery.py
```diff
@@ -96,3 +96,41 @@ def __get_carrier_type_selection(self, cr, uid, context=None):
...
+                ('ZPL', 'ZPL'),
+                ('XML', 'XML')]
+
+    def __get_file_format(self, cr, uid, context=None):
+        """ Wrapper to preserve inheritance for selection field """
+        return self._get_file_format(cr, uid, context=context)
+
+    _columns = {
+        'name': fields.char('Name', size=64, required=True),
+        'account': fields.char('Account Number', size=32, required=True),
```
guewen: No reason to enforce a size.

base_delivery_carrier_label/delivery.py
```diff
@@ -96,3 +96,41 @@ def __get_carrier_type_selection(self, cr, uid, context=None):
...
+        'password': fields.char('Account Password', size=32, required=True),
```
guewen: No reason to enforce a size.

base_delivery_carrier_label/delivery.py
```diff
@@ -96,3 +96,41 @@ def __get_carrier_type_selection(self, cr, uid, context=None):
...
+                ('THE', 'THE'),
+                ('ZPL', 'ZPL'),
+                ('XML', 'XML')]
+
+    def __get_file_format(self, cr, uid, context=None):
+        """ Wrapper to preserve inheritance for selection field """
+        return self._get_file_format(cr, uid, context=context)
+
+    _columns = {
+        'name': fields.char('Name', size=64, required=True),
```
guewen: No reason to enforce a size.

base_delivery_carrier_label/delivery.py
```diff
@@ -96,3 +96,41 @@ def __get_carrier_type_selection(self, cr, uid, context=None):
...
+        'account': fields.char('Account Number', size=32, required=True),
+        'password': fields.char('Account Password', size=32, required=True),
+        'file_format': fields.selection(__get_file_format, 'File Format',
+            help="Default format of the carrier's label you want to print"),
+        'type': fields.selection(__get_carrier_type, 'Type', required=True,
+            help="In case of several carriers, help to know which account belong to which carrier"),
```
guewen: Wrong indentation according to pep8.

florian-dacosta:
Yes I guess it is generic as, for now, in all specific carrier modules (postlogistic, chronopost, laposte, GLS) we need one or multiple account. These fields are always needed.

In the specific carrier module, you create a new class which inherits this one, to add new specific fields, and it also allow you to handle the notion of subaccount. (needed in chronopost) or the idea of licenses (needed in postlogistic)

florian-dacosta:
The last commit is to change a little the behavior of _get_label_sender_address method.
The address of the company can be different for invoice or for delivery. So instead of getting the address of the company to put on the label, we take the address of a contact of the company which has type = delivery if it exists.

bealdav:
Hi Guewen and Yannick

1/ It's a way to have multiple sub-account/license by company. We need it for Chronopost and may be in next projects for other carriers. Common fields are in carrier.account. Data model with inherits allows to have a full flexibility for the complex cases. For Laposte I only will use 'carrier.account' model.
2/ All the formats in _get_file_format are not generic. Some of them are specifics. I think It should be sufficient to have pdf and zpl for a generic module (it will cover 80% of the cases). If more extend in your own module.

What do you think ? specially Yannick ?

I have a project in production next week it could be nice if it was decided these days

Thanks

bealdav:
Hi Florian, 

Do you intend to put this code on master or on 7.0 ?
For now it only on master

yvaucher:
Please restart your proposal by doing a rebase (we want travis to do a check) and targeting 7.0

PR #2, opened 2014-06-30
[PORT] prepare port to 8.0 by disabling all module

yvaucher:


pedrobaeza:
The same problem of the other PR on the approach. Sorry for the incoveniences.

PR #8, opened 2014-07-04
7.0 - Coverage use png badge so it can specify branch

bwrsandman:


yvaucher:
thats just what I was looking for. :+1:

PR #10, opened 2014-07-06
[IMP] split state field in two boolean fields 

bealdav:
with the same feature in base delivery_carrier_label

validated on https://code.launchpad.net/~akretion-team/carriers-deliveries/7-split-default-option-state-from-deliv-method-dbl/+merge/224027

### DIFF
--- a/__unported__/base_delivery_carrier_label/__openerp__.py
+++ b/__unported__/base_delivery_carrier_label/__openerp__.py
@@ -19,7 +19,7 @@
 #
 ##############################################################################
 {'name': 'Base module for carrier labels',
- 'version': '1.1',
+ 'version': '1.2',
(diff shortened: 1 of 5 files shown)


pedrobaeza:
Hi, David, some remarks:
- You're making PR against master branch (future 8.0), so migration script is not going to work because migration folder is 7.0.1.2. Please rename it to 1.2 to be valid in any major version.
- As this is the master branch, maybe it's better to make also the PR against 7.0 branch, and let the migration script only there, so migration path would be: 7.0.1.1 > 7.0.1.2 > 8.0.1.2.

Regards.

bealdav:
damned, you're right

sorry

bealdav:
close because error

PR #11, opened 2014-07-07
[IMP] split state field in two boolean fields with the same feature in b...

bealdav:
with the same feature in base delivery_carrier_label

validated on https://code.launchpad.net/~akretion-team/carriers-deliveries/7-split-default-option-state-from-deliv-method-dbl/+merge/224027

this PR replace the previous closed

### DIFF
--- a/base_delivery_carrier_label/delivery.py
+++ b/base_delivery_carrier_label/delivery.py
@@ -32,12 +32,15 @@ class DeliveryCarrierTemplateOption(orm.Model):
             'Partner Carrier'),
         'name': fields.char(
             'Name',
+            readonly=True,
             size=64),
         'code': fields.char(
             'Code',
+            readonly=True,
             size=64),
         'description': fields.char(
             'Description',
+            readonly=True,
             help="Allow to define a more complete description "
                  "than in the name field."),
     }
@@ -55,21 +58,26 @@ class DeliveryCarrierOption(orm.Model):
     _inherits = {'delivery.carrier.template.option': 'tmpl_option_id'}
 
     _columns = {
-        'state': fields.selection(
-            (('mandatory', 'Mandatory'),
-             ('default_option', 'Optional by Default'),
-             ('option', 'Optional'),
-             ),
-            string='Option Configuration',
-            help="Ensure you add and define correctly all your options or those won't "
-                 "be available for the packager\n"
-                 "- Mandatory: This option will be copied on carrier and cannot be removed\n"
-                 "- Optional by Default: This option will be copied but can be removed\n"
-                 "- Optional: This option can be added later by the user on the Delivery Order."),
+        'mandatory': fields.boolean(
+            'Mandatory',
+            help="If checked, this option is necessarily applied to the picking"),
+        'by_default': fields.boolean(
+            'Applied by Default',
+            help="By check, user can choose to apply this option "
+            "to each pickings\n using this delivery method"),
         'tmpl_option_id': fields.many2one(
             'delivery.carrier.template.option',
             string='Option', required=True, ondelete="cascade"),
         'carrier_id': fields.many2one('delivery.carrier', 'Carrier'),
+        'readonly_flag': fields.boolean(
+            'Readonly Flag',
+            help="When True, help to prevent the user to modify some fields "
+                 "option (if attribute is defined in the view)"),
+    }
+
+    _defaults = {
+        'readonly_flag': False,
+        'by_default': False,
     }
 
 
--- a/base_delivery_carrier_label/stock.py
+++ b/base_delivery_carrier_label/stock.py
@@ -163,7 +163,8 @@ class stock_picking(orm.Model):
             available_option_ids = []
             for available_option in carrier.available_option_ids:
                 available_option_ids.append(available_option.id)
-                if available_option.state in ['default_option', 'mandatory']:
+                if available_option.mandatory \
+                        or available_option.by_default:
                     default_option_ids.append(available_option.id)
             res = {
                 'value': {'carrier_type': carrier.type,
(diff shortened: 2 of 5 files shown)


yvaucher:
:+1: 
Can you just do a rebase ? Travis failed greenly as it was done on a wrong config

bealdav:
rebase done thanks to merge

pedrobaeza:
:+1:

base_delivery_carrier_label/stock.py
```diff
@@ -163,7 +163,8 @@ def carrier_id_change(self, cr, uid, ids, carrier_id, context=None):
             available_option_ids = []
             for available_option in carrier.available_option_ids:
                 available_option_ids.append(available_option.id)
-                if available_option.state in ['default_option', 'mandatory']:
+                if available_option.mandatory \
+                        or available_option.by_default:
```
guewen, line 167: Prefer parenthesis over \, and the `or` or `and` operators should be at the end of the line:

```
    if (available_option.mandatory or
            available_option.by_default):
        ...
```

base_delivery_carrier_label/delivery.py
```diff
@@ -55,21 +58,26 @@ class DeliveryCarrierOption(orm.Model):
...
-                 "- Mandatory: This option will be copied on carrier and cannot be removed\n"
-                 "- Optional by Default: This option will be copied but can be removed\n"
-                 "- Optional: This option can be added later by the user on the Delivery Order."),
+        'mandatory': fields.boolean(
+            'Mandatory',
+            help="If checked, this option is necessarily applied to the picking"),
+        'by_default': fields.boolean(
+            'Applied by Default',
+            help="By check, user can choose to apply this option "
+            "to each pickings\n using this delivery method"),
```
guewen, line 67: "pickings" is not used for users, prefer "Delivery Orders".

base_delivery_carrier_label/delivery.py
```diff
@@ -55,21 +58,26 @@ class DeliveryCarrierOption(orm.Model):
...
-             ),
-            string='Option Configuration',
-            help="Ensure you add and define correctly all your options or those won't "
-                 "be available for the packager\n"
-                 "- Mandatory: This option will be copied on carrier and cannot be removed\n"
-                 "- Optional by Default: This option will be copied but can be removed\n"
-                 "- Optional: This option can be added later by the user on the Delivery Order."),
+        'mandatory': fields.boolean(
+            'Mandatory',
+            help="If checked, this option is necessarily applied to the picking"),
```
guewen, line 63: "pickings" is not used for users, prefer "Delivery Orders".

guewen:
Needs a rebase.

base_delivery_carrier_label/delivery.py
gurneyalex, line 67: and "each" demands an singular

base_delivery_carrier_label/stock.py
yvaucher, line 167: Operator position is a matter of taste here it can still be on start of second line

guewen, line 167: On 08/25/2014 08:57 AM, Yannick Vaucher wrote:

> In base_delivery_carrier_label/stock.py:
> 
> > @@ -163,7 +163,8 @@ def carrier_id_change(self, cr, uid, ids, carrier_id, context=None):
> >              available_option_ids = []
> >              for available_option in carrier.available_option_ids:
> >                  available_option_ids.append(available_option.id)
> > -                if available_option.state in ['default_option', 'mandatory']:
> > -                if available_option.mandatory \
> > -                        or available_option.by_default:
> 
> Operator position is a matter of taste here it can still be on start of
> second line
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/OCA/carrier-delivery/pull/11/files#r16644073.

http://legacy.python.org/dev/peps/pep-0008/#maximum-line-length

> Make sure to indent the continued line appropriately. The preferred
> place to break around a binary operator is after the operator, not
> before it. Some examples:

.

```
class Rectangle(Blob):

    def __init__(self, width, height,
                 color='black', emphasis=None, highlight=0):
        if (width == 0 and height == 0 and
                color == 'red' and emphasis == 'strong' or
                highlight > 100):
            raise ValueError("sorry, you lose")
        if width == 0 and height == 0 and (color == 'red' or
                                           emphasis is None):
            raise ValueError("I don't think so -- values are %s, %s" %
                             (width, height))
        Blob.__init__(self, width, height,
                      color, emphasis, highlight)
```

yvaucher, line 167: Personally I prefer 

```
if (available_option.mandatory
        or available_option.by_default):
```

Over

```
if (available_option.mandatory or
        available_option.by_default):
```

First option seems more readable to me as you direclty read the or linking the two terms

So I agree that ugly \ should be removed but for position of operator unless we get a concensus I don't think we should block a PR for that.
By the way why second is it the preferred one ?

http://legacy.python.org/dev/peps/pep-0008/#multiline-if-statements
Here are exemple of accepted option for multi-line if statements, it include breaking line before operator

guewen, line 167: As always with styling, it is a subjective opinion, I find more readable the operator at the end. I consider the operator as the end of the line as a clue for the reader that the code continues below.

Anyway the **most readable** way is the way that is used **consistently**. And since pep8 says `The preferred place to break around a binary operator is after the operator, not before it.`, we should stick to this rule.

I won't block a PR for that though.

> By the way why second is it the preferred one ?

Because pep8 says it. For consistency.

bealdav, line 167: "By the way why second is it the preferred one ?"
Because natural fast read of the eyes are from the top level left to bottom right
OR in the left, like says Yannick, seems to be more readable for me too but you noticed "the preferred ..." killer sentence ;-) so ...

my 2 cents

PR #17, opened 2014-07-23
7.0 Add class CarrierAccount, so the specifics carriers modules use all the same base

florian-dacosta:
I had to close my last pull request because I did not target the right branch.
So here is my new pull request.
You can see the closed one for more informations : 
https://github.com/OCA/carrier-delivery/pull/1

### DIFF
--- a/base_delivery_carrier_label/delivery.py
+++ b/base_delivery_carrier_label/delivery.py
@@ -96,39 +96,3 @@ class DeliveryCarrier(orm.Model):
             'delivery.carrier.option',
             'carrier_id', 'Option'),
     }
-
-
-class CarrierAccount(orm.Model):
-    _name = 'carrier.account'
-    _description = 'Base account datas'
-
-    def _get_carrier_type(self, cr, uid, context=None):
-        """ To inherit to add carrier type like Chronopost, Postlogistics..."""
-        return []
-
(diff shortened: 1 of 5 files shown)


base_delivery_carrier_label/delivery.py
```diff
@@ -96,3 +96,41 @@ def __get_carrier_type_selection(self, cr, uid, context=None):
             'delivery.carrier.option',
             'carrier_id', 'Option'),
     }
+
+
+class CarrierAccount(orm.Model):
```
bwrsandman: Can you put this class/model in its own python named `carrier_account.py`?

florian-dacosta: Done!
Thanks for your comment

bwrsandman: Thank you, that is much better!

rvalyi:
:+1: LGTM

yvaucher:
:+1: 
Thanks for your work

To answer @guewen question in https://github.com/OCA/carrier-delivery/pull/1
When you deal with a carrier to generate specific labels, you can need an account to generate part of your label, in other case you can get an image as a stamp to add to your label.

It's true you won't need this model if you install only those modules for using default label.

But this seems generic enough to me to be included in base of labels. This is used by postlogistics and by chronopost.

PR #20, opened 2014-07-31
update travis config

yvaucher:


.travis.yml
```diff
@@ -1,23 +1,28 @@
...
- - travis_install_nightly 7.0
- - pip install coveralls flake8
- - pip install PyPDF2==1.18
- - pip install suds-jurko==0.6
- - git clone https://github.com/OCA/stock-logistics-workflow -b 7.0 $HOME/stock-logistics-workflow
- - git clone https://github.com/OCA/webkit-tools -b 7.0 $HOME/webkit-tools
+  - git clone https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
+  - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
+  - travis_install_nightly
+  - pip install coveralls flake8
```
bwrsandman: coveralls and flake are already installed by `travis_install_nightly` you don't need this line

yvaucher:
@bwrsandman thanks
fixed

.travis.yml
```diff
@@ -1,23 +1,27 @@
...
- - pip install PyPDF2==1.18
- - pip install suds-jurko==0.6
- - git clone https://github.com/OCA/stock-logistics-workflow -b 7.0 $HOME/stock-logistics-workflow
- - git clone https://github.com/OCA/webkit-tools -b 7.0 $HOME/webkit-tools
+  - git clone https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
+  - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
+  - travis_install_nightly
+  - pip install PyPDF2==1.18
+  - pip install suds-jurko==0.6
+  - git clone https://github.com/OCA/stock-logistics-workflow -b 7.0 ${HOME}/stock-logistics-workflow
```
pedrobaeza: Instead of 7.0, use ${VERSION}

.travis.yml
```diff
@@ -1,23 +1,27 @@
...
- - travis_install_nightly 7.0
- - pip install coveralls flake8
- - pip install PyPDF2==1.18
- - pip install suds-jurko==0.6
- - git clone https://github.com/OCA/stock-logistics-workflow -b 7.0 $HOME/stock-logistics-workflow
- - git clone https://github.com/OCA/webkit-tools -b 7.0 $HOME/webkit-tools
+  - git clone https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
+  - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
+  - travis_install_nightly
+  - pip install PyPDF2==1.18
```
bwrsandman, line 17: Why do you freeze the version here?

What's wrong with https://github.com/OCA/maintainer-quality-tools/blob/master/travis/requirements.txt#L19

.travis.yml
bwrsandman: I like this format:

```
- git clone https://github.com/OCA/stock-logistics-workflow ${HOME}/stock-logistics-workflow -b ${VERSION}
```

Easier to read

.travis.yml
yvaucher, line 17: Isn't it better to freeze version for tests?
The chosen version are versions I'm sure it works with.

PyPDF2 is a requirement of delivery_carrier_label_dispatch module

pedrobaeza, line 17: IMHO, better to try with the last one, and see if it fails

bwrsandman, line 17: Probably would be good to catch errors if there are version incompatibilities.

I don't like the idea of using a different version than all other repos.

Can you take pypdf out if it is not absolutely necessary

yvaucher, line 17: Last time I checked, PyPDF was no longer maintained.

PyPDF2 should have some more features like for merging PDF. PyPDF2 is also used in reporting_engine for module base_report_assembler.

I would prefer to keep it.

bwrsandman, line 17: > Last time I checked, PyPDF was no longer maintained.

Really??? :confounded: It's now a core odoo 8.0 dependency...

yvaucher, line 17: http://pybrary.net/pyPdf/

Is there a newer than 1.13 version of pyPDF?

bwrsandman, line 17: Seems not

bwrsandman:
:+1: 

Whoever merges this, please create an issue to fix PEP8 on 7.0

yvaucher:
@bwrsandman already there https://github.com/OCA/carrier-delivery/pull/21

PR #22, opened 2014-08-28
add monkey patch to write file only after commit

yvaucher:
This ensure a file is never created on filesystem while transaction will restart.

This former behavior was creating multiple times the same file on serialized error.

### DIFF
--- a/base_delivery_carrier_files/sql_db.py
+++ b/base_delivery_carrier_files/sql_db.py
@@ -1,0 +1,72 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Author: Yannick Vaucher
+#    Copyright 2014 Camptocamp SA
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+"""
+Monkey patching Cursor to be able to write file on fs only once
+the commit is completed. Thus in case of rollback we won't create
+orphan files.
+"""
+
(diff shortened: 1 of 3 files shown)


base_delivery_carrier_files/sql_db.py
```diff
@@ -0,0 +1,72 @@
...
+
+"""
+Monkey patching Cursor to be able to write file on fs only once
+the commit is completed. Thus in case of rollback we won't create
+orphan files.
+"""
+
+from openerp.sql_db import Cursor
+
+Cursor.transaction_actions = []
```
guewen, line 30: transaction_actions is a list of the class here. It will be shared across instances.
You better have to use hasattr / getattr in the methods add_transaction_action,  commit and rollback in order to add the attribute with a new list if it doesn't exist yet. cr._cnx would no longer be necessary BTW

pedrobaeza:
@yvaucher, can you explain the purpose of this monkey patching? Does it matter for getting the file if the transaction is finished? and Is this because a parallel processing or what?

yvaucher:
@pedrobaeza duplicate file which intend to be read by another proceed are extra unwanted files. So yes it matter.

In parallel processing, with a serialized error, it will retry the transaction. Thus it can generate 2 files.

pedrobaeza:
Why there are now PEP8 errors if target branch is OK?

yvaucher:
~~It might need a rebase~~

gurneyalex:
@pedrobaeza because the max line length was changed from 80 to 79 chars

base_delivery_carrier_files/sql_db.py
```diff
@@ -0,0 +1,72 @@
...
+
+def commit(self):
+    """ Perform an SQL `COMMIT`
+    With write of file after commit to ensure it
+    is written only after commit. And won't be
+    in case of rollback.
+    """
+    res = former_commit(self)
+    to_do = [act for act in self.transaction_actions if act[0] == self._cnx]
+    for act in to_do:
```
gurneyalex, line 51: you need some error handling here, and a test to see what happens if an exception is raised by one of the actions.

base_delivery_carrier_files/sql_db.py
```diff
@@ -0,0 +1,72 @@
...
+            act[1](*act[2])
```
guewen, line 52: The action is not removed from the `transaction_actions` once executed.

pedrobaeza:
@yvaucher are you going to work on this?

yvaucher:
This being in 7.0 and in production of a client we are currently migrating I'm not willing to maintain this PR. Thus I'll close it.

Thanks for the reminder

PR #27, opened 2014-09-09
delivery_carrier_label_dispatch - Improve efficiency of label generation by threading it.

guewen:
For instance for postlogistics, using serialized calls to the SOAP web service
is a really slow as it would wait for a response before sending the next
request.

Also improve exception feed back. And thus exception are non blocking for all
label generation. You will only have to relaunch generation for failed ones.

Pre-requisite : 
- [x] https://github.com/OCA/carrier-delivery/pull/28

I moved this PR from Launchpad [0] on behalf of @yvaucher, he's the author.

[0] https://code.launchpad.net/~camptocamp/carriers-deliveries/7.0-threaded-dispatch-label-generation/+merge/215184

### DIFF
--- a/delivery_carrier_label_dispatch/wizard/generate_labels.py
+++ b/delivery_carrier_label_dispatch/wizard/generate_labels.py
@@ -110,7 +110,7 @@ class DeliveryCarrierLabelGenerate(orm.TransientModel):
                 tracking_ids=tracking_ids,
                 context=context)
             thread_cr.commit()
-        except:
+        except Exception:
             thread_cr.rollback()
             try:
                 raise
--- a/delivery_carrier_label_dispatch/wizard/generate_labels.py
+++ b/delivery_carrier_label_dispatch/wizard/generate_labels.py
@@ -20,11 +20,18 @@
 ##############################################################################
 from operator import attrgetter
 from itertools import groupby
+import Queue
+import threading
+import logging
 
+from openerp import pooler
 from openerp.osv import orm, fields
 from openerp.tools.translate import _
 
 from ..pdf_utils import assemble_pdf
+
+
+_logger = logging.getLogger(__name__)
 
 
 class DeliveryCarrierLabelGenerate(orm.TransientModel):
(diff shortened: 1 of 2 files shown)


guewen:
Rebased after correction of long lines in another PR.

yvaucher:
squashed some of my commits

yvaucher:
We had 1 approve from @nbessi on LP

yvaucher:
@guewen needs a rebase

delivery_carrier_label_dispatch/wizard/generate_labels.py
```diff
@@ -40,51 +50,207 @@ def _get_dispatch_ids(self, cr, uid, context=None):
...
+        """ Generate a label in a thread safe context
+
+        Here we declare a specific cursor so do not launch
+        too many threads
+        """
+        picking_out_obj = self.pool['stock.picking.out']
+        # generate the label of the pack
+        tracking_ids = [pack.id] if pack else None
+        # create a cursor to be thread safe
+        thread_cr = pooler.get_db(cr.dbname).cursor()
```
gurneyalex, line 106: should there not be some exception handling here in case we do exceed the number of connection? (not sure how this is displayed to the user, or how things are retried...)

delivery_carrier_label_dispatch/wizard/generate_labels.py
```diff
@@ -40,51 +50,207 @@ def _get_dispatch_ids(self, cr, uid, context=None):
...
+            thread_cr.rollback()
+            try:
+                raise
+            except orm.except_orm as e:
+                # add information on picking and pack in the exception
+                picking_name = _('Picking: %s') % picking.name
+                pack_num = _('Pack: %s') % pack.name if pack else ''
+                raise orm.except_orm(
+                    e.name,
+                    _('%s %s - %s') % (picking_name, pack_num, e.value))
```
gurneyalex: I'm pretty sure you don't want to translate '%s %s - %s' :smirk:

delivery_carrier_label_dispatch/wizard/generate_labels.py
```diff
@@ -40,51 +50,207 @@ def _get_dispatch_ids(self, cr, uid, context=None):
...
+        finally:
+            thread_cr.close()
+
+    def worker(self, q, q_except):
```
gurneyalex: I object to naming a Queue 'q'. It's sounds smart at first, but let's help people. data_queue, error_queue are better IMO.

guewen:
@yvaucher can you take care of the rebase and the various corrections please? I just tried to rebase and there is a conflict I'm not sure how to fix. You have write access to this repo.

delivery_carrier_label_dispatch/wizard/generate_labels.py
```diff
@@ -40,51 +50,207 @@ def _get_dispatch_ids(self, cr, uid, context=None):
...
+        """ A worker to generate labels
+
+        Takes data from queue q
+
+        And if the worker encounters errors, he will add them in
+        q_except queue
+        """
+        while not q.empty():
+            args, kwargs = q.get()
```
gurneyalex: :warning: race condition!

 you've filled the queue before launching the threads, you have 2 threads and 1 element in the queue, both threads test `not q.empty()` and get True, first thread calls `q.get()`  gets the last element, 2 second thread calls `q.get()` and is blocked forever. 

You need to use either 
- `q.get_nowait()` and handle the possible exception, 
- or `q.get(timeout=5)` and handle the possible exception.

delivery_carrier_label_dispatch/wizard/generate_labels.py
```diff
@@ -40,51 +50,207 @@ def _get_dispatch_ids(self, cr, uid, context=None):
...
+
+        # We will not create a partial PDF if some labels weren't
+        # generated thus we raise catched exceptions by the workers
+        # We will try to regroup all orm exception in one
+        if not q_except.empty():
+
+            error_count = {}
+            messages = []
+            while not q_except.empty():
+                e = q_except.get()
```
gurneyalex: no race condition here, as this code is not executed by threads.

delivery_carrier_label_dispatch/wizard/generate_labels.py
```diff
@@ -40,51 +50,207 @@ def _get_dispatch_ids(self, cr, uid, context=None):
...
+                    else:
+                        error_count[e.name] += 1
+                    messages.append(e.value)
+                else:
+                    # raise other exceptions like PoolError if
+                    # too many cursor where created by workers
+                    raise e
+            titles = []
+            for key, v in error_count.iteritems():
+                titles.append('%sx %s' % (v, key))
```
gurneyalex, line 206: why the 'x'?

delivery_carrier_label_dispatch/wizard/generate_labels.py
```diff
@@ -40,51 +50,207 @@ def _get_dispatch_ids(self, cr, uid, context=None):
...
+            if pack:
+                label = self._find_pack_label(join_cr, uid, wizard, pack,
+                                              context=context)
+            else:
+                label = self._find_picking_label(join_cr, uid, wizard, picking,
+                                                 context=context)
+            if not label:
+                continue  # no label could be generated
+            yield label
+        join_cr.close()
```
gurneyalex: it would be better to have the call to `close()` in a `finally` block.

delivery_carrier_label_dispatch/wizard/generate_labels.py
```diff
@@ -40,51 +50,207 @@ def _get_dispatch_ids(self, cr, uid, context=None):
...
-        pdf_list = [data.decode('base64') for data in data_list if data]
-        pdf_file = assemble_pdf(pdf_list)
-        this.write({'label_pdf_file': pdf_file.encode('base64')})
+        attachment_obj = self.pool['ir.attachment']
+
+        for dispatch in this.dispatch_ids:
+            labels = self._get_all_pdf(cr, uid, this, dispatch,
+                                       context=context)
+            labels = (label.datas for label in labels)
+            labels = (label.decode('base64') for label in labels if labels)
```
gurneyalex, line 251: for the record you're filtering out empty labels here, and there is also a check for empty labels in `assemble_pdf` so you could probably dispense from the filtering here. (AFAIK "" <=base64=> "")

bwrsandman:
Branch needs to be rebased

delivery_carrier_label_dispatch/wizard/generate_labels.py
yvaucher, line 206: X is there after the how many times the error is there.
To say 
"""
44x error B
21x error D
"""

yvaucher:
Fixed and rebased

delivery_carrier_label_dispatch/wizard/generate_labels.py
yvaucher, line 106: Not sure how to catch this.

By the way wouldn't this kind of error block generation if all workers where killed?

max3903:
:+1:

yvaucher:
This has been in production for a while now on one of our clients. It should be mergable quite safely.
As I'm the author of the change on launchpad I won't add my vote though. But I recommend to merge it as for service like swiss postlogistics it is a must have when doing a batch of labels.

PR #33, opened 2014-10-27
[REF] porting v8 base_delivery_...label and affiliate modules

bealdav:
Seems to be near to be ready for merge


yvaucher:
You will need to rebase it your update of 8.0 branch has been done

yvaucher:
Some modules where moved from `__unported__` is that intentional ?

`__unported__/delivery_carrier_b2c/__openerp__.py → delivery_carrier_b2c/__openerp__.py`

bealdav:
@yvaucher  yes, just missed to change to installable: True for b2c. I'm working on Coliposte on V8

yvaucher:
Best practice is to do a PR per module port

Usually when I do this I have my 2 or 3 directory to work "à la Bazaar" and use symlink to not mix them

delivery_carrier_deposit/report/deposit_slip.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  Copyright (C) 2014 Akretion
+  The licence is in the file __openerp__.py
+-->
+
+<openerp>
+<data>
+
+<template id="report_generic_deposit_slip">
```
guewen, line 10: I am not sure this report will be translated, I thought we have to put something like: https://github.com/odoo/odoo/blob/8.0/addons/account/views/report_invoice.xml#L149

delivery_carrier_deposit/report/deposit_slip.xml
```diff
@@ -0,0 +1,60 @@
...
+    <div>DEPOSIT SLIP  <span t-field="o.carrier_type"/></div>
+    <div>N° CLIENT: </div>
+    <div>ACCOUNT NAME:  <span t-field="user.company_id.name"/></div>
+    <div>DOCUMENT N°:  <span t-field="o.name"/> - <span t-field="o.create_date"/></div>
+
+    <table class="table table-condensed">
+    <thead>
+        <tr>
+            <th>Sender Ref</th>
+            <th>Name and address<br />destinataire</th>
```
guewen: destinataire → recipient?

delivery_carrier_deposit/report/deposit_slip.xml
```diff
@@ -0,0 +1,60 @@
...
+<data>
+
+<template id="report_generic_deposit_slip">
+
+<t t-call="report.html_container">
+<t t-call="report.internal_layout">
+
+<div class="page">
+
+<t t-foreach="docs" t-as="o">
```
guewen, line 16: Shouldn't be the loop outside of the `<div class="page">` element?

delivery_carrier_deposit/report/report.xml
```diff
@@ -0,0 +1,27 @@
...
+  Copyright (C) 2014 Akretion
+  The licence is in the file __openerp__.py
+-->
+
+<openerp>
+<data>
+
+<report id="report_deposit_slip"
+    model="deposit.slip"
+    string="Generic Deposit Slip pdf"
```
guewen: remove `pdf` from string

delivery_carrier_deposit/report/report.xml
```diff
@@ -0,0 +1,27 @@
...
+    report_type="qweb-pdf"
+    name="delivery_carrier_deposit.report_generic_deposit_slip"
+    file="delivery_carrier_deposit.report.deposit_slip"
+    />
+
+<report id="report_deposit_slip_html"
```
guewen: Is the html report useful?

guewen:
Thanks for your work to port this module. A few things to address but seems rather good otherwise.

delivery_carrier_deposit/report/deposit_slip.xml
bealdav, line 10: fixed in th end of the file

delivery_carrier_deposit/report/deposit_slip.xml
```diff
@@ -0,0 +1,70 @@
...
+
+    <table class="table table-condensed">
+    <thead>
+        <tr>
+            <th>Sender Ref</th>
+            <th>Name and address<br />recipient</th>
+            <th>Parcel numbers</th>
+            <th class="small">ZIP</th>
+            <th class="small">Country</th>
+            <th class="small">Weight<br />(KG)</th>
```
yvaucher: kg instead of KG

kilo and grams must not be capital letters.

delivery_carrier_deposit/i18n/delivery_carrier_deposit.pot
```diff
@@ -17,7 +17,17 @@ msgstr ""
 
 #. module: delivery_carrier_deposit
 #: sql_constraint:deposit.slip:0
-msgid "'Deposit slip' name must be unique!"
+msgid "'Deposit Slip' name must be unique!"
+msgstr ""
+
+#. module: delivery_carrier_deposit
+#: view:website:delivery_carrier_deposit.report_generic_deposit_slip
+msgid "(KG)"
```
yvaucher: kg instead of KG
http://lamar.colostate.edu/~hillger/correct.htm

delivery_carrier_deposit/report/deposit_slip.xml
```diff
@@ -0,0 +1,70 @@
...
+<t t-call="report.internal_layout">
+
+
+<t t-foreach="docs" t-as="o">
+
+<div class="page">
+
+<h1>Deposit Slip</h1>
+
+    <div>DEPOSIT SLIP  <span t-field="o.carrier_type"/></div>
```
yvaucher: Are those Captal letters there for a reason ?

delivery_carrier_deposit/report/deposit_slip.xml
```diff
@@ -0,0 +1,70 @@
...
+    <div>N° CLIENT: </div>
```
yvaucher: N° is french

Here we want something like Customer No.

delivery_carrier_deposit/report/deposit_slip.xml
```diff
@@ -0,0 +1,70 @@
...
+    <div>ACCOUNT NAME:  <span t-field="user.company_id.name"/></div>
+    <div>DOCUMENT N°:  <span t-field="o.name"/> - <span t-field="o.create_date"/></div>
```
yvaucher: Again, N° should be replaced by No. or no.

bealdav:
failed because depends of this PR https://github.com/OCA/partner-contact/pull/54

yvaucher:
Thanks for the changes

delivery_carrier_deposit/report/deposit_slip.xml
```diff
@@ -0,0 +1,70 @@
...
+    <div>Total Weight: <span t-esc="sum([x.weight for x in o.picking_ids])"/></div>
+</div>
+</t>
+
+</t>
+</t>
+
+</template>
+
+<template id="report_eneric_deposit_slip_trans">
```
guewen: s/report_eneric_deposit_slip_trans/report_generic_deposit_slip_trans/

delivery_carrier_deposit/report/report.xml
```diff
@@ -0,0 +1,19 @@
...
+-->
+
+<openerp>
+<data>
+
+<report id="report_deposit_slip"
+    model="deposit.slip"
+    string="Generic Deposit Slip"
+    report_type="qweb-pdf"
+    name="delivery_carrier_deposit.report_generic_deposit_slip"
```
guewen, line 14: So you want the translated report here

bealdav:
I have added _get_packages_from_picking() method, 
I also done _get_packages_from_moves() that I can add. Should I also add this method ?

guewen:
> I have added _get_packages_from_picking() method,

What is it used for?

delivery_carrier_deposit/report/report.xml
guewen, line 14: You are calling the inner part of the report, you need to call `report_generic_deposit_slip_trans`

bealdav:
It's a common method which can be used for each carriers. It avoids to duplicate code ?

guewen:
Is it normal that delivery_carrier_deposit is now in the same PR?

guewen:
I work on this migration: https://github.com/OCA/carrier-delivery/pull/41 so I'm doing the review at the same time

alexis-via:
@guewen I ported the module delivery_carrier_deposit to the new API because I wanted to enhance this module for a customer project that is about to go live on v8 and I took the advantage of this to port it to the new API.

We are lacking time to split everything in multiple PRs, but the advantage is that we benefit from real-life testing on real-life projects... and it' the best way to find bugs and issues !

base_delivery_carrier_label/stock.py
```diff
@@ -23,9 +23,23 @@
 ##############################################################################
 from openerp.osv import orm, fields
 from openerp.tools.translate import _
+import openerp.addons.decimal_precision as dp
 
 
-class stock_picking(orm.Model):
+class StockQuantPackage(orm.Model):
+    _inherit = 'stock.quant.package'
+    _order = 'id desc'
```
guewen: What is the reason for overriding this? The default is:

```
    _parent_order = 'name'
    _order = 'parent_left'

```

delivery_carrier_deposit/stock.py
```diff
@@ -0,0 +1,127 @@
...
+        self.weight = weight
+        self.number_of_packages = number_of_packages
+
+    @api.model
+    def _get_carrier_type_selection(self):
+        return self.env['delivery.carrier']._get_carrier_type_selection()
+
+    name = fields.Char(
+        string='Name', readonly=True, states={'draft': [('readonly', False)]},
+        default='/', copy='/')
```
guewen: `copy` is supposed to be a boolean

delivery_carrier_deposit/stock.py
```diff
@@ -0,0 +1,127 @@
...
+        if vals.get('name', '/') == '/':
+            vals['name'] = self.env['ir.sequence'].next_by_code(
+                'delivery.deposit')
+        return super(DepositSlip, self).create(vals)
+
+    @api.one
+    def copy(self, default=None):
+        default = dict(default or {})
+        default['name'] = self.env['ir.sequence'].next_by_code(
+            'delivery.deposit')
```
guewen: Why not put `default['name'] = '/'`, so `create()` will call the next sequence number? (remove duplication)

delivery_carrier_deposit/wizard/deposit.py
```diff
@@ -0,0 +1,81 @@
...
+    def _prepare_deposit_slip(self):
+        return {
+            'carrier_type': self.carrier_type,
+            'company_id': self.env.user.company_id.id,
+            }
+
+    @api.multi
+    def create_deposit_slip(self):
+        # I can't set api.one because I return an action
+        assert len(self) == 1, 'only 1 recordset allowed'
```
guewen: Use `self.ensure_one()`

delivery_carrier_deposit/wizard/deposit.py
```diff
@@ -0,0 +1,81 @@
...
+        self = self[0]
```
guewen: You can remove this line. When a recordset contains 1 record, it is considered as a "singleton" and the attributes work directly on the record.

delivery_carrier_deposit/report/report.xml
guewen, line 14: Still needs to be corrected here.

guewen:
@bealdav can you review and merge that in your PR please? https://github.com/bealdav/carrier-delivery/pull/1

guewen:
We have currently no way to print the labels from the new javascript "barcode interface". Are you going to work on that? I am not sure how we should do, because the picking disappears once the packing is finished. Do you have an idea?

bealdav:
@guewen no idea for barcode interface. We do not use this interface for now. Alexis tried but js seems to be less inefficient than python.
As you can see here https://github.com/akretion/carrier-delivery-colipostefr/blob/8.0/delivery_carrier_label_colissimo/stock.py#L19, you have to override 2 methods if you want to control the 2 interfaces: maybe to switch in base module

bealdav:
@guewen I'll review and test your works early next week on a real cases. Thanks a lot frr your work

guewen:
> As you can see here https://github.com/akretion/carrier-delivery-colipostefr/blob/8.0/delivery_carrier_label_colissimo/stock.py#L19, you have to override 2 methods if you want to control the 2 interfaces: maybe to switch in base module

Thanks. The issue here won't be to generate the labels but to allow the user to print them :unamused:

bealdav:
LGTM for me

base_delivery_carrier_label/stock.py
```diff
@@ -0,0 +1,298 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from openerp import models, fields, api, exceptions, _
+import openerp.addons.decimal_precision as dp
+
+
+class StockQuantPackage(models.Model):
+    _inherit = 'stock.quant.package'
+    _order = 'id desc'
```
guewen: I already asked this question and think I got not answer.
The override of the order here is highly suspicious.

For the record, here is the one defined in `stock`:

```
 _parent_order = 'name'
_order = 'parent_left'
```

The order here is based on `parent_left` so I guess it is for respecting the tree's order.
Maybe you want to change the `_parent_order`, but what is the reason at a first place to change that in this module?

:-1: for that override

delivery_carrier_deposit/report/deposit_slip.xml
guewen, line 10: This is not fixed. Please see https://github.com/OCA/carrier-delivery/pull/33/files#r20997336

base_delivery_carrier_label/stock.py
alexis-via: Well, I just wanted to see the latest packages at the top, and not the oldest packages at the top. I'll revert it if you're not happy with this ; it's not important, it's just a small usabilty stuff than I can put in another module.

delivery_carrier_deposit/report/deposit_slip.xml
alexis-via, line 10: @guewen About the language of the qweb report, by default, if you don't have the block of XML code like https://github.com/odoo/odoo/blob/8.0/addons/account/views/report_invoice.xml#L149 then the report is generated in the language of the user, so if the user is not using English, then the report will be translated.

guewen, line 10: @alexis-via I know... I asked @bealdav to add it. But please read my comments **below** (https://github.com/OCA/carrier-delivery/pull/33/files#r20997336)

```
+<report id="report_deposit_slip"
+ model="deposit.slip"
+ string="Generic Deposit Slip"
+ report_type="qweb-pdf"
+ name="delivery_carrier_deposit.report_generic_deposit_slip"
+ file="delivery_carrier_deposit.report.deposit_slip"
+ />
```

It points to the `delivery_carrier_deposit.report_generic_deposit_slip` template, it must points to the translated template! 

```
+<report id="report_deposit_slip"
+ model="deposit.slip"
+ string="Generic Deposit Slip"
+ report_type="qweb-pdf"
+ name="delivery_carrier_deposit.report_generic_deposit_slip_trans"
+ file="delivery_carrier_deposit.report.deposit_slip"
+ />
```

guewen, line 10: Or you consider the deposit slip should be in the language of the user, in that case you remove the `report_generic_deposit_slip_trans` template.

guewen:
@bealdav is it still work in progress?

bealdav:
I have not fixed the translation problem in the report report_deposit_slip. I you have some time for this thank you.

alexis-via:
For me this is stable and can be merged. I already use this branch in production.

yvaucher:
I put it as needs review then

pedrobaeza:
Please put module descriptions with OCA template and on a README.rst, and I will merge this.

delivery_carrier_deposit/report/deposit_slip.xml
yvaucher, line 10: This is fixed

yvaucher:
Almost there, as @pedrobaeza said please move description in README.rst using template:
https://github.com/OCA/maintainers-tools/blob/master/template/module/README.rst

base_delivery_carrier_label/stock.py
```diff
@@ -0,0 +1,307 @@
...
+    @api.multi
+    @api.returns('stock.quant.package')
+    def _get_packages_from_picking(self):
+        """ Get all the packages from the picking """
+        self.ensure_one()
+        operation_obj = self.env['stock.pack.operation']
+        operations = operation_obj.search(
+            [('result_package_id', '!=', False),
+             ('picking_id', '=', self.id)]
+        )
```
guewen, line 228: I think I spotted an issue there but I am not sure whether it is a misuse from my part or... I dont know.

It won't work in this scenario:
- Configure the warehouse in 3 steps (pick + pack + ship)
- On the pick, put all the lines in packages
- On the pack, keep the lines in the same packages
- On the ship, keep the lines in the same packages again

The packages have no `result_package_id`, so no labels are generated.

@bealdav @alexis-via any idea?

guewen:
Similar scenario that don't work as we could expect not strictly related to this module I fear (in fact I observed it in delivery_carrier_colipostefr but I wonder if it is not a bug in the stock module):
-  Configure the warehouse in 3 steps (pick + pack + ship)
-  On the pick, put all the lines in packages
-  On the pack, keep the same packages
-  On the ship, transfer the existing packages in new packages

The packages in ship have no product_id so generation of labels might fails.
Example, here https://github.com/akretion/carrier-delivery-colipostefr/blob/8.0/delivery_carrier_colipostefr/stock.py#L236 the weight is 0 because `pack_op.product_id` is empty.

base_delivery_carrier_label/stock.py
guewen, line 228: > The packages have no result_package_id, so no labels are generated.

If I guess (can't say "understand" at this point) correctly, the `stock.pack.operation` could have a `package_id`, a `result_package_id` or both. If we have only a `package_id`, we should keep this package. So, the logic should be changed throughout the module to look first in `result_package_id` then in `package_id`.
Does this analysis sounds good to you?

bealdav:
@guewen, I don't tested it with these scenari, then I suppose your are right. I need some time for it.
Thanks to point it

guewen:
> @guewen, I don't tested it with these scenari, then I suppose your are right. I need some time for it.
> Thanks to point it

I will try that and make a new PR upon this one for the first scenario.
The second one I am unsure about how it should be resolved.

guewen:
Here it is: https://github.com/bealdav/carrier-delivery/issues/2

guewen:
I can't reproduce the scenario in https://github.com/OCA/carrier-delivery/pull/33#issuecomment-72036213 so it was maybe related to the fix I did in bealdav#2 and https://github.com/akretion/carrier-delivery-colipostefr/pull/5

UPDATE: yes I can reproduce it, working on a fix.
UPDATE: got it in https://github.com/akretion/carrier-delivery-colipostefr/pull/5

yvaucher:
@guewen @bealdav can we remove the needs fixing here?

yvaucher:
:+1: 

Thanks for the work @bealdav I'll proceed with the merge of this PR

PR #34, opened 2014-10-28
8 update from 7.0

bealdav:
intial refresh PR before PR on modules


.travis.yml
```diff
@@ -13,11 +13,16 @@ virtualenv:
...
-  - travis_install_nightly ${VERSION}
+  - travis_install_nightly
+  - pip install PyPDF2==1.18
+  - pip install suds-jurko==0.6
+  - git clone https://github.com/OCA/stock-logistics-workflow ${HOME}/stock-logistics-workflow -b ${VERSION}
+  - git clone https://github.com/OCA/partner-contact ${HOME}/partner-contact -b ${VERSION}
+  - git clone https://github.com/OCA/webkit-tools ${HOME}/webkit-tools -b ${VERSION}
 
 script:
   - travis_run_flake8
```
pedrobaeza: This can be removed

pedrobaeza:
You shouldn't move README.md and LICENSE to `__unported__`

pedrobaeza:
The rest seems OK.

.travis.yml
bealdav: from which line to which line exactly ? 
I need this line for 1 module:
git clone https://github.com/OCA/partner-contact ${HOME}/partner-contact -b ${VERSION}

pedrobaeza: I mean `- travis_run_flake8`. Flake8 check is now done in `travis_run_tests`.

yvaucher: @bealdav travis_run_flake8 can be remove as it is now part of travis_run_tests

.coveragerc
```diff
@@ -1,9 +1,6 @@
-# Config file .coveragerc
-# adapt the include for your project
-
 [report]
 include =
-    */carrier-delivery/*
+    */OCA/carrier-delivery/*
 
 omit =
     */tests/*
```
yvaucher: .coveragerc can be removed but I won't block this PR for that

__unported__/delivery_carrier_deposit/__openerp__.py
```diff
@@ -0,0 +1,66 @@
...
+    'website': 'http://www.akretion.com/',
+    'data': [
+        'stock_view.xml',
+        'wizard/deposit.xml',
+        'ir_sequence_data.xml',
+        'report/report.xml',
+        'security/ir.model.access.csv',
+    ],
+    'demo': [],
+    'installable': True,
```
yvaucher: Should be set to False

__unported__/delivery_carrier_b2c/__openerp__.py
```diff
@@ -0,0 +1,59 @@
...
+----
+
+
+    """,
+    'website': 'http://www.akretion.com/',
+    'data': [
+        'partner_view.xml',
+    ],
+    'tests': [],
+    'installable': True,
```
yvaucher: Should be set to False

__unported__/delivery_optional_invoice_line/__openerp__.py
```diff
@@ -0,0 +1,44 @@
...
+    'depends': [
+        'delivery',
+    ],
+    'data': [
+        'delivery_view.xml',
+    ],
+    'test': [
+        'test/delivery_optional_invoice_line.yml',
+    ],
+    'installable': True
```
yvaucher: Should be set to False

__unported__/delivery_carrier_label_dispatch/__openerp__.py
```diff
@@ -45,7 +45,7 @@
         'wizard/generate_labels_view.xml',
     ],
     'tests': [],
-    'installable': False,
+    'installable': True,
```
yvaucher: Should be set to False

__unported__/delivery_carrier_deposit/__openerp__.py
bealdav: It's not installable because it is in **unported** folder, change to False is a waste of time.
I you mv module folder it's to install it, isn't it ?

pedrobaeza: Yeah, more or less, but it's a double check to avoid using non valid modules in a version for what is not intended. It was agreed this way in the beginning.

yvaucher: moving in `__unported__` folder is to say clearly it isn't ported thus we won't check pep8 and pylint it.

and setting to installable to False is to say in current state it doesn't work so a user downloading this repo won't be able to install it even though he adds a symlink.

It was part of migration procedure: https://github.com/OCA/maintainers-tools/wiki/Migration-Launchpad-%E2%86%92-GitHub under "Set all the modules of the 8.0 branches to "installable": False"

I would prefer to have all repo at the same state.

bealdav: OK I undertsand,  we started to do like that but as you said it's double check but community time could be used in more efficient way if we can avoid to do double work for the same feature.
At community beginning on git there was no unported folder and then 'installable': False was mandatory, but now I really think it's a waste of time.
What does the community think ? Thank you

pedrobaeza: Yannick has explained very well the motivation, and in the link he has provided is a command line that do it automatically, so it's not too much time consuming. Please do it.

yvaucher:
thanks :+1:

yvaucher:
I proceed with the merge as this is a maintenance PR

PR #42, opened 2014-11-28
add a default value to company_id

bealdav:


delivery_carrier_deposit/stock.py
```diff
@@ -59,11 +60,20 @@ def _get_carrier_type_selection(self, cr, uid, context=None):
             'Company'),
     }
 
+    def _get_default_company(self, cr, uid, context=None):
+        company_id = self.pool['res.users']._get_company(cr, uid, context=context)
+        if not company_id:
```
pedrobaeza: You don't need to put this check, because in the security rule, you also include the possibility of company_id=False.

pedrobaeza:
Please fix PEP8 error:

```
./delivery_carrier_deposit/stock.py:64:80: E501 line too long (82 > 79 characters)
```

pedrobaeza:
:+1:

bguillot:
:+1: 

Thanks

yvaucher:
:+1:

PR #47, opened 2014-12-22
base_delivery_carrier_files module migration to Odoo v8

kkarolis:


__unported__/base_delivery_carrier_files/carrier_file.py
```diff
@@ -147,15 +130,29 @@ def generate_files(self, cr, uid, carrier_file_id, picking_ids,
...
             else:
                 carrier_file_id = carrier_file_id[0]
-        carrier_file = self.browse(cr, uid, carrier_file_id, context=context)
-        return self._generate_files(cr, uid, carrier_file, picking_ids,
-                                    context=context)
-
-
-class delivery_carrier(orm.Model):
+        carrier_file = self.browse(carrier_file_id)
+        return self._generate_files(carrier_file, picking_ids)
```
yvaucher: Here you can call

```
carrier_file._generate_files(picking_ids)
```

And rewrite _generate_files to use self instead of carrier_file

__unported__/base_delivery_carrier_files/carrier_file.py
```diff
@@ -76,17 +57,17 @@ def _write_file(self, cr, uid, carrier_file, filename, file_content,
...
+                                        (carrier_file.name,))
         full_path = os.path.join(carrier_file.export_path, filename)
         with open(full_path, 'w') as file_handle:
             file_handle.write(file_content)
         return True
 
-    def _generate_files(self, cr, uid, carrier_file, picking_ids,
-                        context=None):
+    @api.multi
+    def _generate_files(self, carrier_file, picking_ids):
```
yvaucher: If this is a api.multi, you should rewrite it to use `self` instead of `carrier_file`

__unported__/base_delivery_carrier_files/carrier_file.py
```diff
@@ -97,13 +78,13 @@ def _generate_files(self, cr, uid, carrier_file, picking_ids,
...
         :return: True if successful
         """
-        if context is None:
-            context = {}
-        picking_obj = self.pool.get('stock.picking')
+        # Convert frozendict context to editable context
+        context = dict(self.env.context or {})
+        picking_obj = self.env["stock.picking"]
         log = logging.getLogger('delivery.carrier.file')
         file_generator = new_file_generator(carrier_file.type)
```
yvaucher:
```
self.type
```

__unported__/base_delivery_carrier_files/carrier_file.py
```diff
@@ -97,13 +78,13 @@ def _generate_files(self, cr, uid, carrier_file, picking_ids,
...
-        pickings = [picking for picking in
-                    picking_obj.browse(cr, uid, picking_ids, context=context)]
+        pickings = [picking for picking in picking_obj.browse(picking_ids)]
+
         # must return a list of generated pickings ids to update
         files = file_generator.generate_files(pickings, carrier_file)
```
yvaucher: s/carrier_file/self

__unported__/base_delivery_carrier_files/carrier_file_view.xml
```diff
@@ -2,7 +2,6 @@
 <openerp>
     <data>
         <record id="view_delivery_carrier_file_tree" model="ir.ui.view">
-            <field name="name">delivery.carrier.file.tree</field>
```
yvaucher: Why removing view names ?

__unported__/base_delivery_carrier_files/carrier_file.py
```diff
@@ -22,49 +22,30 @@
...
-                                           'generated when a delivery order '
-                                           'is processed. If activated, each '
-                                           'delivery order will be exported '
-                                           'in a separate file.'),
-    }
-
-    def _write_file(self, cr, uid, carrier_file, filename, file_content,
-                    context=None):
+    @api.multi
+    def _write_file(self, carrier_file, filename, file_content):
```
yvaucher: Remove `carrier_file` to use `self` instead.

__unported__/base_delivery_carrier_files/wizard/generate_carrier_files_view.xml
```diff
@@ -40,12 +41,12 @@
         </record>
 
         <record model="ir.values" id="delivery_carrier_file_generate_ir_values">
-            <field name="object" eval="1" />
+            <!-- <field name="object" eval="1" /> -->
```
yvaucher: Commented line to remove I see no `object` field on `ir.values` model

yvaucher:
Hello,

Thanks for the work done here. I added few inline comments.
Plus, don't forget to move the module out of `__unported__` directory

pedrobaeza:
@kkarolis any news on @yvaucher's remarks?

kkarolis:
@pedrobaeza I was busy the whole month but started working on it just yesterday. The problem is that the latest odoo updates broke the module and some unexpected errors showed up. I am working to fix them right now

pedrobaeza:
OK, just tell me if you need help.

kkarolis:
There is an issue which affects this module(prevents execution of test/carrier_file_manual.yml). It will work after https://github.com/odoo/odoo/pull/4936 is merged.

pedrobaeza:
We should make a workaround meanwhile. Can you empty that field in your module?

kkarolis:
Well, it really is outside the scope of this module. The only workaround we could do is to comment out the test itself.

pedrobaeza:
If you docummented the reason why you do it (putting that due to the issue #...), I think it can be a solution.

base_delivery_carrier_files/carrier_file.py
```diff
@@ -0,0 +1,153 @@
...
+            # we pass the errors because the files can still be
+            # generated manually
+            # at first I would like to open a new cursor and
+            # commit the write after each file created
+            # but I encountered lock because the picking
+            # was already modified in the current transaction
+            try:
+                # Instead of replacing everything, append to existing context
+                if self.with_context(**context)._write_file(
+                    filename, file_content):
```
yvaucher:
```
./base_delivery_carrier_files/carrier_file.py:105:5: E125 continuation line with same indent as next logical line
```

Please add 4 spaces here so we can see it is not part of the block

base_delivery_carrier_files/carrier_file.py
```diff
@@ -0,0 +1,153 @@
...
+            filename, file_content, picking_ids = f
+            # we pass the errors because the files can still be
+            # generated manually
+            # at first I would like to open a new cursor and
+            # commit the write after each file created
+            # but I encountered lock because the picking
+            # was already modified in the current transaction
+            try:
+                # Instead of replacing everything, append to existing context
+                if self.with_context(**context)._write_file(
```
yvaucher: Why copying the whole context to merge it to itself afterwards?
You just want to alter `picking_id` if necessary.

You could do that by writing something like

    self = self.with_context(picking_id=pickings and pickings[0].id)

base_delivery_carrier_files/carrier_file.py
```diff
@@ -0,0 +1,153 @@
...
+                                 we have to generate a file
+        :return: True if successful
+        """
+        # Convert frozendict context to editable context
+        context = dict(self.env.context or {})
+        log = logging.getLogger('delivery.carrier.file')
+        file_generator = new_file_generator(self.type)
+
+        picking_obj = self.env["stock.picking"]
+        pickings = [picking for picking in picking_obj.browse(picking_ids)]
```
yvaucher: I would suggest to browse those pickings only once:

```
pickings = self.env['stock.picking'].browse(picking_ids)
```

And then create a picking_list for `generate_files` method.

Thus you wouldn't need to browse it again at line 106

base_delivery_carrier_files/carrier_file.py
```diff
@@ -0,0 +1,153 @@
...
+        for all picking_ids
+
+        :param browse_record carrier_file: browsable carrier file
+                                           configuration
+        :param list picking_ids: list of ids of pickings for which
+                                 we have to generate a file
+        :return: True if successful
+        """
+        # Convert frozendict context to editable context
+        context = dict(self.env.context or {})
```
yvaucher: Why do you copy entire self.env.context if it is to only change key `picking_id` afterward?

I think you could only define an empty dict here.

base_delivery_carrier_files/stock.py
```diff
@@ -0,0 +1,126 @@
...
+        :return: True if successful
+        """
+        carrier_file_ids = {}
+        for picking in self:
+            if picking.picking_type_id.code != 'outgoing':
+                continue
+            if not recreate and picking.carrier_file_generated:
+                continue
+            carrier = picking.carrier_id
+            if not carrier or not carrier.carrier_file_id:
```
yvaucher: no need to check carrier.
If `carrier` is Null, `carrier.carrier_file_id` will be Null too.

base_delivery_carrier_files/stock.py
```diff
@@ -0,0 +1,126 @@
...
+        return True
+
+    @api.multi
+    def action_done(self):
+        result = super(stock_picking, self).action_done()
+        self.generate_carrier_files(auto=True)
+        return result
+
+    @api.multi
+    def copy(self, default=None):
```
yvaucher: You should use `copy=False` as attribute of field `carrier_file_generated`

base_delivery_carrier_files/wizard/generate_carrier_files.py
```diff
@@ -0,0 +1,58 @@
...
+from openerp.tools.translate import _
+
+
+class DeliveryCarrierFileGenerate(models.TransientModel):
+
+    _name = 'delivery.carrier.file.generate'
+
+    @api.multi
+    def _get_picking_ids(self):
+        context = self.env.context or {}
```
yvaucher: `or {}` should not be necessary when getting `self.env.context` this is done by method `__new__()` of env which create a frozen dict.

base_delivery_carrier_files/wizard/generate_carrier_files.py
```diff
@@ -0,0 +1,58 @@
...
+        if (context.get('active_model', False) == 'stock.picking' and
+                context.get('active_ids', False)):
```
yvaucher: not necessary to define `False` as it would be `None` if key is not found in dict.

base_delivery_carrier_files/wizard/generate_carrier_files.py
```diff
@@ -0,0 +1,58 @@
...
+from openerp import models, fields, exceptions, api
+from openerp.tools.translate import _
+
+
+class DeliveryCarrierFileGenerate(models.TransientModel):
+
+    _name = 'delivery.carrier.file.generate'
+
+    @api.multi
+    def _get_picking_ids(self):
```
yvaucher: Please rename it `_get_pickings` as you are not returning ids but a record set

base_delivery_carrier_files/wizard/generate_carrier_files.py
```diff
@@ -0,0 +1,58 @@
...
+                                                recreate=self.recreate)
+        return {'type': 'ir.actions.act_window_close'}
+
+    picking_ids = fields.Many2many('stock.picking',
+                                   string='Delivery Orders',
+                                   default=_get_picking_ids)
+    recreate = fields.Boolean(
+        'Recreate files',
+        help="If this option is used, new files will be generated "
+        "for selected picking even if they already had one.\n"
```
yvaucher: Would be better to indent those lines after `=` sign.

base_delivery_carrier_files/stock.py
```diff
@@ -0,0 +1,126 @@
...
+        done_pickings = self.browse([id for id, key in res.iteritems() if
+                                     key == "done"])
+        if done_pickings:
+            done_pickings.generate_carrier_files()
+        for picking in pickings:
+            picking.state = res[picking.id]
+
+    @api.multi
+    def _get_pickings(self):
+        return super(stock_picking, self)._get_pickings(None, None)
```
yvaucher: Is this method really necessary? If yes, please provide a docstring explaining why.

yvaucher:
Thanks for you changes, I added some more remarks.
About commenting a test, I feel it is quite a regression unless you explain why. I could be comprehensible for a temporary bug on official branch that need fixing with no easy work around.

yvaucher:
@kkarolis any update?

kkarolis:
@yvaucher the previous issue seems to have gone away, however now theres another problem, which seems like an odoo bug, didin't have time to look into it further. The problem is related to setting default picking lines on the generator wizard. Upon opening the wizard, the picking lines just don't get attached to the wizard (even though the function returning default values returns the recordset). Even on documentation its shown that it should work, however it does not. Worst of all, I haven't managed to find any sources in odoo 8.0 which return default m2m field to check it if it works in any case.

kkarolis:
Hello, sorry for the delay again. I have checked the problem why travis build fails and noticed that it struggles with the creation of demo data in *. To be more exact stock.move's defined in carrier_file_demo.yml cannot be created. The error occurs only for the first time when its installed TOGETHER with all other dependencies. If installed by steps, e.g. first install delivery followed by this module, the error does not occur. As far as I have investigated the issue seems to be my redefinition of state field on stock.picking model. I am unable to reproduce the exact same behaviour of state field as it was before in the old api. Tried removing api.multi decorator, even redefining the compute to always move state to draft (for rec in self: rec.state = 'draft') does not seem to help. Im a little bit stuck here, could anyone provide any solutions to try? Thanks

gurneyalex:
created @kkarolis in the OCA database, but we don't have a CLA as far as I can tell.

base_delivery_carrier_files/wizard/generate_carrier_files.py
```diff
@@ -0,0 +1,58 @@
...
+        Call the creation of the delivery carrier files
+        """
+        if not self.pickings:
+            raise exceptions.except_orm(_('Error'),
+                                        _('No delivery orders selected'))
+        self.pickings.generate_carrier_files(
+            auto=False, recreate=self.recreate)
+        return {'type': 'ir.actions.act_window_close'}
+
+    pickings = fields.Many2many('stock.picking',
```
yvaucher, line 49: Why did you renamed this m2m removing `ids`?
By convention we use `_ids` everywhere for o2m and m2m

kkarolis, line 49: Well the convention was to use ids because it actually was ids previously, now its just a recordset right? I even remember a similar remark regarding this earlier in this MR. Do you still wish me to rename it?

pedrobaeza, line 49: We have talked about this in some occassions, and the "convention" is not clear. It's very confused to have code like this: `self.line_id.order_id.picking_ids.ids`. I also prefer this naming convention. I already do in all my new modules. The only thing is that it must have `oldname='picking_ids'` for easying migration path.

base_delivery_carrier_files/wizard/generate_carrier_files_view.xml
```diff
@@ -3,12 +3,13 @@
     <data>
 
         <record id="view_delivery_carrier_file_generate" model="ir.ui.view">
-            <field name="name">delivery.carrier.file.generate</field>
+            <field name="name">delivery.carrier.file.generate.form</field>
             <field name="model">delivery.carrier.file.generate</field>
             <field name="arch" type="xml">
-                <form string="Carrier Files" version="7.0">
+                <form string="Carrier Files" version="8.0">
```
yvaucher: You can drop the `version="8.0"` this is deprecated and was there only for 7.0

base_delivery_carrier_files/test/carrier_file.yml
```diff
@@ -15,7 +15,8 @@
 -
   I confirm outgoing shipment of 130 kgm Ice-cream.
 -
-  !workflow {model: stock.picking, action: button_confirm, ref: outgoing_shipment_carrier_file}
+  !python {model: stock.picking}: |
+    self.action_confirm(cr, uid, ref("outgoing_shipment_carrier_file"))
```
yvaucher: You can use
`!python {model: stock.picking, id: <xml ref>}: |`

Then you can use self as in new API:
`self.action_confirm()`

base_delivery_carrier_files/test/carrier_file_manual.yml
```diff
@@ -15,7 +15,8 @@
 -
   I confirm outgoing shipment of 130 kgm Ice-cream.
 -
-  !workflow {model: stock.picking, action: button_confirm, ref: outgoing_shipment_carrier_file_manual}
+  !python {model: stock.picking}: |
+    self.action_confirm(cr, uid, ref("outgoing_shipment_carrier_file"))
```
yvaucher: Same as in previous test file, you can use new v8.0 API by setting id: xmlid

base_delivery_carrier_files/wizard/generate_carrier_files.py
```diff
@@ -0,0 +1,58 @@
...
+            return self.env["stock.picking"].browse(
+                context["active_ids"])
+
+    @api.multi
+    def action_generate(self):
+        """
+        Call the creation of the delivery carrier files
+        """
+        if not self.pickings:
+            raise exceptions.except_orm(_('Error'),
```
yvaucher: except_orm is deprecated, please use `Warning`

yvaucher:
@kkarolis The diff is not clean, could you rebase instead of doing merges?

kkarolis:
Hey, I have made the changes, however I still need some help regarding the travis fails. The issue is the one I mentioned earlier, thanks :)

yvaucher:
@kkarolis What if you define the state field in yaml file ?

A previous hack which was working in earlier version for computed store fields was to do something like that:
https://github.com/odoo/odoo/issues/3922#issuecomment-75042705

Otherwise it could be due to selection redefinition: https://github.com/odoo/odoo/issues/2797

kkarolis:
I have tried, adding `state: 'draft'` to related pickings without any success. I have also tried inserting the following block model with older api, however the issue still remains.

``` python
from openerp.osv import fields


class stock_picking(models.Model):
    _inherit = 'stock.picking'
    _columns = {
        'state': fields.selection([
            ('draft', 'Draft'),
            ('cancel', 'Cancelled'),
            ('waiting', 'Waiting Another Operation'),
            ('confirmed', 'Waiting Availability'),
            ('partially_available', 'Partially Available'),
            ('assigned', 'Ready to Transfer'),
            ('done', 'Transferred')], string='State'),
    }

from openerp import fields
```

tremlin:
Thanks for the migration!
I have no idea how to fix the yaml demo error except by reverting the `state` field of `stock.picking` completely back to the old API. If you decide to keep the new API, then the `_state_get` method of `stock.picking` should have the same signature as the `_state_get` from the core. That is, it should take additional arguments `field_name` and `arg` and it should return a dictionary instead of `None`. Besides of this incompatibility the code worked fine for me.

paulius-sladkevicius:
What left to do that we can merge it to 8.0?

yvaucher:
@komsas There is still the error in tests, it might requires to stick to old API in this case. Plus reporting the issue on Odoo

yvaucher:
Can you do a rebase to get travis status ?

yvaucher:
@kkarolis Sorry I missed your rebase

:+1:

mileo:
Hi,

How do you fix undefined method get? 

We are having the same issue in https://github.com/OCA/l10n-brazil/pull/342, on sale field amount_untaxed with track_visibility='aways'

Thanks in advance

PR #48, opened 2014-12-24
define context if not set in stock.py

bealdav:
define context in generate_labels().
In some case ie errpeek you send model('stock.picking.out').generate_labels([my_id]) and context is not passed
and it breaks at context_attachment = context.copy()

### DIFF
--- a/base_delivery_carrier_label/stock.py
+++ b/base_delivery_carrier_label/stock.py
@@ -110,6 +110,8 @@ class stock_picking(orm.Model):
         the labels only of these trackings.
 
         """
+        if context is None:
+            context = {}
         shipping_label_obj = self.pool.get('shipping.label')
 
         pickings = self.browse(cr, uid, ids, context=context)


pedrobaeza:
:+1:, although Travis error is confused.

bealdav:
@pedrobaeza yes really confuse ;-(

bealdav:
:white_check_mark: PR has been fixed
Here is the fail on travis. It's not about the PR.
Thanks for the the review

```
2014-12-24 12:07:36,872 23892 TEST openerp_test openerp.tools.yaml_import: Creating !record {model: stock.partial.picking, id: partial_outgoing}:
with {'picking_id': 'outgoing_shipment_carrier_file'}
2014-12-24 12:07:36,884 23892 TEST openerp_test openerp.tools.yaml_import: Creating !python {model: stock.partial.picking}: |
with self.do_partial(cr, uid, [ref('partial_outgoing')], context=context)
2014-12-24 12:07:36,889 23892 ERROR openerp_test openerp.tools.yaml_import: (u'Warning!', u'The unit of measure rounding does not allow you to ship "130.0 kg", only rounding of "0.0 kg" is accepted by the Unit of Measure.')
```

sebastienbeau:
:+1: 
The trouble with travis already exist in the main branch (7.0) and it's not related to this PR

PR #51, opened 2015-02-04
[IMP] add apps site description for carrier gls

bealdav:
To have a good display on https://www.odoo.com/apps

### DIFF
--- a/delivery_carrier_label_gls/static/description/index.html
+++ b/delivery_carrier_label_gls/static/description/index.html
@@ -1,0 +1,41 @@
+
+<section class="oe_container oe_website_only">
+
+<div class="oe_row oe_more_spaced">
+    <h3 class="oe_slogan">GLS Services Integration with Odoo</h3>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Can you please use OCA template for this purpose: https://github.com/OCA/maintainers-tools/blob/master/template/module/static/description/index.html

bealdav:
OCA template is for v8, in v7 description is different

pedrobaeza:
I think for HTML description can be used the same, isn't it?

bealdav:
Have you got an example of a module with this template displayed correctly in apps ? 
If I can see one, I'll update my page.
Thanks

pedrobaeza:
No, not yet, but it's a good moment to try :wink:

Viggor:
Despite there is no unified system between description in Odoo system and Odoo apps (I could be wrong) I give a :+1: . The travis fail is not caused by this modification.

yvaucher:
For v8.0 it definitely will need to follow OCA template. I have no strong opinion about v7.0 html though as most of 7.0 modules' readme are not following OCA template
:+1:

PR #57, opened 2015-03-11
7 gls feedback ws

bealdav:
Allow to save data about web service exchange in attachment

gls_colis

Allow to send these data to the web service provider in case of logistics incident

### DIFF
--- a/delivery_carrier_label_gls/stock.py
+++ b/delivery_carrier_label_gls/stock.py
@@ -244,7 +244,9 @@ class StockPicking(orm.Model):
             if label['tracking_number']:
                 label_info['name'] = '%s%s.zpl' % (label['tracking_number'],
                                                    label['filename'])
-            labels.append(label_info)
+            if picking.company_id.country_id == 'FR':
+                # TODO implements traceability to other countries than FR
+                labels.append(label_info)
             traceability.append(self._record_ws_exchange(label, pack))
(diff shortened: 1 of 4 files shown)


bealdav:
@alexis-via  @guewen do you think we could have the same principle for Coliposte V8 international ?
thanks for a review

for Coliposte we need to store xml soap request. Call me for more info

Viggor:
Works fine with french partners, but with foreign partners there is a exception in GLS lib.

delivery_carrier_label_gls/stock.py
```diff
@@ -250,6 +259,40 @@ def _generate_gls_labels(
...
+            'file_type': 'text/plain',
+        }
+        return self.pool['ir.attachment'].create(
+            cr, uid, data, context=context)
+
+    def _record_ws_exchange(self, label, pack):
+        trac_infos = ''
+        if 'raw_response' in label and 'request' in label:
+            trac_infos = (
+                u'Séquence Colis GLS:\n====================\n%s \n\n'
```
yvaucher, line 288: Is it wanted to have Séquence in french ?

yvaucher:
Only a little question otherwise :+1:

bealdav:
see https://github.com/OCA/carrier-delivery/pull/91

PR #62, opened 2015-06-09
Do not return duplicate packages when several moves have the same one

guewen:
Because it will print several labels for the same package.


pedrobaeza:
:+1:

yvaucher:
:+1:

PR #65, opened 2015-06-25
Add try/catch on pycountry import

mdietrichc2c:
Even if the module is not installed, the code from `delivery_carrier_label_gls` is called when the server is launched. I added some try/catches so that `import pycountry` does not fail if the egg is not present.

### DIFF
--- a/delivery_carrier_label_gls/report/label.py
+++ b/delivery_carrier_label_gls/report/label.py
@@ -19,15 +19,14 @@
 #
 ###############################################################################
 
+import logging
 from mako.template import Template
 from mako.exceptions import RichTraceback
 from .label_helper import AbstractLabel
 from .exception_helper import (InvalidAccountNumber)
 import httplib
 from unidecode import unidecode
-import logging
 import os
-import pycountry
 
 REPORT_CODING = 'cp1252'
 ERROR_BEHAVIOR = 'backslashreplace'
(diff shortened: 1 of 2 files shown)


delivery_carrier_label_gls/report/label.py
```diff
@@ -45,18 +43,25 @@
...
+        """
+        GLS_prefix = []
+        for elm in pycountry.countries:
+            GLS_prefix.append(str(elm.alpha2))
+        GLS_prefix[GLS_prefix.index('ME')] = 'CS'
+        return GLS_prefix
+
+    GLS_COUNTRIES_PREFIX = GLS_countries_prefix()
+except ImportError:
+    logger.warn(_('Please install python lib pycountry'))
```
yvaucher: You will need to update pot file for this change or remove translation as it is in logs I think it's ok.

yvaucher:
:+1:

pedrobaeza:
With new Transifex script it is not needed to regenerate pot

delivery_carrier_label_gls/report/label.py
```diff
@@ -45,18 +42,25 @@
...
+        GLS_prefix = []
+        for elm in pycountry.countries:
+            GLS_prefix.append(str(elm.alpha2))
+        GLS_prefix[GLS_prefix.index('ME')] = 'CS'
+        return GLS_prefix
+
+    GLS_COUNTRIES_PREFIX = GLS_countries_prefix()
+except ImportError:
+    logger.warn('Please install python lib pycountry')
+    GLS_COUNTRIES_PREFIX = False
```
pedrobaeza: This should be `[]` for being consistent with the type

mdietrichc2c:
@yvaucher @pedrobaeza both points are now fixed.

pedrobaeza:
:+1:

guewen:
@mdietrichc2c could you squash the 3 commits in one please?

mdietrichc2c:
@guewen it's now squashed.

guewen:
Thanks :+1:

PR #66, opened 2015-07-02
Add index on shipping_label.tracking_id

guewen:
Because this field is used for lookups.
Got an index scan of 0.033ms instead of a seq scan of 83.557ms.


yvaucher:
Definitely :+1:

mdietrichc2c:
:+1:

yvaucher:
I proceed with the merge, runbot being red is unrelated

PR #67, opened 2015-07-09
Fix error when user has no language

guewen:
Set default language to 'en_US' when the user has no language


pedrobaeza:
:+1: 

Thanks for the tests

charbeljc:
Hi @pedrobaeza could we have this merged ? Regards, and good week-end

pedrobaeza:
We need 2 approvals for merging, but as tests has been provided and the change is pretty clear, I fast-track.

charbeljc:
Yes  @pedrobaeza ! many thanks

PR #68, opened 2015-07-29
[FIX] Do not raise ImportError if user doesn't use module.

tafaRU:


delivery_carrier_label_postlogistics/postlogistics/web_service.py
```diff
@@ -28,6 +28,15 @@
 from openerp.tools.translate import _
 
 _compile_itemid = re.compile(r'[^0-9A-Za-z+\-_]')
+_logger = logging.getLogger(__name__)
+
+try:
```
pedrobaeza, line 33: As it's declared on external_dependencies, you should make this import on `init_connection` and the methods where they are needed, and let the module manager to avoid the installation of the module in case the library is missing. This way, you avoid the annoying warning each time the server with this module uninstalled is started.

yvaucher:
:+1:

lmignon:
:+1:

delivery_carrier_label_postlogistics/postlogistics/web_service.py
yvaucher, line 33: I merge this. As stated in other PR I prefer to keep imports on top.

pedrobaeza, line 33: OK, it was told that way, although the constant warning each time the server is started is not optimal.

yvaucher, line 33: Wasn't there an opened issue on Odoo for this unwanted load of the module when static folder is present?

pedrobaeza, line 33: Yeah, I think so, but I think it's staled for now

PR #72, opened 2015-10-15
[9.0][PORT] base_delivery_carrier_label

yvaucher:


yvaucher:
@bealdav A little review ?

bealdav:
It seems than improvements in v8 are not there like unit tests ?

yvaucher:
@bealdav Ok this PR is quite old I'll check if something has been merged in the mean time

yvaucher:
@bealdav Rebased on updated modules with last changes.

base_delivery_carrier_label/__openerp__.py
```diff
@@ -1,40 +1,22 @@
...
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
-
+# © 2013-2015 Yannick Vaucher (Camptocamp SA)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {'name': 'Base module for carrier labels',
- 'version': '8.0.1.2.0',
+ 'version': '9.0.1.2.0',
```
guewen: `s/9.0.1.2.0/9.0.1.0.0/`

base_delivery_carrier_label/models/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# © 2013-2015 Yannick Vaucher (Camptocamp SA)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
guewen: not sure there is a point to leave those lines in `__init__.py` files

base_delivery_carrier_label/models/stock.py
```diff
@@ -1,28 +1,59 @@
...
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# © 2012-TODAY David BEAL Akretion <http://www.akretion.com>.
+# © 2012-TODAY Sébastien BEAU Akretion <http://www.akretion.com>.
+# © 2013-2014 Yannick Vaucher (Camptocamp SA)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openerp import models, fields, api, exceptions, _
+from openerp.exceptions import Warning as UserError
```
guewen: you could directly import UserError: `from openerp.exceptions import UserError`

guewen:
`base_delivery_carrier_label/models/stock.py:6:1: F401 'exceptions' imported but unused`

guewen:
Error in the tests:

``` python

2016-04-26 15:32:36,690 23971 ERROR openerp_test openerp.addons.base_delivery_carrier_label.tests.test_get_weight: ERROR: test_get_weight (openerp.addons.base_delivery_carrier_label.tests.test_get_weight.TestGetWeight)

2016-04-26 15:32:36,690 23971 ERROR openerp_test openerp.addons.base_delivery_carrier_label.tests.test_get_weight: ` Test quant.package.get_weight and pack.operation.get_weight.

2016-04-26 15:32:36,690 23971 ERROR openerp_test openerp.addons.base_delivery_carrier_label.tests.test_get_weight: Traceback (most recent call last):

2016-04-26 15:32:36,690 23971 ERROR openerp_test openerp.addons.base_delivery_carrier_label.tests.test_get_weight: `   File "/home/travis/build/OCA/carrier-delivery/base_delivery_carrier_label/tests/test_get_weight.py", line 86, in test_get_weight

2016-04-26 15:32:36,690 23971 ERROR openerp_test openerp.addons.base_delivery_carrier_label.tests.test_get_weight: `     picking = self._generate_picking(products)

2016-04-26 15:32:36,690 23971 ERROR openerp_test openerp.addons.base_delivery_carrier_label.tests.test_get_weight: `   File "/home/travis/build/OCA/carrier-delivery/base_delivery_carrier_label/tests/test_get_weight.py", line 76, in _generate_picking

2016-04-26 15:32:36,690 23971 ERROR openerp_test openerp.addons.base_delivery_carrier_label.tests.test_get_weight: `     order.action_button_confirm()

2016-04-26 15:32:36,690 23971 ERROR openerp_test openerp.addons.base_delivery_carrier_label.tests.test_get_weight: ` AttributeError: 'sale.order' object has no attribute 'action_button_confirm'

```

guewen:
> AttributeError: 'sale.order' object has no attribute 'action_button_confirm'

It is `action_confirm` now

base_delivery_carrier_label/views/stock.xml
```diff
@@ -0,0 +1,68 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+  <data>
```
mdietrichc2c: `<odoo>`

base_delivery_carrier_label/models/__init__.py
bealdav: I'm also not sure.

yvaucher:
@alexis-via I don't find from where method `set_pack_weight` is called on model picking.

@bealdav @guewen 
Tests on get_weight are making trouble, a lot has changed to compute weight on 9.0. We now have a weight field already on `stock.quant.package`. Plus logistic units have disapeared, not sure how to replace them, should we add a second weight field on `stock.quant.package` to include the weight of the packaging ?

hparfr:
> Tests on get_weight are making trouble,

It's good: it's an highlight on breaking changes :)

> Plus logistic units have disapeared

Logistic unit weight is needed for the carriers who wants a total weight for the package.
And it's the responsibility of the package to include it his calculation of his own weight.

yvaucher:
@hparfr Thus have you any clue if in official addons of 9.0 we can still manage logistic units or if we need ... again ... to recreate an object they dropped ?

yvaucher:
Ok after more thinking, I will add a new field on `stock.quant.package` object named `total_weight` that allow to override the computed `weight` field value (if not defined value will still be computed as it is now otherwise we put total_weight in it).

Thus in case you weight your final package before delivery you will only have to set the value the balance gave you.

As I understand the weight value is important for the colissimo carrier.

hparfr:
Suggestion of simplifications : 
- remove generate_default_label method
- stop the partial use of packages and picking instances *

\* For instance, change

``` py
 def generate_labels(self, package_ids=None):
```

to 

``` py
def generate_labels(self):
```

> A list of package ids can be given, in that case it will generate  the labels only of these packages.

This feature adds complexity in the juggling between packages_id and i'm not sure about the business needs on this one.

What do you think @yvaucher ?

yvaucher:
@hparfr 

`package_ids` parameter is used in `delivery_carrier_label_dispatch` module.

And `generate_default_label` was a hook used in `delivery_carrier_label_default_webkit` which probably needs to be ported in Qweb.

What can be questionable is to keep both logic of label per picking or per package. What we could do is forcing the user to make packages thus we have only a single case to manage.

hparfr:
> And generate_default_label was a hook used in delivery_carrier_label_default_webkit which probably needs to be ported in Qweb.

`delivery_carrier_label_default_webkit` has not been ported in v8, we may drop support of it for v9 by removing `generate_default_label.` from base module. 

## 

> What we could do is forcing the user to make packages thus we have only a single case to manage.

👍

yvaucher:
Fixed weight calculs and tests. I removed the logic with logistic units providing a weight for the container.

README.md
```diff
@@ -13,13 +13,18 @@ You'll find:
  - Generation of shipping labels for specific carrier (PostLogistics, ...)
 
 [//]: # (addons)
+Available addons
+---------------
+addon | version | summary
+--- | --- | ---
+[base_delivery_carrier_label](base_delivery_carrier_label/) | 9.0.1.2.0 | Base module for carrier labels
```
lasley: 9.0.1.0.0

base_delivery_carrier_label/__openerp__.py
```diff
@@ -1,40 +1,22 @@
...
-          'res_config_view.xml',
-          'security/ir.model.access.csv',
-          ],
+ 'data': [
+     'views/delivery.xml',
+     'views/stock.xml',
+     'views/res_config.xml',
+     'security/ir.model.access.csv',
+ ],
  'tests': [],
```
lasley, line 18: Remove empty key

base_delivery_carrier_label/models/delivery.py
```diff
@@ -1,24 +1,7 @@
...
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# © 2012-TODAY Sébastien BEAU Akretion <http://www.akretion.com>.
+# © 2014 Yannick Vaucher (Camptocamp SA)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openerp import models, fields, api
```
lasley: Would you mind splitting these models into one per file?

base_delivery_carrier_label/models/stock.py
```diff
@@ -1,40 +1,102 @@
 # -*- coding: utf-8 -*-
```
lasley: Would you mind splitting these models into one per file?

README.md
pedrobaeza: This it automatically done by a bot, so there's no need to change it

lasley: TIL - Hooray automation!

base_delivery_carrier_label/models/delivery.py
yvaucher: ~~Can you explain why should I do that?~~
~~I prefer to keep it that way as it is in most OCA code base~~

Ok I thought you wanted me to split the imports.

Those models are completely related. Thus I prefer to keep them in the same file.

base_delivery_carrier_label/models/stock.py
yvaucher: IMO this is readable enough this way.

base_delivery_carrier_label/models/delivery.py
lasley: Hi @yvaucher - OCA guidlines say individual files though which is mentioned in the differences section between OCA and Odoo guidelines, as well as a few other places - https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#differences-with-odoo-guidelines

yvaucher:
Splitted python files by model

yvaucher:
I proceed with this PR as on itself it has no use and is only a dependency for others thus unlocking dev for other modules.

PR #79, opened 2015-12-18
7.0 small usability improvements

alexis-via:
Add view on carrier.account
Show "generate labels" when outgoing picking is done
Fix deprecated warning


yvaucher:
travis errors are not related.

So basically :+1: 

I just have a concern about the state on which the button needs to be available. Because the module was created with in mind that preparing the package includes putting the label on it and done state mean it has been sent, not that it needs to be labeled. While you propose to print the label only once the package is prepared. Maybe we are missing a state?

PR #87, opened 2016-05-31
[7.0] postlogistics v7.0.1.2

yvaucher:
This version contains the following improvements:
- Add an option to use integration endpoint.
- Add an option to let Odoo generate the tracking numbers based on pack count and picking name.
- Implement delivery instructions ZAW and cash delivery adding the required fields on picking.
- Improve error handling for parameter for layout (eg A6) output format (eg PDF) and resolution (eg 300).
- When generating labels for packs an agregation of all tracking numbers of the packs will be written separated by ';' in field carrier_tracking_ref on the picking


delivery_carrier_label_postlogistics/postlogistics/web_service.py
```diff
@@ -44,12 +61,29 @@ def __init__(self, company):
...
-            username=company.postlogistics_username,
-            password=company.postlogistics_password)
-        self.client = Client(
-            company.postlogistics_wsdl_url,
-            transport=t)
+        if company.postlogistics_test_mode:
+            # We must change location for namespace as suds will take the wrong
+            # one and find nothing
+            # In order to have the test_mode working, it is necessary to patch
+            # suds with https://fedorahosted.org/suds/attachment/ticket/239/suds_recursion.patch # noqa
```
lepistone, line 69: Do you know if this was merged in the suds-jurko fork? If not maybe the patch could be proposed there?

lepistone:
Thanks for your work @yvaucher. Have you investigated the travis failure? It doesn't seem to be related to your change but still it would be great to understand what is going on.

delivery_carrier_label_postlogistics/postlogistics/web_service.py
yvaucher, line 69: This module is already based on suds jurko fork. This is not the official patch in there dev branch, but it works. suds-jurko has another patch which didn't worked well in this case.

The patch is required mostly to develop the module and requires to have access to a test account which must be asked so it should touch few users.

Ok I'll add a note in `__openerp__.py` (no readme.rst in 7.0) and add a copy of the patch to apply in a patch folder.

yvaucher:
I had a look at travis failure but I have no idea why the last one fails. I know it will always fails as unrelated modules have PEP8 issues plus it was already failing on 7.0 branch. But why travis can't even generate an instance of openerp v7.0, I can't tell. A lot has changed in MQT, the definition of .travis.yml might be deprecated ?

Anyway why it is failing is currently unrelated to this PR.

delivery_carrier_label_postlogistics/postlogistics/web_service.py
lepistone, line 69: I think a note on `__openerp__.py` would be perfect, thanks!

lepistone:
@yvaucher please give a look to #88 thanks!

yvaucher:
@lepistone Thanks for the fix on travis. I just rebased on it.

delivery_carrier_label_postlogistics/postlogistics/web_service.py
yvaucher, line 69: Note and patch added in the module.

lepistone:
Thanks for the update @yvaucher! 👍

PR #88, opened 2016-06-01
modern CI setup from MQT template

lepistone:


pedrobaeza:
:+1: if Travis doesn't complain

guewen:
:+1:

.travis.yml
```diff
@@ -1,39 +1,45 @@
+language: python
 sudo: false
-cache: pip
+cache:
+  apt: true
+  directories:
+    - $HOME/.cache/pip
+
+python:
+  - "2.7"
```
yvaucher, line 9: Why not keeping the version just bellow the language ?

yvaucher, line 9: Okay this is in the template. I guess we just want to change randomly the order sometimes...

yvaucher:
Merging as it fixes travis

Thanks

PR #89, opened 2016-06-09
[REF] make selection function overwritable

tremlin:
This PR makes the selection function of two fields in `base_delivery_carrier_files` overwriteable.


yvaucher:
Should it be already overwritable ?

Anyway it doesn't hurt to be explicit :+1:

lasley:
For clarity, overwriteable here means in terms of inheritance due to use of a string method name instead of the method reference in the field declarations. All for it 👍

yvaucher:
Thanks

PR #90, opened 2016-06-10
Proposal - Handling of multiple deliveries, rate quotes & labels for one picking

lasley:
Hi Friends,

From a high level, this module provides for multiple carrier deliveries on one stock picking, multiple rate quotes associated with each delivery, and a label per delivery (instead of one per picking). 

The additional model that provides these is the `stock.delivery.group`, which is a O2M on `stock.picking`. Think of the delivery group as a picking specifically tied to a package. The use case here being a situation where you are able to pick all of the items, but not fit them in the same package, thus needing multiple packages and delivery labels within the same picking context.

As an added feature, there are delivery package inherits from quant.pack, and templates that allow for the customization of standardized box sizes for shipping. There is support for multiple weight UOMs vs Odoo core forcing kg.

This is definitely a rough outline at this point, it's almost certainly named wrong, and will likely need to be broken into a few modules. I would like to also use the `base_delivery_carrier_label` instead of my home grown one, but this does the job for the moment.

This does break some things, namely the built in carrier_tracking_ref & associated link. It is also a bit clunky in terms of workflow & UI. I haven't bothered solving those issues yet, as I wanted to make sure that I am somewhat on the right track here before going too deep down the rabbit hole. 

The goal here, aside from providing a more powerful shipping context, was to provide a base for external carrier connectors. Working on a prototype for EasyPost right now based on this, and it works pretty well https://github.com/laslabs/odoo-connector-easypost/tree/9.0/connector_easypost

Please let me know if I am on the right track, and if this is something wanted in OCA.


pedrobaeza:
Isn't this similar to https://github.com/OCA/stock-logistics-workflow/tree/9.0/picking_dispatch?

lasley:
Hmmm looks like that hits a lot of my use case in terms of the grouping, then would just need to add in the label, rate, and tracking logic. Looks like I'll be upgrading that module real soon, thanks @pedrobaeza 

Can you think of anything off the top of your head that would provide the ability to save multiple rate quotes (before label generation) & some sort of tracking object in order to follow a delivery during it's stages? I figure it's worth asking, because I totally missed this other module and I would much rather not reinvent wheels 😉

pedrobaeza:
The only module I remember for multiple rate quotes is https://github.com/odoomrp/odoomrp-wip/tree/8.0/delivery_multi_expense, but it's more for having several rate concepts.

A module delivery_tracking would be very interesting indeed. I would do it one directly associated with the pickings, and another delivery_tracking_dispatch as glue module with picking_dispatch to associate delivery tracking objects to dispatch object (in case of grouping).

lasley:
Interesting- the delivery_multi_expense actually hits another side of the project, which is handling the vendor payments. I need to reverse them though, as it is typically a grouped payment for a batch of labels.

I like the idea of linking both against the picking then creating the separate module for the dispatches, so we could provide that functionality without the need for the workflow dependency if not required.

I have somewhat of an idea of the delivery_tracking in this PR actually - it's just a model, but should provide a general idea of what I'm thinking - https://github.com/laslabs/carrier-delivery/blob/e59a7cf00370fd665861940678e5d4146ec9f8f6/stock_delivery/models/stock_delivery_operation.py 

The goal with the delivery_tracking will be to allow for web hooks, or polling, in external connector interfaces to allow for updating of the tracking statuses. There may be some workflows, like if the partner is a carrier themselves, where this would make sense without external interfaces though.

pedrobaeza:
Although not directly related, look what we have made for tracking mails, also being based on standard object, providing a base object for reflecting the status, and with several later modules linking to other Odoo parts: https://github.com/OCA/social/pull/67. The idea of other modules that uses webservices for getting info is wonderful.

lasley:
Nice, this is basically the same thing just in a different context! It looks like everything is too unique between the two in order to abstract anything, which saddens me, but this is some solid reference in terms of structure/flow.

As always @pedrobaeza thanks for your advice - I will begin glueing these modules together and adding missing functionality to meet my project specs 👍 

I too am I excited about using external web services for info, I think this will be the next step in providing true ubiquity to Odoo. Odoo can basically do everything we need it to with minimal effort, now we need it to talk to everything we need it to 😉

rafaelbn:
@lasley maybe you can migrate that modules to v9, here in OCA :smile:

lasley:
@rafaelbn - Definitely whatever we end up with on my end will be submitted this direction 👍

PR #73, opened 2015-10-15
[9.0][PORT] delivery_carrier_label_postlogistics port to 9.0

yvaucher:
- Reactivate
- Move files in views and models
- Adapt XML views
- Update README.rst
- Move licenses under it's own menu

Depends on
- [x] #72

PR #92, opened 2016-06-30
9.0 postlogistics 1.1

yvaucher:
Adds changes of #87 

Depends on #73 and #87


yvaucher:
merging to be able to make version 10.0

PR #93, opened 2016-07-04
do not choose weight instead of weight_net

bealdav:
in base_delivery_carrier_label

There is no way to the user to know which field is taken account, then we should take both.
If think it's better than `max(weight, weight_net)`


hparfr:
👍

yvaucher:
:+1: Agreed it's good that way, the weight with packaging has the priority over the net weight.

PR #98, opened 2016-07-13
[FIX] prevent create deposit slip in tree/form view

bealdav:
There is a wizard to create deposit.slip, so it improve usability.

Before this PR, it's error prone.


bealdav:
ping @yvaucher

yvaucher:
:+1:

pedrobaeza:
:+1:

PR #103, opened 2016-09-26
Carrier label: Missing acl for stock user/manager

cyrilgdn:


yvaucher (approved):

base_delivery_carrier_label/security/ir.model.access.csv
```diff
@@ -7,3 +7,6 @@ access_shipping_label_user,shipping.label user,model_shipping_label,stock.group_
 access_shipping_label_manager,shipping.label manager,model_shipping_label,stock.group_stock_manager,1,1,1,1
 access_carrier_account_salesman,carrier.account.salesman,model_carrier_account,base.group_sale_salesman,1,0,0,0
 access_carrier_account_sale_manager,carrier.account.sale.manager,model_carrier_account,base.group_sale_manager,1,1,1,1
+access_delivery_carrier_option_stock_user,delivery.carrier.option stock_user,model_delivery_carrier_option,stock.group_stock_user,1,1,1,1
```
tafaRU, line 10: I think here acl should be 1,0,0,0

tafaRU (changes requested):

base_delivery_carrier_label/security/ir.model.access.csv
yvaucher: Nope this is right.

Template option is the main configuration on delivery method and when setting delivery method on the picking, it will copy template option from delivery method as options on the picking. Thus a user needs to be allowed to select a delivery method and create options.

tafaRU: Ok, I approve then. Thanks for the explanation!

tafaRU (approved):

simahawk (approved):

PR #109, opened 2016-10-21
[IMP] add pciking_type_id on deposit_slip

bguillot:
Hi,

This PR add picking_type_id on deposit_slip. 
It is used to filter the picking linked to the deposit. And can be used to have the link between the deposit and the warehouse.

### DIFF
--- a/delivery_carrier_deposit/wizard/deposit.py
+++ b/delivery_carrier_deposit/wizard/deposit.py
@@ -35,6 +35,13 @@ class DeliveryDepositWizard(models.TransientModel):
     def _get_carrier_type_selection(self):
         return self.env['delivery.carrier']._get_carrier_type_selection()
 
+    @api.model
+    def _get_default_picking_type(self):
+        warehouse = self.env['stock.warehouse'].search(
+            [('company_id', '=', self.env.user.company_id.id)])
+        return (warehouse and warehouse[0].out_type_id
+                and warehouse[0].out_type_id.id or None)
+
     carrier_type = fields.Selection(
         '_get_carrier_type_selection', string='Delivery Method Type',
         required=True, help="Carrier type (combines several delivery "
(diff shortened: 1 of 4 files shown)


delivery_carrier_deposit/wizard/deposit.py
```diff
@@ -40,23 +40,33 @@ def _get_carrier_type_selection(self):
...
     def _prepare_deposit_slip(self):
         return {
             'carrier_type': self.carrier_type,
             'company_id': self.env.user.company_id.id,
+            'picking_type_id': self.picking_type_id.id,
             }
 
+    @api.model
+    def _get_deposit_pickings(self):
+        return self.env['stock.picking'].search([
```
bealdav, line 65: _get_deposit_domain() could be more easy to override by 2 other modules with append() than this method

delivery_carrier_deposit/wizard/deposit.py
```diff
@@ -40,23 +40,33 @@ def _get_carrier_type_selection(self):
         required=True, help="Carrier type (combines several delivery "
         "methods). Make sure that the option 'Deposit Slip' is checked on "
         "the delivery methods that have this carrier type.")
+    picking_type_id = fields.Many2one(
+        comodel_name='stock.picking.type',
+        string='Picking Type',
+        required=True)
```
bealdav: Maybe a default value with the first picking_type_id of out type in the company could be benefit here

delivery_carrier_deposit/wizard/deposit.py
bguillot: That's right but only change the domain is very restrictive in my opinion.

You could want to get pickings with very specific search from sale_order for instance.

And I see two case of overriding
1. for a **carrier** in that case you want to test depending on the carrier_type and you have your own domain 

`if self.carrier_type  == 'carrier_type':
     domain = []
 else:
     super ..`

Because you maybe don't want the other carrier addition in the domain

   2 a specific domain for a customer so you will maybe not even call the super() to make your specific search

In the two case, changing the domain or directly searching the pickings ends with the same amount of searchs but more flexible to return pickings instead of domain.

This is what I think but if you think it's more safe to only return the domain, I agree to change for your idea.

bealdav:
fix conflict and then :+1: 
cc @yvaucher

yvaucher:
fixed conflict and merged https://github.com/OCA/delivery-carrier/commit/3dd2ebf5885d1207462f8152bd03edd45f261637

PR #110, opened 2016-11-12
IMP add a dedicate tree view to deposit.slip picking_ids field

bealdav:
## Old view
av
No weight or package here 

## New view
af

User now focus on main info

ping @yvaucher @alexis-via @hparfr @guewen


pedrobaeza (approved):
Good idea!

hparfr:
👍

yvaucher (approved):

yvaucher:
Thanks

PR #113, opened 2016-12-10
[9.0][ADD] delivery_multi_destination

pedrobaeza:
Multiple destinations for the same delivery method
==================================================

Module `delivery` in version 8 allows to set different price rules depending on the destination. This is what is called a delivery grid.

In version 9, for simplifying delivery methods, Odoo has plained the structure, lowering destinations at delivery method level, and removing delivery grid model.

This is not usable when you have different prices according the destination of your delivery.

This module restores the same concept, reusing the same model for nesting several "children" delivery methods, one per possible destination. It has been designed to reuse all possible extensions to the base delivery, without the need to create a glue module for having multiple destinations.

This module also handles if you're migrating from version 8 and you had `delivery` module installed, to keep the delivery grids.

Installation
============

If you installed the module on a version 8 migrated database, some operations will be done for recovering delivery grids. If so, you need to have **openupgradelib** library installed.

Configuration
=============

To configure delivery methods with multiple destinations:

* Go to Inventory > Configuration > Delivery > Delivery Methods
* Create or edit an existing record.
* Select "Destination type" = "Multiple destinations".
* Introduce a line for each destination in the new tab "Destinations"
* Lines have priority, so you have to put first the lines with more restricted
  destinations.

Usage
=====

* When using the delivery method in a Sales order, delivery address will be used for computing the delivery price according introduced destinations.

@Tecnativa


pedrobaeza:
@luismontalba please review this one that you migrated to 10.0

@rafaelbn please review also

luismontalba (approved):

pedrobaeza:
As this has been merged already in 10.0 with equal code base, I merge this one as well.

PR #115, opened 2017-01-06
[9][base_delivery_carrier_label]Fix package view

florian-dacosta:
The weight field is already added in delivery module so it is duplicated in the view.
https://github.com/OCA/OCB/blob/9.0/addons/delivery/views/delivery_view.xml#L224


yvaucher (approved):
Nice catch

pedrobaeza (approved):

PR #119, opened 2017-03-24
[ADD][9.0] stock_picking_delivery_rate

woodbrettm:
* Supersedes oca/stock-logistics-workflow#238
* @tedsalmon
* @lasley 
* @pedrobaeza 
* @obulkin

Review comments have been attended to!

PR #122, opened 2017-03-28
[9.0][ADD] sale_delivery_rate: Sale order delivery rates

tedsalmon:
This is PR is a move from https://github.com/OCA/stock-logistics-workflow/pull/297

Original PR text:

This PR is dependent on #238

* Add `delivery.carrier.rate` model that inherits from stock.picking.rate but maps to `sale.order records `instead. It includes a method that produces `stock.picking.rate` records based on delivery.carrier.rate records

* Extend `sale.order` model as needed to reflect new relation
* Extend `delivery.carrier` model to allow sale order delivery prices to be calculated based on delivery carrier rates
* Overload action_confirm in `stock.move` to update stock pickings with rate quotes when they are generated based on sale orders with quotes
* Add access controls, demo data, and unit tests as needed

cc @lasley


lasley (approved):
code LGTM, but please switch LasLabs icon for OCA

obulkin:
Thanks, @tedsalmon. Just to clarify for anyone looking at this PR for the first time, the dependency listed above as 238 has also been moved to this repo (#119) and has been merged.

lasley:
@lukebranch - this should provide the website rates you were looking for in LasLabs/odoo-connector-easypost#3 if you would like to review

lukebranch:
@lasley , that's awesome news, thank you!

I'll definitely review this as soon as possible.

sale_delivery_rate/README.rst
```diff
@@ -0,0 +1,75 @@
...
+===================
+Sale Delivery Rates
+===================
+
+This module extends the notion of delivery rate quotes introduced in 
+``stock_picking_rate`` to sale orders. The new quote type is compatible with the 
+``Set price`` button used on sale orders to calculate a delivery price. Also, 
+when a sale order with such rate quotes is confirmed and results in the 
+creation of a stock picking, the picking will be updated with equivalent stock 
+picking delivery quotes.
```
woodbrettm: Could you simplify this up a bit? Having a hard time following.

sale_delivery_rate/demo/delivery_carrier.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+    Copyright 2017 LasLabs Inc.
+    License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+-->
```
woodbrettm: xml headers should be that but like so:
```xml

<?xml version="1.0" encoding="utf-8"?>
<!-- Copyright <YEAR(S)> <AUTHOR(S)>
     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
```

sale_delivery_rate/demo/delivery_carrier.xml
```diff
@@ -0,0 +1,20 @@
...
+
+<odoo>
+    <record id="demo_carrier_1" model="delivery.carrier">
```
woodbrettm: nice I like your idea of adding demo in front of the id. I think I'll start doing that too :)

sale_delivery_rate/models/delivery_carrier.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from openerp import api, fields, models
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = 'delivery.carrier'
```
woodbrettm, line 9: Should be empty line between class definition and the _inherit.

sale_delivery_rate/demo/delivery_carrier_rate.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+    Copyright 2017 LasLabs Inc.
+    License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+-->
+
+<odoo>
+    <record id="demo_car_rate_1" model="delivery.carrier.rate">
```
woodbrettm: Should be an empty line between the odoo tag and record tags.

sale_delivery_rate/models/delivery_carrier.py
```diff
@@ -0,0 +1,32 @@
...
+    delivery_type = fields.Selection(
+        selection_add=[('base_on_rate', 'Based on Rates')],
+    )
+
+    @api.model
+    def base_on_rate_get_shipping_price_from_so(self, sale_orders):
+        prices = []
+
+        for sale_order in sale_orders:
+            order_rate = self.env['delivery.carrier.rate'].search([
```
woodbrettm, line 20: is `order_rate` guaranteed to be only 1 record with the provided domains? May be best to add `limit=1` just in case.

woodbrettm (changes requested):
Mostly just small things and the icon needs to be changed to oca as @lasley mentioned. Other than that looks great!

sale_delivery_rate/demo/delivery_carrier_rate.xml
obulkin: Is this an OCA rule? I don't see anything about this in the style guide and kind of prefer the tighter spacing myself.

sale_delivery_rate/demo/delivery_carrier.xml
obulkin: I've found that it can be kind of a pain to figure out whether an XML ID corresponds to a demo record or a regular one, so I've been trying to make that clear.

sale_delivery_rate/models/delivery_carrier.py
obulkin: My linter seems to be happy either way, and I like the tighter spacing a bit more. AFAIK, PEP 8 only requires space around class and method blocks, not inside them.

sale_delivery_rate/models/delivery_carrier.py
obulkin: Good point. There's no business reason to have multiple rates for the same sale order and carrier, but the field definitions don't actually prohibit this, so adding a limit would protect against this edge case. On the other hand, it might actually be best to let this scenario generate an exception since the connector shouldn't be creating it in the first place.

@lasley - What would you recommend here?

sale_delivery_rate/README.rst
obulkin: I can try rewording this, but I'm not sure I can do much better. Can you try out the sale order -> stock picking flow on Runbot and let me know if it's still unclear?

sale_delivery_rate/models/delivery_carrier.py
lasley: I'd add a constraint in `delivery.carrier.rate` for `UNIQUE(sale_order_id, service_id)`

lasley (commented):
Please switch LasLabs logo icon for OCA

sale_delivery_rate/models/delivery_carrier.py
lasley: Dev preference

sale_delivery_rate/demo/delivery_carrier.xml
lasley: I like too. Maybe submit this suggestion to our Contribution docs and subsequently the template module (if accepted in docs)?

lasley: Although I think `demo` should be a suffix instead of prefix.

lasley: `model_name_demo_something`

sale_delivery_rate/demo/delivery_carrier_rate.xml
woodbrettm: The oca module template seems to show them separated with empty lines, but I guess preference by developer on what they do.

sale_delivery_rate/models/delivery_carrier.py
obulkin: Ah. Good idea. I'll do that.

sale_delivery_rate/demo/delivery_carrier.xml
obulkin: Ok. I'll switch these up. I think I've been using an older LasLabs module template that differs in a few small ways like this.

sale_delivery_rate/demo/delivery_carrier.xml
obulkin: Sounds good. I'll do that.

obulkin:
Thanks @lasley and @bmw95. I've pushed up a new commit addressing these comments.

lasley (approved):

woodbrettm (approved):
LGTM!

PR #131, opened 2017-04-26
[9.0] Use correct syntax to add options with onchange on stock picking

mdietrichc2c:
Small fix for an issue I found : if the carrier is edited manually on a stock picking, the values given to the one2many field "option_ids" are a RecordSet.

It works fine when editing, but afterwards, when the picking is written, the values are with the format `{'option_ids': [(1, <ID of the option>, <value of the option>)]}`, not `{'option_ids': [(6, False, [<ID of the option>])]}`, which leads to the values not being written in the picking.


guewen (approved):
tested

yvaucher (approved):
Thanks

PR #135, opened 2017-05-26
[10.0][MIG] sale_delivery_rate

woodbrettm:
Hi there,

Migration of sale_delivery_rate from 9.0 to 10.0.

Cheers,
Brett

@lasley @obulkin


woodbrettm:
Runbot is failing because of model.access rules on a transient model in base_delivery_carrier_label [here](https://github.com/OCA/delivery-carrier/blob/10.0/base_delivery_carrier_label/security/ir.model.access.csv#L13). (line 14 as well). Should those be removed from that module?

Other than that, ready for review!

lasley:
@bmw95 we can ignore the Runbot warning here, because it is also failing for the same thing on v10. Might as well submit a quick PR though to fix now that you've investigated

lasley (approved):
Thanks @bmw95

woodbrettm:
base_delivery_carrier_label security fix has now been submitted to #136

woodbrettm:
nice this is good to go. Rebased to 10.0 now that #136 is merged.

lasley:
@obulkin please review

sale_delivery_rate/models/delivery_carrier_rate.py
```diff
@@ -25,9 +29,9 @@ class DeliveryCarrierRate(models.Model):
...
         stock_picking.ensure_one()
-        recordset_data = self.copy_data()
-
-        for record_data in recordset_data:
-            del record_data['sale_order_id']
-            record_data['picking_id'] = stock_picking.id
-            self.env['stock.picking.rate'].create(record_data)
+        for record in self:
+            recordset_data = record.copy_data()
+            for record_data in recordset_data:
```
obulkin: There's no need for an inner loop here now that you're calling `copy_data` against a single record. The result will be a single dictionary inside a list, so you can just do `record_data = record.copy_data()[0]` instead.

obulkin (changes requested):

sale_delivery_rate/models/delivery_carrier_rate.py
woodbrettm: oh yeah weird. Didn't catch that while upgrading, thanks!

woodbrettm: Hmm. Actually no this is correct. The code is ensuring the arg is a singleton: `stock_picking.ensure_one()`not `self.ensure_one()`. I think I remember coming across this earlier. Removing the loop causes multiple records to be copied at a time.

woodbrettm: oh got it I misunderstood. don't need to loop through the record_data. For some reason I was thinking "don't loop through the records . . ."

obulkin (approved):

PR #138, opened 2017-09-26
[10.0][add] delivery_multi_destination

luismontalba:
Multiple destinations for the same delivery method
========================================

This module allows to set different price rules depending on the destination.

This module restores the concept of delivery grid, reusing the same model for nesting several "children" delivery methods, one per possible destination.
It has been designed to reuse all possible extensions to the base delivery, without the need to create a glue module for having multiple destinations.

Configuration
-----------------
To configure delivery methods with multiple destinations:

* Go to Inventory > Configuration > Delivery > Delivery Methods
* Create or edit an existing record.
* Select "Destination type" = "Multiple destinations".
* Introduce a line for each destination in the new tab "Destinations"
* Lines have priority, so you have to put first the lines with more restricted destinations.

Usage
--------
* When using the delivery method in a Sales order, delivery address will be used for computing the delivery price according introduced destinations.

cc @Tecnativa


delivery_multi_destination/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "license": "AGPL-3",
+    "installable": True,
+    "depends": [
+        "delivery",
+    ],
+    "data": [
+        "views/delivery_carrier_view.xml",
+    ],
+    # "uninstall_hook": "uninstall_hook",
+    # "post_init_hook": "post_init_hook",
```
chienandalu: Can't you drop this lines?

chienandalu (approved):
Tested on runbot :+1:

delivery_multi_destination/__manifest__.py
luismontalba: Done

rami-wafaie (approved):
Looks fine for me too 
:+1:

lasley (approved):
Thanks @luismontalba

lasley:
Can you squash your two commits please?

luismontalba:
@lasley , my two commits squashed

lasley:
Thanks @luismontalba

PR #140, opened 2017-10-24
[IMP] stock_picking_delivery_rate: Add button + action

lasley:
* Add an action to compute rates and optionally delete the old ones
* Add a button in the stock picking header to allow for manual rate computation


lasley:
Yeah so now that I've got all the proper stuff rigged up with carriers, this is pointless. Closing

PR #141, opened 2017-10-24
[ADD] stock_picking_delivery: Module to create deliveries 

lasley:
* Create a new module to add a wizard allowing for dispatch generation with a type of box

This came from a lot of back and forth on how shipment should actually work, both with customers and OCA. The high level history:
* [The original](https://github.com/OCA/delivery-carrier/pull/90) - Some of which became `stock_picking_delivery_rate`, which is a dependency of this module
* [stock_picking_package_info](https://github.com/OCA/stock-logistics-workflow/pull/247) - Covered what we needed, but ended up just adding complexity and merging too many concepts

So now we have this module, `stock_picking_delivery`. It adds a new concept of a `delivery.package` that is meant to be used to reference the types of boxes that are used during shipping. We have ultimately decided that trying to lump this concept in with the pre-existing `product.packaging` does not work in practice. 

This also adds some basic UOM conversions for the deliveries, because we're weird in the US and use our own units. A delivery package can also optionally be assigned to a product in order to allow for purchase workflows.

This is still a WIP as we implement with customers.


lasley:
Going with [`stock_package_info`](https://github.com/OCA/stock-logistics-workflow/pull/384) after all

PR #143, opened 2017-11-07
[IMP] stock_picking_delivery_rate: Add reference & cancellation

lasley:
This PR adds a few features to `stock_picking_delivery_rate`
* Add a reference to the associated rate on the purchase order lines
* Add a process for the cancellation of rates and their related purchase orders
* Add an optional m2o to link a rate with a quant package

At the moment, cancellation is at the purchase order level. I don't think this is totally ideal, but I'm not sure of a good strategy of processing a cancellation for just one purchase order line item. Open to ideas 😄 

Depends on:
- [ ] https://github.com/OCA/purchase-workflow/pull/474

### DIFF
--- a/oca_dependencies.txt
+++ b/oca_dependencies.txt
@@ -1,6 +1,8 @@
 # list the OCA project dependencies, one per line
 # add a github url if you need a forked version
 partner-contact
+# https://github.com/OCA/purchase-workflow/pull/474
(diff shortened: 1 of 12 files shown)


rousseldenis:
Closing this as I think no more work will be done. Please feel free to reopen it.

PR #148, opened 2017-12-20
[FIX+IMP] delivery_multi_destination: Don't show children + demo

pedrobaeza:
* Don't show children carriers on many2one selections
* Don't search by default children carriers
* Add demo data

@Tecnativa

### DIFF
--- a/delivery_multi_destination/views/delivery_carrier_view.xml
+++ b/delivery_multi_destination/views/delivery_carrier_view.xml
@@ -11,7 +11,7 @@
         <field name="arch" type="xml">
             <field name="delivery_type" position="before">
                 <field name="destination_type"
-                       invisible="context.get('show_children_carriers', False)"
+                       invisible="context.get('show_children_fields', False)"
                 />
             </field>
             <xpath expr="//field[@name='fixed_price']/ancestor::page"
@@ -27,7 +27,7 @@
                       attrs="{'invisible': [('destination_type', '!=', 'multi')]}"
                 >
                     <field name="child_ids"
-                           context="{'show_children_carriers': True}"
+                           context="{'show_children_fields': True}"
                     />
                 </page>
             </notebook>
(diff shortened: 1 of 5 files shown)


chienandalu (commented):
Shouldn't by default these be hidden?
image

pedrobaeza:
Yes, they should, but in local they were...

delivery_multi_destination/views/delivery_carrier_view.xml
```diff
@@ -41,23 +41,23 @@
...
-                       invisible="not context.get('show_children_carriers', False)"
+                       invisible="not context.get('show_children_fields', False)"
                 />
             </field>
         </field>
     </record>
 
     <record id="delivery.action_delivery_carrier_form" model="ir.actions.act_window">
-        <field name="domain">[('parent_id', '=', False)]</field>
+        <field name="context">{'show_children_carriers': True}</field>
```
chienandalu: When the menu option is hit this context is passed, and the default view is tree, so the children are visible in the tree. So I tried to set it to `False` here and to `True` in the form field that loads the tree.

delivery_multi_destination/views/delivery_carrier_view.xml
```diff
@@ -27,7 +27,7 @@
                       attrs="{'invisible': [('destination_type', '!=', 'multi')]}"
                 >
                     <field name="child_ids"
-                           context="{'show_children_carriers': True}"
+                           context="{'show_children_fields': True}"
```
chienandalu, line 30: Here a try to set `show_children_carriers` to `True` but seems that the action context prevails.

chienandalu (commented):
I tried it locally and it behaves the same way. I tried to get it right with no luck. Seems to be some context priority issues.

delivery_multi_destination/views/delivery_carrier_view.xml
pedrobaeza: Solution is to keep both, domain and context. I have done it.

luismontalba (approved):
Tested on runbot

chienandalu (approved):
Perfect!

PR #149, opened 2018-01-09
[9.0] Fix delivery_carrier_deposit : Make picking removal possible again in deposit slip

florian-dacosta:
In version 9.0, we can't remove picking from deposit slip, because Odoo try to delete it.
In version 8.0 and before it was possible, I guess the many2many widget on one2many field was doing the job.

In version 9.0 and after, we need to add an option so Odoo just remove the link with the picking and does not try to delete the picking.

@bealdav @hparfr @EBII

### DIFF
--- a/delivery_carrier_deposit/views/stock_view.xml
+++ b/delivery_carrier_deposit/views/stock_view.xml
@@ -76,7 +76,7 @@
                     <field name="number_of_packages"/>
                 </group>
                 <group name="pickings">
-                    <field name="picking_ids" nolabel="1" widget="many2many"/>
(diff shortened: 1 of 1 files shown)


hparfr (approved):

bealdav (approved):
It's an undocumented behavior in v10 only used 3 times :-( in core.
In v10 there is related code in js to drive the behavior.

In v11 js code disappeared but xml stay as dead code

florian-dacosta:
@bealdav I did a quick test, and yes, it seems that v11 behavior is the same as v8 and before.
(No need to specify the option no delete, Odoo won't delete it if many2many widget is set.)
So this options is only needed for version 9 and 10.

yvaucher (approved):

PR #152, opened 2018-02-27
[ADD] delivery_auto_refresh: Auto-refresh delivery price in sales orders

pedrobaeza:
Auto-refresh delivery price in sales orders
===========================================

This module automates the delivery price handling for the following cases:

* If you change any line in your draft sales order (SO), when saving, the delivery price will be adjusted without having to click on "→ Set price".
* If specified in the system parameter, the delivery line can be also auto-added when creating/saving.
* If you deliver a different quantity than the ordered one, the delivery price is adjusted on the linked SO when the picking is transferred.

Configuration
=============

* Activate developer mode.
* Go to *Settings > Technical > Parameters > System Parameters*.
* Locate the setting with key "delivery_auto_refresh.auto_add_delivery_line" or create a new one if not exists.
* Put a non Falsy value (1, True...) if you want to add automatically the delivery line on save.

Known issues / Roadmap
======================

* After confirming the sales order, the price of the delivery line (if exists) will be only updated after the picking is transferred, but not when you might modify the order lines.
* There can be some duplicate delivery unset/set for assuring that the refresh is done on all cases.
* On multiple deliveries, second and successive pickings update the delivery price, but you can't invoice the new delivery price.

cc @Tecnativa

### DIFF
--- a/delivery_auto_refresh/README.rst
+++ b/delivery_auto_refresh/README.rst
@@ -1,0 +1,81 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+===========================================
+Auto-refresh delivery price in sales orders
+===========================================
+
+This module automates the delivery price handling for the following cases:
+
+* If you change any line in your draft sales order (SO), when saving, the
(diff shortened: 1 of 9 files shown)


carlosdauden (approved):
Code LGTM

luismontalba (approved):
Tested on runbot

PR #128, opened 2017-04-20
[10.0] migrate delivery_carrier_label postlogistics

yvaucher:
cherry pick of commits from 9.0

Plus

* Simplify signature of soap connector by removing company
* As carrier inherits from product, change postlogistics partner as
  a service product

Specific changes to v 10.0 are in https://github.com/yvaucher/carrier-delivery/commit/abfa3395b1666f41b7673f4f24d8878a6e5f2dd3

PR #155, opened 2018-05-28
[11.0][MIG] delivery_carrier_label_postlogistics

grindtildeath:
Based on #128 
Depends on : 
 - [x] #156 
 - [x] https://github.com/OCA/server-tools/pull/1261


delivery_carrier_label_postlogistics/models/delivery.py
```diff
@@ -109,18 +109,20 @@ class DeliveryCarrierOption(models.Model):
...
-            (be it from cache or context) """
+    @api.depends('carrier_id.allowed_tmpl_options_ids')
+    def _compute_allowed_tmpl_options_ids(self):
+        """ Gets the available template options from related delivery.carrier
+            (be it from cache or context)."""
         for option in self:
-            domain = self.env.context.get('default_tmpl_option_id_domain')
-            option.tmpl_option_id_domain = (
-                option.carrier_id.allowed_options_domain or domain or '[]')
+            domain = self.env.context.get('default_allowed_tmpl_options_ids')
```
ghost: doesn't appear to me that `domain` is the correct name for the content of it now

delivery_carrier_label_postlogistics/models/delivery.py
```diff
@@ -109,18 +109,20 @@ class DeliveryCarrierOption(models.Model):
...
+            option.allowed_tmpl_options_ids = (
+                option.carrier_id.allowed_tmpl_options_ids or domain or '[]')
```
ghost: `or '[]'` seems quite dangerous to use here, since you're writing to `Many2many` :smile:

tschanzt (commented):
There is a lot of commented xml around. I think you should add a comment why it is still needed or it should be dropped.

asaunier:
Perhaps you could squash the old consecutive commits of the same authors?

grindtildeath:
@asaunier I will do it as soon as https://github.com/OCA/delivery-carrier/pull/128 is merged, but I prefer keeping it so if more is to be changed.

simahawk:
https://github.com/OCA/server-tools/pull/1261 merged

Build red

grindtildeath:
@simahawk @yvaucher @tschanzt @sebalix @asaunier 
Can you guys please pass byy for a review and get this old PR merged ?

oca_dependencies.txt
```diff
@@ -0,0 +1,7 @@
+# list the OCA project dependencies, one per line
+# add a github url if you need a forked version
+partner-contact
+server-tools
+stock-logistics-workflow
+webkit-tools
+web
```
sebalix, line 7: We only need `server-tools` here (the module depends on `configuration_helper` from OCA/server-tools and `base_delivery_carrier_label` depending on `delivery` only which is a standard one).

sebalix (changes requested):

tschanzt (approved):

asaunier (approved):

simahawk (approved):

oca_dependencies.txt
pedrobaeza: Please attend this comment for not making slower this CI

sebalix:
@grindtildeath can you cherry-pick the commit from https://github.com/OCA/delivery-carrier/pull/191 too?

PR #156, opened 2018-05-29
[11.0][FIX] Remove occurences of carrier type

grindtildeath:
Field carrier_type was originally type and was used to differentiate delivery carriers.
However odoo did implement delivery_type since v9 with this exact goal :
https://github.com/odoo/odoo/blob/11.0/addons/delivery/models/delivery_carrier.py#L22
This is also the field used in the implementation of enterprise modules for carriers.
Therefore we should use it instead of redefining a new one..

BTW Would be great to backport this to 10.0 and 9.0.

### DIFF
--- a/base_delivery_carrier_label/__manifest__.py
+++ b/base_delivery_carrier_label/__manifest__.py
@@ -1,7 +1,7 @@
 # Copyright 2013-2015 Yannick Vaucher (Camptocamp SA)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 {'name': 'Base module for carrier labels',
- 'version': '11.0.1.0.0',
+ 'version': '11.0.1.0.1',
  'author': "Camptocamp,Akretion,Odoo Community Association (OCA)",
  'maintainer': 'Camptocamp',
  'category': 'Delivery',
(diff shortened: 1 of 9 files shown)


yvaucher (approved):

simahawk (approved):
LGTM, just version upgrade missing.
Approving to not get back ;)

pedrobaeza:
Don't you need a migration script or at least an `oldname` attribute?

grindtildeath:
@pedrobaeza I did add the `oldname`, so should be fine as it is now?

pedrobaeza:
It depends on the values. They should be the same?

grindtildeath:
@pedrobaeza Yep, but the change also has to be done in implementation modules. 
E.g : https://github.com/OCA/delivery-carrier/blob/10.0/delivery_carrier_label_gls/delivery.py#L28
This should then be migrated as :
`delivery_type = fields.Selection(selection_add=('gls', 'Gls'))`
And should work easily.
We did it in v10 for postlogistics : https://github.com/yvaucher/carrier-delivery/blob/9811dbf2165e0d9af4fef86a25865079df8e33f5/delivery_carrier_label_postlogistics/models/delivery.py#L131

pedrobaeza:
OK, thanks

asaunier (approved):

PR #158, opened 2018-06-13
[10.0] add delivery_dropoff_site

legalsylvain:
This module is a port and a refactor of the @akretion module delivery_dropoff_site V7.0 available here : https://github.com/akretion/carrier-delivery-ak/tree/7.0/delivery_dropoff_site
Original authors @sebastienbeau, @bealdav.

It add a new model ```dropoff.site``` that are places where some of your carriers can deliver your sales.

In the dropoff site model, you can define an adress

dropoff_site_form

Optionnaly, you can define opening hours. (because most of the drop-off sites are shops.)

dropoff_site_form_calendar

Finally, drop-off site field is available on sale order (in the shipping_address_id field). 
If set, a new field 'Final Recipient' is available to set the name ofthe  partner that will
pick up the parcel in the dropoff site.


bealdav:
@legalsylvain 
Based on experience, not sure that's the better way to manage these cases. 
e.g. laposte
Alternative : only store carrier provider and dropoff site ID, then by querying webservice you get all the informations at the moment you need it

my 2 cents

Ask to @sebastienbeau

legalsylvain:
Hi @bealdav. Thanks for your review. 

In fact, I talked with @sebastienbeau when I refactored this module.

1. Calling webservice can be done to populate the database (by cron). So it's not incompatible.
2. In some cases, we don't have any webservices, so this module is usefull in that case. For exemple, the network of AMAP, in France. Customers can recover the order in some "Point de retrait".

regards.

delivery_dropoff_site/models/delivery_carrier.py
```diff
@@ -0,0 +1,16 @@
+# coding: utf-8
+# Copyright (C) 2018 - Today: GRAP (http://www.grap.coop)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import fields, models
```
bealdav: s/ openerp / odoo
in several files

delivery_dropoff_site/models/stock_picking.py
```diff
@@ -0,0 +1,19 @@
...
+    _inherit = 'stock.picking'
+
+    final_shipping_partner_id = fields.Many2one(
+         comodel_name='res.partner', string='Final Recipient',
+         help="It is the partner that will pick up the parcel "
+              "in the dropoff site.")
+
+    @api.model
+    def create(self, vals):
+        return super(StockPicking, self).create(vals)
```
bealdav: why override here ?

delivery_dropoff_site/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,35 @@
+This module extend Odoo functionnalities, regarding delivery features to
+add a new concept of drop-off Sites.
+
+Main international carriers provide transportation services to specific areas
+managed by them or by subcontractors.
+
+Then, recipients come pick up their packages in these sites.
+
+Usage
+=====
```
bealdav: Why not use USAGE.rst file ?

It could be improve presentation polymorphism : a la odoo apps, etc.

bealdav (changes requested):

bealdav (approved):

delivery_dropoff_site/__manifest__.py
```diff
@@ -0,0 +1,42 @@
...
+    'author': 'Akretion,GRAP,Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'summary': "Send goods to sites in which customers come pick up package",
+    'category': 'Delivery',
+    'depends': [
+        'delivery',
+        'base_geolocalize',
+        'resource',
+    ],
+    'website': 'https://www.odoo-community.org',
```
MeetKD: Can you please change website key to `https://github.com/OCA/delivery-carrier`

delivery_dropoff_site/models/res_partner.py
```diff
@@ -0,0 +1,51 @@
+# coding: utf-8
+# Copyright (C) 2014 - Today: Akretion (http://www.akretion.com)
+# Copyright (C) 2018 - Today: GRAP (http://www.grap.coop)
+# @author Aymeric Lecomte <aymeric.lecomte@akretion.com>
+# @author David BEAL <david.beal@akretion.com>
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
MeetKD, line 7: Author is already defined in contributor's readme file. Dont you think so it should be removed from all py files ?

MeetKD (changes requested):

delivery_dropoff_site/__manifest__.py
legalsylvain: indeed. (https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#modules)
done.

delivery_dropoff_site/models/res_partner.py
legalsylvain: Well, 
- I dont see any OCA rules that says that we should remove authors and contributors from the header section. 
- There are such header in most of the OCA modules.

I think it's not the place here to talk about new general OCA rules. If you think it's usefull, please make an issue on maintainer-tools.

In the meantime, let's go forward with that PR.

thanks for your review.

legalsylvain: @MeetKD, could you update your review ?
thanks !

MeetKD (approved):

PR #164, opened 2018-08-30
[11.0][NEW] partner_delivery_schedule: New module for set a delivery schedule for partners

sergio-teruel:
cc @Tecnativa


partner_delivery_schedule/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    'name': 'Partner Delivery Schedule',
+    'summary': 'Set on partners a schedule for delivery goods',
+    'version': '11.0.1.0.0',
+    'development_status': 'Production/Stable',
```
pedrobaeza: It should be `Beta` when the module first arrives. It gets to `Production` when it's present in 2 versions. See https://odoo-community.org/page/module-maturity-levels

partner_delivery_schedule/models/partner_delivery_schedule.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class PartnerDeliverySchedule(models.Model):
+    _name = 'partner.delivery.schedule'
+
+    name = fields.Char(compute='_compute_name', store=True)
+    partner_id = fields.Many2one(
```
pedrobaeza: This shouldn't have `partner_id`. In my POV, you define generic delivery interval: "from 10:00 to 14.00", and then you assign that delivery interval to the partners you want. It should also include optional days (multiple selection), as on Mondays can be from 10:00 to 14:00, but on Tuesdays and Fridays from 9:00 to 13.00, and no delivery on Saturdays nor Sundays. The `name_get` can be composed with initial letters of each day (ordered) + delivery interval

pedrobaeza (changes requested):
Put in the roadmap of the README to include a method in res.partner model for telling if a datetime is inside the delivery slots of that partner

sergio-teruel:
Changes done

sergio-teruel:
@pedrobaeza I have added partner helper. Thanks

partner_delivery_schedule/models/partner_delivery_schedule.py
```diff
@@ -0,0 +1,74 @@
...
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class DeliverySchedule(models.Model):
+    _name = 'delivery.schedule'
+
+    name = fields.Char(required=True)
+    hour_from = fields.Float(string='From')
+    hour_to = fields.Float(string='To', default=24.00)
```
pedrobaeza: required

partner_delivery_schedule/models/partner_delivery_schedule.py
```diff
@@ -0,0 +1,74 @@
...
+    monday = fields.Boolean()
```
pedrobaeza: Put by default the 5 working days as True

partner_delivery_schedule/models/partner_delivery_schedule.py
```diff
@@ -0,0 +1,74 @@
...
+    def _check_day_selected(self):
+        if not any([self[x[0]] for x in self._days_of_week()]):
+            raise ValidationError(
+                _('Error ! You must set one day to delivery.')
+            )
+        return True
+
+    def _days_of_week(self):
+        return [
+            ('monday', _('Monday')),
```
pedrobaeza, line 44: You should add a translatable term for the initial of each day, as it's not enough taking the first letter of this. Example: in Spanish, we have `Martes` and `Miércoles`, which will display in both `M`

partner_delivery_schedule/models/res_partner.py
```diff
@@ -0,0 +1,33 @@
...
+        date = fields.Datetime.from_string(date_str)
+        week_day = day_list[date.weekday()][0]
+        delivery_records = self.delivery_schedule_ids.filtered(week_day)
+        if not delivery_records:
+            return False
+        request_delivery_hour = int(date.hour) + float(date.minute / 60)
+        rec_to_delivery = delivery_records.filtered(lambda x: (
+            request_delivery_hour >= x.hour_from and
+            request_delivery_hour < x.hour_to))
+        if not rec_to_delivery:
```
pedrobaeza: Simplify it with `return bool(rec_to_delivery)`

partner_delivery_schedule/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_delivery_schedule_user,access_delivery_schedule_user,model_delivery_schedule,base.group_partner_manager,1,0,0,0
```
pedrobaeza: I think you should add a general rule read only for employees, as they access contacts although they don't have that permission. Salesman should be also allowed to change delivery times (more than stock users). See delivery module for copying the permission scheme with delivery carrier.

pedrobaeza (changes requested):

sergio-teruel:
@pedrobaeza changes done, simplify partner function and I have added tests for record access...

partner_delivery_schedule/models/partner_delivery_schedule.py
sergio-teruel: Ok, but I have taken into account the two first letterf for every day.. so Martes is 'Ma' and Miercoles is 'Mi'

pedrobaeza: I see... but it's better to use only one letter for not doing the m2m tag very long, although you have the option to be more if you want to translate it that way

partner_delivery_schedule/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_delivery_schedule_user,access_delivery_schedule_user,model_delivery_schedule,base.group_user,1,0,0,0
```
pedrobaeza, line 2: I still have doubts with a portal user that accesses the partner if we are going to have permission access. Can you please check?

sergio-teruel: So, a portal user has not access to partner view due to property_delivery_carrier_id field added by delivery module...
Anyway.... a portal user has not access to a partner view (I think), in old versions there was a menu to access to partner but now only access by website templates... so if you do not render the field you can access to a partner view

partner_delivery_schedule/models/partner_delivery_schedule.py
sergio-teruel: Ohh!!! sorry but Odoo do not generate a traslatable term  one word... (Of course), so if in english is M (monday) the translatable term not exists.., when a user check all days the name_get set to 'all days'..
I think that there is enought available space into m2m_tag field..

sergio-teruel: seleccion_020

partner_delivery_schedule/security/ir.model.access.csv
pedrobaeza: Cleared by direct conversation

partner_delivery_schedule/models/partner_delivery_schedule.py
pedrobaeza: OK, we let it this way

pedrobaeza (approved):

cristinamartinrod (approved):
Works perfefct

rafaelbn (approved):
I think this is RTM :+1:

pedrobaeza:
My comments aren't still honored.

rousseldenis:
> My comments aren't still honored.

@pedrobaeza They are

rousseldenis (approved):
Code review. LGTM

pedrobaeza:
@rousseldenis I'm seeing now. It's weird that the comments are unfolded although they are changed. I didn't see previously the label "Outdated". I prefer previous behavior...

Any way he has told me to not merge until he checks one last thing.

pedrobaeza:
OK, I got the confirmation, so merging. Thanks to all!

PR #165, opened 2018-09-10
[ADD] delivery_price_by_category

SimoRubi:
Add price rule based on the category of products in sale order


tafaRU:
@SimoRubi, I noticed there is no runbot instance where to test it.
Could you please give a try to trigger new builds as follows?
`git commit --amend`
`git push -f`

SimoRubi:
@tafaRU force pushed and still no runbot, maybe @sbidoul can help us understanding what is the issue?

sbidoul:
Unfortunately I've a limited understanding of how runbot works. Maybe @gurneyalex can help.

SimoRubi:
@tafaRU runbot is now up and running

eLBati (changes requested):
Created a delivery method with 1 rule: All / Saleable / Software; 3.00
Created a sale order for `Agrolait`; manually set the new delivery method.
Added a line with `Ice cream`
Clicked on `set price`
Got
```
  File "/home/odoo/odoo-10.0/addons/delivery/models/sale_order.py", line 27, in _compute_delivery_price
    order.carrier_id.with_context(order_id=order.id).price, order.pricelist_id.currency_id)
  File "/.repo_requirements/odoo/odoo/fields.py", line 915, in __get__
    self.determine_value(record)
  File "/.repo_requirements/odoo/odoo/fields.py", line 1027, in determine_value
    self.compute_value(recs)
  File "/.repo_requirements/odoo/odoo/fields.py", line 981, in compute_value
    self._compute_value(records)
  File "/.repo_requirements/odoo/odoo/fields.py", line 972, in _compute_value
    getattr(records, self.compute)()
  File "<decorator-gen-143>", line 2, in get_price
  File "/.repo_requirements/odoo/odoo/api.py", line 369, in loop
    result = [method(rec, *args, **kwargs) for rec in self]
  File "/home/odoo/odoo-10.0/addons/delivery/models/delivery_carrier.py", line 140, in get_price
    self.price = computed_price * (1.0 + (float(self.margin) / 100.0))
TypeError: unsupported operand type(s) for *: 'NoneType' and 'float'
```

tafaRU:
@SimoRubi, FYI, the same error written above also appears when you try the following steps:

Create a delivery method with 1 rule: All / Saleable / Software; 3.00
Create a sale order for Agrolait; manually set the new delivery method.
Add a line with Ice cream
Save the sale order.

delivery_price_by_category/models/delivery_price_rule.py
```diff
@@ -19,6 +19,13 @@ class PriceRule(models.Model):
     def _onchange_product_category_price(self):
         self.list_base_price = self.product_category_price
 
+    @api.onchange('product_category_id')
+    def _onchange_product_category_id(self):
+        if self.product_category_id:
+            # Reset the fields when they get hidden in the form
+            self.list_base_price = 0.0
+            self.write(self.default_get(['variable', 'operator', 'max_value']))
```
rousseldenis: @SimoRubi Are you serious ?

SimoRubi: @rousseldenis in the method evaluating the rules, I have to call `super` in order not to break inheritance.
`super` evaluates all the rules, then it would evaluate also the _category_ rules, but it would evaluate them based on the fields `'variable', 'operator', 'max_value'`.
This is the only solution that I found that doesn't `return` in a loop nor break inheritance, I'd be happy to find another one though

rousseldenis: Mmmh, I think the problem is elsewhere. But definitively, you can't call write() in an onchange.

SimoRubi: @rousseldenis why can't I do that? Isn't it the same as assigning a variable?
Here I see the `write` call just as a shortcut to
```
self.variable = self.default_get(['variable'])['variable']
self.operator = self.default_get(['operator'])['operator']
self.max_value = self.default_get(['max_value'])['max_value']
```

pedrobaeza: `write` dumps data to DB, which you don't want to do. You should do `update` instead for performing the update on the cache, and eventually being dumped to DB if you click on save button.

delivery_price_by_category/models/delivery_carrier.py
```diff
@@ -0,0 +1,53 @@
...
+    _inherit = 'delivery.carrier'
+
+    @api.multi
+    def get_price_available(self, order):
+        if 'order_id' not in self.env.context:
+            return super(DeliveryCarrier, self) \
+                .with_context(order_id=order.id).get_price_available(order)
+        return super(DeliveryCarrier, self).get_price_available(order)
+
+    def get_price_from_picking(self, total, weight, volume, quantity):
```
rousseldenis: @SimoRubi I think your assertion is not correct. I would have done your tests in get_price_available function as the only condition to compute in the other manner is line.product_category_id. Don't override get_price_from_picking. And then you avoid annoying context changes.

delivery_price_by_category/models/delivery_price_rule.py
SimoRubi: Thanks! I thought that the `in_onchange` mode would have prevented writing to DB, I didn't know the existence of `update`

SimoRubi: Fixed up in https://github.com/OCA/delivery-carrier/pull/165/commits/96ec61fd587c50e28aa9996e394d12fea140208c

delivery_price_by_category/models/delivery_carrier.py
SimoRubi: Thanks I'll work on your approach, that might actually simplify the whole process

SimoRubi: @rousseldenis I tried your approach in https://github.com/OCA/delivery-carrier/pull/165/commits/5005b6dcd6dcd48b0b8a0a3318bb23a36760b580, is that what you meant?

delivery_price_by_category/models/delivery_carrier.py
```diff
@@ -0,0 +1,60 @@
...
+
+from odoo import api, models
+from odoo.tools import safe_eval
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = 'delivery.carrier'
+
+    @api.multi
+    def get_price_available(self, order):
```
rousseldenis, line 13:
```suggestion
    @api.multi
    def get_price_available(self, order):
        self.ensure_one()
        category_price = 0.0
        price_dict = self.get_price_dict(order)
        for line in self.price_rule_ids:
            if line.product_category_id:
                products = order.mapped('order_line.product_id')
                if any(product.categ_id == line.product_category_id for product in products):
                    category_price = line.product_category_price
                    break
        if category_price:
            return category_price
         # Note that this will evaluate all the price_rule_ids again and
        # our category rules might interfere withthe correct computation
        return super(DeliveryCarrier, self).get_price_available(order)
```

Your tests on the else part can be removed. Keep it simple :smile: 

And I think the loop can be avoided if (but I don't know the process) if you find directly the category price

SimoRubi: @rousseldenis thanks for the code suggestion! 
The test in the `else` part is necessary because if the rules are:

1. standard rule that will pass
2. category rule that will pass

your method would apply the price of rule 2 but the expected behavior is to have the price of rule 1

rousseldenis: Sorry, mixed break and continue... end of day...

SimoRubi: @rousseldenis ahah yes after 8 hours is better to `break` than to `continue`..

Anyway, what I said in https://github.com/OCA/delivery-carrier/pull/165#discussion_r231609393 is not resolved with a `continue`: I think we should always check both types of rule in order to get the 1st one which satisfies the condition (`safe_eval` for standard rules and `category` for _category_ rules).

Too bad the core method is not really written to be easily overridden.

SimoRubi:
@rousseldenis is this final version ok for you?

SimoRubi:
@eLBati I checked locally and now the steps you provided do not generate any error, could you please update your review?
@tafaRU same as above, could you please review?

Thanks

SimoRubi:
@rousseldenis if this final version is ok for you I'll fix everything up to make commit history more clear.

SimoRubi:
@rousseldenis ?

rousseldenis (approved):
Code review

rousseldenis:
@eLBati

SimoRubi:
@eLBati can you update your review? Thanks

rousseldenis:
@eLBati

eLBati (approved):
Sorry for the delay

PR #167, opened 2018-10-18
[11.0][NEW] partner_delivery_zone: New module to group partners, orders and pickings by delivery zones

sergio-teruel:
cc @Tecnativa


partner_delivery_zone/models/res_partner.py
```diff
@@ -0,0 +1,20 @@
...
+
+    delivery_zone_id = fields.Many2one(
+        comodel_name='partner.delivery.zone',
+        string='Delivery Zone',
+        ondelete='restrict',
+        index=True,
+    )
+
+    @api.model
+    def _commercial_fields(self):
```
rousseldenis: @sergio-teruel Really? This could be too restrictive as it forbids to have a company with several delivery zones.

sergio-teruel: Uuuupss!!! You are right

pedrobaeza: You can still provide a default when creating the contacts.

sergio-teruel:
@rousseldenis Changes done, also update the test

rousseldenis (approved):
Code review.

@sergio-teruel Could the 'zone' be generalized ? It could be nice to have a general defintion of it. And then reuse it on some interesting functional points (like sale zones, or ... and add after that a geoengine layer).

rousseldenis:
@sergio-teruel pylint and flake8 fail

pedrobaeza:
@rousseldenis unfortunately we don't have budget for more changes, as this being very small (the model has 2 fields), it can be extracted or replicated easily if someone has that need to put this on other place without depending on this module.

@sergio-teruel, little flake8 fail:
```
partner_delivery_zone/models/res_partner.py:3:1: F401 'odoo.api' imported but unused
```
I'm reviewing code now.

partner_delivery_zone/models/sale_order.py
```diff
@@ -0,0 +1,18 @@
...
+    delivery_zone_id = fields.Many2one(
+        comodel_name='partner.delivery.zone',
+        string="Delivery Zone",
+        ondelete='restrict',
+        index=True,
+    )
+
+    @api.onchange('partner_shipping_id')
+    def onchange_partner_shipping_id_delivery_zone(self):
+        self.delivery_zone_id = self.partner_shipping_id.delivery_zone_id.id
```
pedrobaeza: Only change this if the partner has a delivery zone set. If not, preserve the previous value.

partner_delivery_zone/models/stock_picking.py
```diff
@@ -0,0 +1,17 @@
...
+
+    delivery_zone_id = fields.Many2one(
+        comodel_name="partner.delivery.zone",
+        string="Delivery Zone",
+        index=True,
+    )
+
+    @api.onchange('partner_id')
+    def onchange_partner_id_zone(self):
+        self.delivery_zone_id = self.partner_id.delivery_zone_id
```
pedrobaeza: The same about preserving previous value

partner_delivery_zone/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,4 @@
+This module allows set delivery zones on partner, this information is written
```
pedrobaeza: s/allow set/allows to set.

Put a `.` instead of comma before `this information...`

partner_delivery_zone/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,4 @@
...
+in sale orders and stock pickings.
+Also adds searchs and groups in partners, sales orders asn stock pickings
```
pedrobaeza: s/Also/It also
s/searchs/searches
s/asn/and

partner_delivery_zone/readme/ROADMAP.rst
```diff
@@ -0,0 +1,2 @@
+* For a future, the system will can set delivery zones based on partner zip.
```
pedrobaeza: s/a/the
s/the system/system
s/will can/can

partner_delivery_zone/readme/ROADMAP.rst
```diff
@@ -0,0 +1,2 @@
...
+  I will create a rules for that.
```
pedrobaeza: Remove this sentence

partner_delivery_zone/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
+To use this module you need to:
+
+#. Go to *Contacts > Create* or *Sales > Customers > Create*.
```
pedrobaeza: Put second part of this sentence as a new point before "Create a sales order..."

partner_delivery_zone/views/res_partner_view.xml
```diff
@@ -0,0 +1,31 @@
...
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+    <record id="view_partner_form" model="ir.ui.view">
+        <field name="model">res.partner</field>
+        <field name="name">partner delivery zone</field>
+        <field name="inherit_id" ref="base.view_partner_form"/>
+        <field name="arch" type="xml">
+            <field name="industry_id" position="after">
+                <field name="delivery_zone_id"/>
```
pedrobaeza, line 12: Delivery zone is not shown nor put as default for children contacts managed from the same view.

pedrobaeza (changes requested):

partner_delivery_zone/tests/test_partner_delivery_zone.py
```diff
@@ -0,0 +1,69 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo.tests import SavepointCase
+
+
+class TestPartnerDeliveryZone(SavepointCase):
+    at_install = False
+    post_install = True
```
rousseldenis, line 9: @sergio-teruel @pedrobaeza

pedrobaeza: Well, we put this in our modules by default, and prefer to leave it, as we are maintaining them.

rousseldenis: @pedrobaeza To make reviewers work easy, I will port the debate at higher level. I want to avoid possible future debates (time eating) if some modules collide on tests execution (you are not maintainer on all modules).

I will stop merging such ones before decision.

Moving to https://github.com/OCA/odoo-community.org/

pedrobaeza: @rousseldenis OK, we are going to remove this part, but as you say, take into account that we are not going to face this debate again. As soon as we have an error in our integration tests because of this, I'm going to put it directly. @sergio-teruel please remove it meanwhile.

rousseldenis: > I'm going to put it directly

Depending on future guidelines uh ? ;-)

pedrobaeza: You can't enforce that rule, and if so, we will be out of OCA automatically. If you don't perform integration tests nor continuous deployment, it doesn't mean that others shouldn't do them.

rousseldenis: > If you don't perform integration tests nor continuous deployment

Of course we do (that's the main part of process)! But you can't do it for a whole OCA repository ! Moreover, integration test cases could be unique by project!

I think you don't understand difference between integration tests and unit tests.

> Well, we put this in our modules by default

I think that's quite new in your dev processes as it should have been seen before. Along all versions of Odoo, the tests process hasn't changed a lot, but always been at install. So, we were wrong for so many years ?

pedrobaeza: > Of course we do (that's the main part of process)! But you can't do it for a whole OCA repository ! Moreover, integration test cases could be unique by project!
> I think you don't understand difference between integration tests and unit tests.

That's the key: we detect problems when we perform integration tests on our projects, and this is when we add this post_install, for getting them green. As it's more work to detect it afterwards, now we put them by default. And it's curious that we haven't found the other problem yet... (well, in fact it's not curious, but a very specific case that rarely or never happens).

> I think that's quite new in your dev processes as it should have been seen before. Along all versions of Odoo, the tests process hasn't changed a lot, but always been at install. So, we were wrong for so many years ?

Not so new. We are performing this for about 6-7 months. There are already 30-35 modules with these keys across OCA.

rousseldenis: > Not so new. We are performing this for about 6-7 months

Quite comparing to years running travis

pedrobaeza: Yeah, because Travis only performs unit tests + repository integration tests, not full integration tests.

rousseldenis: > Yeah, because Travis only performs unit tests + repository integration tests, not full integration tests.

So, you agree you introduce a new way of testing

pedrobaeza: We are not debating about that. We are debating about forcing to not be possible to do this with OCA modules, what is what you are trying to do forbidding this by guidelines.

rousseldenis: IMHO you have to override that way of testing in your integration process. You cannot constrain developers to think about 'what could happen with ...possible... installed modules'.

pedrobaeza: The same can apply to your reasoning. And this is not something that can be adapted easily. But we are not forcing you that. If we find any case that this is problematic, we solve it. You don't have to think about anything. And I repeat, till now, 0 problems in all your integration tests of our deployments in v9, v10 and v11, so the theoretical case where post_install is a problem hasn't arisen yet (and I doubt it does, at least because of the post_install part). Any way, we have already done a PR over Odoo for improving this situation and to not need post_install anymore, but this will be for v13 hopefully.

sergio-teruel:
@rousseldenis @pedrobaeza Changes done!.
Added default delivery zone for child_ids field in context and fix typo and form view for contacts

pedrobaeza (approved):
Thanks for the changes. Next time please put a new fixup commit instead of amending previous one, as this way it's very difficult to review latest changes.

sergio-teruel:
Oks. Thanks

PR #100, opened 2016-08-04
[9.0] port delivery_carrier_label_default

yvaucher:

PR #169, opened 2018-11-15
[11.0][MIG] delivery_carrier_label_default

sebalix:
Migrated from the 9.0 PR: https://github.com/OCA/delivery-carrier/pull/100


delivery_carrier_label_default/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2013-2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+{
+    "name": "Default label for carrier labels",
+    "summary": ("This module defines a basic label to print "
```
simahawk: FYI: you are already inside parenthesis, you can skip these ;)

delivery_carrier_label_default/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+                "when no specific carrier is selected."),
+    "version": "11.0.1.0.0",
+    "development_status": "Beta",
```
simahawk: is it really beta?

delivery_carrier_label_default/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "category": "Delivery",
+    "website": "https://github.com/OCA/delivery-carrier",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
```
simahawk: you can skip these flags

delivery_carrier_label_default/tests/test_print_label.py
```diff
@@ -1,28 +1,50 @@
...
 from openerp.tests import common
 
+from .common import HTMLRenderMixin
+
+MOD_PATH = 'odoo.addons.base_delivery_carrier_label.models'
+LABEL_MODEL = MOD_PATH + '.shipping_label.ShippingLabel'
+
+
+class TestPrintLabel(common.TransactionCase, HTMLRenderMixin):
+    """When running tests Odoo 11 renders PDF reports as HTML, so we are
```
simahawk:
```py
"""Test label printing.

When running tests Odoo 11 renders PDF reports as HTML,
so we are going to test the shipping labels as HTML document only.
A good side effect: we are able to test the rendered content.
"""
```

delivery_carrier_label_default/tests/test_print_label.py
```diff
@@ -1,28 +1,50 @@
...
-class TestPrintLabel(common.TransactionCase):
+    def check_label_content(self, b64_datas):
+        html_datas = base64.b64decode(b64_datas)
+        node = self.to_xml_node(html_datas)[0]
+        for div_class in ['page', 'address', 'recipient']:
+            tags = self.find_div_class(node, div_class)
+            self.assertEqual(len(tags), 1)
 
-    def test_print_default_label(self):
+    @mock.patch(LABEL_MODEL + '._selection_file_type')
```
simahawk: as the mocking is required for all tests you can move this to `setUp` + add a comment saying why you patch it.

delivery_carrier_label_default/models/stock.py
```diff
@@ -1,27 +1,19 @@
...
-        report_obj = self.env['report']
-        pdf_report = report_obj.get_pdf(
-            self, 'delivery_carrier_label_default.report_default_label',
-            data=data)
-
-        return {'name': '%s.pdf' % "Shipping Label",  # report.report_name,
-                'file': pdf_report,
-                'file_type': 'pdf',
-                }
+        report = self.env.ref('delivery_carrier_label_default.default_label')
```
simahawk, line 13: note for improvement: maybe we can make this configurable at company level.
Not mandatory: if you don't want to do it now (it's a migration) you can drop a comment here for the future.

delivery_carrier_label_default/views/report_default_label.xml
```diff
@@ -1,12 +1,14 @@
 <odoo>
 <template id="report_default_label">
-  <t t-call="report.html_container">
+  <t t-call="web.html_container">
```
simahawk: you can use `web.basic_layout` and remove `<div class="article o_report_layout_standard">`

simahawk (changes requested):

simahawk:
also, please reword the last commit as `delivery_carrier_label_default: use new README structure`

delivery_carrier_label_default/__manifest__.py
sebalix: I have not enough perspective on that point, there was no development status before, so I preferred to be conservative about this status at first, but if you think it's `Production/Stable`, so be it. What do you think?

delivery_carrier_label_default/tests/test_print_label.py
sebalix: it doesn't work, I suppose that once we left the `setUp` method, the mocking disappears and is not available on `test_` methods. Looking for another solution.

delivery_carrier_label_default/views/report_default_label.xml
sebalix: fixed, thanks for the tip

delivery_carrier_label_default/models/stock.py
sebalix: note added

delivery_carrier_label_default/tests/test_print_label.py
sebalix: I implemented a decorator to patch the field, a half way solution.

sebalix:
Fixed + rebased.

delivery_carrier_label_default/__manifest__.py
simahawk: well, if I'm not wrong this module is around since ages and been used in prod since a lot of time. 
Hence, I think it should be stable.

delivery_carrier_label_default/tests/test_print_label.py
simahawk: well, I meant to use `self.patch` + start (you can find some example in existing modules, event for instance).
Anyway, not important anymore... :)

delivery_carrier_label_default/tests/test_print_label.py
```diff
@@ -0,0 +1,130 @@
...
+        labels = self.env['shipping.label'].search(
+            [('res_id', '=', self.picking.id)])
+        self.assertEquals(len(labels), 2)
+        for label in labels:
+            self.assertTrue(label.datas)
+            self.assertEquals(label.name, "Shipping Label.html")
+            self.assertEquals(label.file_type, 'html')
+            self.check_label_content(label.datas)
+
+    def setUp(self):
```
simahawk: pls, do not put tests set up at the bottom: they are the very beginning of a test suite...

Plus, pls, move the base class to SavepointCase and this code here to setUpClass otherwise you create all these records again again and again for each test.

Finally, use this trick at the beginning of setUpClass to turn off tracking:
```py
cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```

sebalix: I agree, it was already at the bottom so I tend to reduce the diff size if possible ;) Np to move it.

delivery_carrier_label_default/README.rst
```diff
@@ -0,0 +1,82 @@
+================================
```
Tonow-c2c, line 1: May be change Readme to modular like [https://github.com/OCA/maintainer-tools/tree/master/template/module/readme](https://github.com/OCA/maintainer-tools/tree/master/template/module/readme)

sebalix: It was already done, check this commit: 603b0bf88c79a6f31f0138952054ece723a58b9b

delivery_carrier_label_default/tests/test_print_label.py
sebalix: Done.

simahawk (approved):
LG, tnx for your work!

delivery_carrier_label_default/README.rst
Tonow-c2c: thanks sorry

Tonow-c2c (approved):
LGTM

delivery_carrier_label_default/models/stock.py
pedrobaeza: Please put the note in ROADMAP.rst file and regenerate README.

pedrobaeza (changes requested):

sebalix:
@pedrobaeza done, ready to merge

pedrobaeza (approved):

PR #173, opened 2019-01-11
[11.0][MIG] delivery_price_rule_untaxed

tarteo:


SimoRubi:
@tarteo please fixup all the migration commits into only one, that makes the review process way easier, thanks!

tarteo:
:+1: rebased it.

delivery_price_rule_untaxed/models/delivery_carrier.py
```diff
@@ -11,14 +10,20 @@ class DeliveryCarrier(models.Model):
...
 
     @api.multi
-    def get_price_available(self, order):
+    def _get_price_available(self, order):
         self.ensure_one()
+
+        delivery_line = order.order_line.filtered(
+            lambda line: line.is_delivery
+        )
+        delivery_price = delivery_line and delivery_line.price_subtotal or 0.0
```
SimoRubi: What if there is more than one?

delivery_price_rule_untaxed/tests/test_delivery_price_rule_untaxed.py
```diff
@@ -31,19 +29,22 @@ def test_untaxed(self):
                 'product_id': self.product_4.id,
                 'product_uom_qty': 1,
                 'product_uom': self.product_uom_unit.id,
-                'price_unit': 750.00,
+                'price_unit': 1250.00,
             })],
             'carrier_id': self.untaxed_delivery.id
         })
 
         # Raise exception because total untaxed (750) is above threshold (300)
```
SimoRubi: Please update comments according to code changes

SimoRubi (changes requested):
Thanks!
I see that you added something about delivery price, could you please add a little explanation in the README?
Moreover, please separate in another commit the changes for the new functionality so it can be easily migrated back to `10.0` when merged.

delivery_price_rule_untaxed/models/delivery_carrier.py
tarteo: This is never the case as only one is created by module `delivery`

SimoRubi: Then please add a `delivery_line[0]` or another kind of check/filter somewhere, otherwise this line would raise an exception if `delivery_line` has more than one record

tarteo: Sure

tarteo:
@SimoRubi Thanks for the review I applied your suggestions. I didn't add new functionality to this module. .

SimoRubi (approved):
Thanks! ok for me

delivery_price_rule_untaxed/models/delivery_carrier.py
```diff
@@ -11,14 +10,20 @@ class DeliveryCarrier(models.Model):
...
 
     @api.multi
-    def get_price_available(self, order):
+    def _get_price_available(self, order):
         self.ensure_one()
+
+        delivery_line = order.order_line.filtered(
+            lambda line: line.is_delivery
+        )
+        delivery_price = delivery_line and delivery_line[0].price_subtotal or 0.0
```
rousseldenis: @tarteo Prefer the use of first()

```python
from odoo.fields import first
delivery_price = first(delivery_line).price_subtotal if delivery_line else 0.0
```

rousseldenis (changes requested):
Minor comment

delivery_price_rule_untaxed/models/delivery_carrier.py
tarteo: Thanks I wasn't aware of this method. :+1:

rousseldenis (approved):
Code review

rousseldenis:
@tarteo Is this ok ?

tarteo:
@rousseldenis Yes :+1:

PR #178, opened 2019-01-31
[10.0][MIG] delivery_carrier_label_postlogistics_shop_logo: Migration to 10.0

grindtildeath:


pedrobaeza:
Why not integrating this on main module?

grindtildeath:
@pedrobaeza There's already an option in the main module to add the company logo to be printed on the labels.
Considering this module inherited `sale.shop` I suppose there should be another dependency to have a different logo between different shops, but I'm only figuring it out now.

pedrobaeza:
OK, just curiosity. Thanks for answering.

rousseldenis:
@grindtildeath Please check travis

grindtildeath:
As `sale.shop` model has been removed from odoo there's no need to migrate this module.

PR #179, opened 2019-03-12
[11.0][FIX] partner_delivery_zone: Get delivery zone from partner when a picking is created and no sale line is linked (pick-pack...)

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/partner_delivery_zone/tests/test_partner_delivery_zone.py
+++ b/partner_delivery_zone/tests/test_partner_delivery_zone.py
@@ -11,6 +11,7 @@ class TestPartnerDeliveryZone(SavepointCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        cls.warehouse = cls.env.ref('stock.warehouse0')
         cls.delivery_zone_a = cls.env['partner.delivery.zone'].create({
             'name': 'Delivery Zone A',
             'code': '10',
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

rousseldenis:
@sergio-teruel Maybe a little test to enlight this ?

rafaelbn (approved):
:+1: 

Please @sergio-teruel could you and a test?

sergio-teruel:
@pedrobaeza @rousseldenis @rafaelbn I change the way to obtain de deliverry zone. In the other PR I obtain the delivery zone from move parter but If the user changes the delivery zone in sale order the two other moves (pick and pack) will have the partne delivery zone instead of sale order delivery zone.
Can you review this please!!

partner_delivery_zone/tests/test_partner_delivery_zone.py
```diff
@@ -97,3 +98,9 @@ def test_default_line_discount_value(self):
         res = self.partner.fields_view_get(view_id=view.id, view_type='form')
         ctx = self._get_ctx_from_view(res)
         self.assertTrue('default_delivery_zone_id' in ctx)
+
+    def test_wharehouse_three_steps(self):
+        self.warehouse.delivery_steps = 'pick_pack_ship'
+        self.order.action_confirm()
+        zone = self.order.picking_ids.mapped('delivery_zone_id')
+        self.assertEqual(len(zone), 1)
```
pedrobaeza: This is not testing what you want: it will be false positive also if only the last picking has the zone. You have to do:  `for picking in self.order.picking_ids: self.assertEqual(picking.delivery_zone_id, ...)`

sergio-teruel:
Yes!! sorry. Changes done and test for a sale order without delivery zone

pedrobaeza (approved):

pedrobaeza:
@rousseldenis do you make the honors?

rousseldenis (approved):

PR #191, opened 2019-05-09
[10.0][FIX] delivery_carrier_label_postlogistics: fix integration WSDL file

sebalix:
The XSD schema can't be found at the specified URL, but the schema is
stored locally next to the WSDL file (like in the 'production' folder).


yvaucher (approved):

yvaucher:
Fast tracking as same as in 9.0

PR #195, opened 2019-05-27
[IMP]base_delivery_carrier_label: improve generation button

jesusVMayor:
Show only the label generation button when the picking don't have tracking ref.

### DIFF
--- a/base_delivery_carrier_label/__manifest__.py
+++ b/base_delivery_carrier_label/__manifest__.py
@@ -2,7 +2,7 @@
 # Copyright 2013-2015 Yannick Vaucher (Camptocamp SA)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {'name': 'Base module for carrier labels',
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

rousseldenis:
Isn't it a little bit restrictive ?

Maybe some people want to print the label without that ? It could be a carrier option ?

jesusVMayor:
I made it more flexible thanks to a idea from @zahravelasco
Now there is a boolean field for recreate the delivery.

rafaelbn:
Could we merge this @jesusVMayor ? this is v10 and we should upstream in v11 and v12. @rousseldenis what do you think?

rafaelbn (approved):

rousseldenis (approved):
Tested

rousseldenis:
@jesusVMayor Could you just update module version ?

jesusVMayor:
Version number changed.

Issue #192, opened 2019-05-13
RFC delivery carrier in v12 and above


bealdav:
According different features implemented on carriers modules, I think we need to refactor several points like :

1. remove carrier.account model https://github.com/OCA/delivery-carrier/blob/11.0/base_delivery_carrier_label/models/carrier_account.py to be replaced by server env or any equivalent features
2. in the purpose to reduce useless code, always use package to define parcel (no more picking)
3. open carriers features to the whole python ecosystem (or more) : other python ERP, other applications dealing with carriers, etc. An example has been made here http://akretion.github.io/roulier https://pypi.org/project/roulier
4. other stuff

For now there is https://github.com/OCA/delivery-carrier/pull/187 to be merged. I don't want to block the merge by this RFC, just plan evolution

cc @Timo17100-c2c @yvaucher @guewen

bealdav:
About 1. I propose to implement the same as that
https://github.com/OCA/server-env/blob/12.0/mail_environment/models/ir_mail_server.py
but for carrier.account . What do you think @yvaucher ? Can I make a PR on #187 ?

pedrobaeza:
Please don't depend on all the *_environment stuff.

bealdav:
In the whole ERP we need a good solution to manage credentials of external accounts. At Akretion we decided to use keychain (a version modified in v12) which itself based on server environment tools. C2C (and others) use server env only.

Please @pedrobaeza could you precise what is your generic option to deal with external credential accounts in the whole erp ? Thanks

pedrobaeza:
You were talking about *_environment, not keychain. That's why I told you about not using that. `keychain` is good for keeping some secrets.

florian-dacosta:
I guess the *_environment part can be put in a submodule then.

bealdav:
That's OK for me to add another module.
https://github.com/OCA/server-env/pull/22

bealdav:
2. @yvaucher @guewen @damdam-s are you open to modify base_delivery_label to avoid to deliver without package ?

yvaucher:
@bealdav about your question on delivery without package. And following the call we had.

We might have some little things to tweak on the OCA/rma modules when returning goods. Nevertheless, it would make indeed more sense if we rely on pack only and not consider on itself picking as a pack.
And we could enforce the creation of a pack if we print a label.

So let's go for labels generation only with packs.

[...]

PR #196, opened 2019-05-27
[WIP][12.0][ADD] Correos Express Label

angelmoya:
WIP

In spanish because this is a spanish carrier.

Falta por revisar el número de bultos, y contra reembolso.


bealdav:
@angelmoya Thanks for this contribution.

Have you seen https://github.com/OCA/delivery-carrier/issues/192#issuecomment-493087373
Are you ok for that bring these updates in base_delivery_carrier_label ?

Do you think your module can deal with these updates ?

Thanks

pedrobaeza:
If this is only for Spain, then maybe it's better to propose it to OCA/l10n-spain

rousseldenis:
> If this is only for Spain, then maybe it's better to propose it to OCA/l10n-spain

España y Portugal :-)

bealdav:
@angelmoya other topic on carriers side

Are you aware of the initiative on carriers :

https://github.com/akretion/roulier
https://github.com/akretion/roulier-gls-fr/tree/develop WIP

Carriers data process are hard to maintain, and it could help to share our works with whole python ecosystem and even more ecosystems (other languages) with python entry-points and json data (dict like)

What do you think ?

angelmoya:
Ok, I will check

rousseldenis:
> Ok, I will check

Maybe it could be nice to integrate the roulier lib to base_delivery_carrier_label ?

@bealdav @angelmoya Don't hesitate to create an RFC. We can talk about that as it is OCA sprint in spain now. I'm available to review an so on

bealdav:
RFC created https://github.com/OCA/delivery-carrier/issues/197

angelmoya:
Finally I think that will be easier to do on diferent modules, anyway before we need to check #199

rousseldenis:
> Finally I think that will be easier to do on diferent modules, anyway before we need to check #199

Yes, move forward, let the discussion about that big change become more mature

pedrobaeza:
Superseded by #434

PR #198, opened 2019-05-28
[9.0][IMP] delivery_carrier_label_postlogistics: use WebService v2.4

sebalix:


delivery_carrier_label_postlogistics/models/company.py
```diff
@@ -11,6 +17,11 @@ class ResCompany(models.Model):
     postlogistics_wsdl_url = fields.Char(compute='_get_wsdl_url',
                                          string='WSDL URL',
                                          store=True)
+    postlogistics_version = fields.Selection(
+        selection=POSTLOGISTICS_VERSIONS,
+        string="Version",
+        default=POSTLOGISTICS_VERSIONS[0][0],
```
i-vyshnevska, line 23: seems this should be a required field, maybe migration with initializing would be appropriate here :thinking:

sebalix: @i-vyshnevska the default value will be set when the field is created, and the field is not available on the company form. Maybe the required field could be added on the configuration panel to not allow the user to set the False value.
I say that because adding required fields on such common data model causes some trouble in unit tests which create companies, then we are forced to add this module as dependency on the others modules creating company records even if there is no functional links between them.
What do you think?

delivery_carrier_label_postlogistics/postlogistics/web_service.py
```diff
@@ -54,30 +52,12 @@ def __init__(self, company):
         self.init_connection(company)
 
     def init_connection(self, company):
-        if company.postlogistics_test_mode:
```
TDu, line 57: So this is not needed anymore with the new version of the api, I guess ?

TDu (approved):
LGTM

delivery_carrier_label_postlogistics/postlogistics/web_service.py
sebalix: All the namespace/location tricks is included in the WSDL file, so I do not see the point to keep this here, as well as the verification of the TLS certificate which is working now.

yvaucher: Good to know it has finally been fixed at the webservice barcode end.

yvaucher (approved):
LGTM

rousseldenis:
@sebalix Is this ready ?

sebalix:
@rousseldenis it is running well on production for some months now. We are still having issues with the test mode on some features, but it was already the case with 2.2.

PR #199, opened 2019-05-29
[FIX] carrier type is not the same that delivery type

angelmoya:
In OCA Sprint Code we was checking the migration to versión 12.0, and we think that it was a mistake in migration to version 11.0.

In version 10.0 we  have carrier_type field, it define the carrier company, and how to comunicate to send data and get labels.

In versión 11.0 we have delivery_type field, it define how to compute prices, it not the same scope.

In #155 we check that postlogistic is added to delivery_type, but I think that could be better if you can define postlogistic in carrier_type and delivery_type regardless.

It is usual to have delivery method with diferent sale price than purchase price, so if you habe two fields you can send data to your carrier but mantain fixed prices.

### DIFF
--- a/base_delivery_carrier_label/models/delivery_carrier.py
+++ b/base_delivery_carrier_label/models/delivery_carrier.py
@@ -8,7 +8,9 @@ from odoo import api, fields, models
 class DeliveryCarrier(models.Model):
     _inherit = 'delivery.carrier'
 
-    delivery_type = fields.Selection(oldname='carrier_type')
+    carrier_type = fields.Selection(
+        string='Type',
+    )
     code = fields.Char(
         help="Delivery Method Code (according to carrier)",
     )


angelmoya:
@jctrey

pedrobaeza:
I agree with this approach

jctrey:
Much better, why not create a base module for all delivery carrier functionalities as label and tracking link, instead of define the commented field in "base_delivery_carrier_label" module.

grindtildeath (changes requested):
"In versión 11.0 we have delivery_type field, it define how to compute prices, it not the same scope."

Sorry, but this is wrong: https://github.com/odoo/odoo/blob/11.0/addons/delivery/models/delivery_carrier.py#L22

Price computation should be done using `rate_shipment` as explained in the official module.

We changed this to 11.0 since we didn't want to interfere with what was already existing in 10.0, but in no way should we revert to this.

angelmoya:
This way you can not create delivery method free of charge with delivery company to send parcel and get labels.

jctrey:
@grindtildeath you're right "Price computation should be done using rate_shipment as explained in the official module." but the suggested field will be very useful for other purposes, as detect the way an specific carrier (as FedEx, UPS, ...) will compose tracking urls and so on regardless of the price calculation method wich will be shared by many carriers.

grindtildeath:
@angelmoya @jctrey 
But why would you need to hijack an existing field with a well defined purpose instead of defining another field with a clear purpose as well ? I don't get it...

jctrey:
@grindtildeath yes, I think that's the perfect solution, to create a new field, problem solved

rafaelbn:
@angelmoya thank! can you make bots :green_apple: ? :smile:

angelmoya:
We can cancel that, we will use other field.

PR #200, opened 2019-05-31
[12.0][WIP] ADD delivery carrier label for ASM

ljimenezsidoo:


chienandalu:
There's already a full fledged module for GLS ASM here (almost done): https://github.com/OCA/l10n-spain/pull/1392

pedrobaeza:
Yeah, closing as the other is more complete, and exclusive of Spain, so the host is also more suitable.

PR #201, opened 2019-06-05
[11.0][FIX] delivery_carrier_label_default: Print one label per page

sergio-teruel:
Hi, when you select more than one picking to print the shipping label, there is no break por page.

Before this PR:
Selección_252

After this PR:
Selección_253
 
cc @Tecnativa

### DIFF
--- a/delivery_carrier_label_default/__manifest__.py
+++ b/delivery_carrier_label_default/__manifest__.py
@@ -4,7 +4,7 @@
     "name": "Default label for carrier labels",
     "summary": "This module defines a basic label to print "
                "when no specific carrier is selected.",
-    "version": "11.0.1.0.0",
+    "version": "11.0.1.0.1",
     "development_status": "Production/Stable",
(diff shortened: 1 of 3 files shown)


Tardo:
I have an error when try print labels from runbot...
odoo_error

sergio-teruel:
the error is not caused by the new changes

chienandalu (commented):
Seems that .travis is outdated and is causing the issue @Tardo commented https://github.com/OCA/delivery-carrier/blob/11.0/.travis.yml#L35

pedrobaeza:
OK, we don't know right now the reason for blocking that version, but let's code review this without trying in runbot, and 12.0 should be clearer.

pedrobaeza (approved):

chienandalu (approved):
All right then :+1:

pedrobaeza:
Let's fast track this.

PR #202, opened 2019-06-12
[12.0][MIG] port delivery_price_rule_untaxed to v12

rdeheele:


rdeheele:
the ci/runbot fails to build but I don't see any logs in ci/runbot showing errors
@pedrobaeza any advices? thanks

pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0

For the runbot question, it was a problem in the server tonight. When you fix the commit history, it will probably go well

rdeheele:
@pedrobaeza thanks, I re-created it in #203

PR #186, opened 2019-04-24
[MIG] port delivery_carrier_label_default to V10

mFlayyan:
migrating deliver_carrier_label_default from 9.0 to 10.0

PR #207, opened 2019-07-17
[MIG] port delivery_carrier_label_default to V10

thomaspaulb:
Replaces https://github.com/OCA/delivery-carrier/pull/186


thomaspaulb:
@pedrobaeza Could you help me here also? I don't see any warnings in runbot here, it looks like some kind of timeout.

pedrobaeza:
It seems all the builds are stuck. This is probably a failed tests + the use of any printing, as the way wkhtmltopdf opens connections hangs threads in these conditions.

thomaspaulb:
@pedrobaeza I managed to reproduce this condition on my local. The test generates a report, during which wkhtmltopdf tries to download CSS files from http://localhost:8069, and this request does not complete. Even running `wget http://localhost:8069/web` in parallel does not complete.

I failed to see this before because in my local, I usually run tests with `--xmlrpc-port=18079`, in which case the request _does_ complete and all three tests run successfully. However, on further inspection this is because the base URL of Odoo in the report HTML is still set to port 8069, so it goes through, but without loading the CSS. Not a problem for this test, but also not a real solution.

Do you have any idea why requests to Odoo don't complete during the test, and how to solve this?

pedrobaeza:
It's because HttpCase tests create a new cursor and there's a DB lock without releasing it when there's an error in a previous tests. It shouldn't happen if all goes right, but with any problem you will get to that stuck situation.

thomaspaulb:
@pedrobaeza But if that is true, shouldn't I see the error in the previous tests in the log? It's now green upon using the other port - if your explanation holds, it should be red.

pedrobaeza:
Well, I'm telling you my previous experiences, but @moylop260 knows better the internals of this.

delivery_carrier_label_default/models/stock.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
-# Copyright 2013-2016 Camptocamp SA
```
rousseldenis: You cannot remove exisitng author. Just add yourself to exisitng (new line)

delivery_carrier_label_default/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# © 2016 Sunflower IT (http://sunflowerweb.nl)
```
rousseldenis: Please add here original author.

delivery_carrier_label_default/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{'name': 'Default label for carrier labels',
+ 'version': '10.0.1.0.0',
+ 'author': "Sunflower IT,Camptocamp,Odoo Community Association (OCA)",
+ 'maintainer': 'Sunflower IT',
```
rousseldenis: s/maintainer/maintainers and value as list()

And please follow https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py

delivery_carrier_label_default/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+ 'category': 'Delivery',
```
rousseldenis: Please change this to 'Warehouse' or it does not appear in godd category in OCA shop.

moylop260:
Hi @thomaspaulb 

Let me try to explain it since that it a matter so complex for me.

The method `generate_default_label` of `delivery_carrier_label_default/models/stock.py` file has the following code:
 - `pdf_report = self.env['report'].get_pdf([self.id], 'delivery_carrier_label_default.report_default_label')`

It runs `wkhtmltopdf` command that requires get data from the self-instance but externally.
I mean, accessing from external requests.

So it transform your `odoo.tests.TransactionCase` (back-end) to `odoo.tests.HttpCase` (front-end) where it requires that odoo instance is running finally.

You can see the following wkhtmltopdf freeze command waiting for instance finish of load, but it is waiting for wkhtmltopdf at the same time.
 - wkhtmltopdf_error


Now, let's try using HttpCase where the tests is running after instance is entirely loaded:
 - httpcase

Conclusion your final patch should be:

```patch
diff --git a/delivery_carrier_label_default/tests/test_print_label.py b/delivery_carrier_label_default/tests/test_print_label.py
index 6fc34c8..71fac46 100644
--- a/delivery_carrier_label_default/tests/test_print_label.py
+++ b/delivery_carrier_label_default/tests/test_print_label.py
@@ -1,10 +1,10 @@
 # -*- coding: utf-8 -*-
 # © 2016 Sunflower IT (http://sunflowerweb.nl)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-from odoo.tests.common import TransactionCase
+from odoo.tests.common import HttpCase


-class TestPrintLabel(TransactionCase):
+class TestPrintLabel(HttpCase):

     def test_print_default_label(self):
         self.picking.generate_labels()
```

moylop260:
About using another port the issue is that it parameter is used just for second time of the command `odoo-bin`

I mean,
First time the default port is `8069`:
 - https://travis-ci.org/OCA/delivery-carrier/jobs/610902351#L524

So the database configuration parameters are stored using 8069 by default.

The second time the port is changed to `18079`
 - https://travis-ci.org/OCA/delivery-carrier/jobs/610902351#L1048
but configuration parameters are still stored using 8069

So  you won't have a freeze here since that the requests were refused but wkhtmltopdf error is muted.
You can see the following image using 18079 port for second odoo-bin command:
 - Screen Shot 2019-11-22 at 15 56 25

Notice a `ConnectionRefusedError` in `wkthmltopdf` but without a raise for odoo.

moylop260:
Maybe you are interested to see the old thread about:
 - https://github.com/OCA/maintainer-quality-tools/issues/509

thomaspaulb:
@moylop260 Thank you so much for this detailed explanation, very clear.

moylop260 (approved):
👍 
Technical review (no functional tests)

delivery_carrier_label_default/tests/test_print_label.py
```diff
@@ -1,18 +1,18 @@
 # -*- coding: utf-8 -*-
-# Copyright 2016 Camptocamp SA
```
rousseldenis: Same as above

rousseldenis (changes requested):
Code review

thomaspaulb:
@rousseldenis Thanks, did the fixes

rousseldenis:
Seems ok, could you squash your commits ?

thomaspaulb:
@rousseldenis If you give your final approval here, we can get this merged.

rousseldenis (approved):

PR #213, opened 2019-08-20
[10.0][IMP] stock_picking_delivery_info_computation: Allow to edit/compute volume on done pickings

pedrobaeza:
As this information remains isolated, you have both possibilities of recomputing or manually editing the field in any state.

cc @Tecnativa

### DIFF
--- a/stock_picking_delivery_info_computation/README.rst
+++ b/stock_picking_delivery_info_computation/README.rst
@@ -38,11 +38,10 @@ quantities, unles you introduce anything in *done* quantity, switching to this
 column in that case.
 
 The volume is auto-computed when the picking is generated from a sales order,
-or a backorder is created from original picking, but remains editable while
-the picking is not validated for any possible manual edition.
+or a backorder is created from original picking, but remains editable for any
+possible manual edition.
 
-You also have a button available while the picking is not validated for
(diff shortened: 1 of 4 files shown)


rousseldenis:
> As this information remains isolated, you have both possibilities of recomputing or manually editing the field in any state.

No needs of security feature on that (to avoid losing info - without recomputing) ?

pedrobaeza:
Well, that's something to comment apart, as this PR is not changing that. Only allowing to click or edit on any state. Any way, I don't think so, as picking are only editable by warehouse users, and both users and managers must be able to change this information IMO.

HviorForgeFlow:
If yo are going to let users modify this field, maybe it could be tracked in chatter when is not in draft.

pedrobaeza:
@hveficent that's also an extra feature, nothing to do with state itself. Field is not added by this module, but only put with the computation possibility. I would let it as is.

HviorForgeFlow (approved):

rafaelbn (approved):
Tested functionally :+1:

victor-ditay (approved):

pedrobaeza:
@rousseldenis can we merge then?

rousseldenis (approved):

PR #214, opened 2019-08-20
12 roulier

bealdav:


bealdav:
ping @hparfr

delivery_roulier/models/stock_picking.py
```diff
@@ -189,7 +189,8 @@ def _roulier_convert_address(self, partner):
 
         for tel in ['mobile', 'phone']:
             if address.get(tel):
-                address[tel] = address[tel].replace(u'\u00A0', '')
+                address[tel] = address[tel].replace(u'\u00A0', '').\
```
bealdav, line 192: le u'' n'est pas nécessaire en python3 @florian-dacosta

delivery_roulier/models/stock_picking.py
```diff
@@ -0,0 +1,249 @@
...
+
+        By default, the first account encoutered for this type.
+        Depending on your case, you may store it on the picking or
+        compute it from your business rules.
+
+        Accounts are resolved at runtime (can be != for dev/prod)
+        """
+        self.ensure_one()
+        domain = [
+            ("name", "=", self.carrier_id.delivery_type),
```
hbrunn: why don't you search for [`delivery_type`](https://github.com/OCA/delivery-carrier/blob/12.0/base_delivery_carrier_label/models/carrier_account.py#L20) here?

delivery_roulier/models/carrier_account.py
```diff
@@ -0,0 +1,27 @@
...
+
+    @api.constrains("name")
+    def name_constraint(self):
+        values = [
+            x[0] for x in
+            self.env["delivery.carrier"]._fields['delivery_type'].selection
+            if x[0] not in ("fixed", "base_on_rule")]
+        for rec in self:
+            if rec.name not in values:
+                raise UserError(MESSAGE % values)
```
hbrunn: see below, I don't believe we need this constraint

delivery_roulier/models/delivery_carrier.py
```diff
@@ -0,0 +1,45 @@
...
+class DeliveryCarrier(models.Model):
+    _inherit = 'delivery.carrier'
+
+    # module using roulier don't use native method to get labels
+    # pass False value to avoid failure.
+    def send_shipping(self, pickings):
+        res = super().send_shipping(pickings)
+        if not res:
+            return [{'exact_price': False, 'tracking_number': False}]
+
```
hbrunn, line 22:
```suggestion
        return res

```

delivery_roulier/models/delivery_carrier.py
```diff
@@ -0,0 +1,45 @@
...
+    def _is_roulier(self):
+        self.ensure_one()
+        return self.delivery_type in roulier.get_carriers()
+
+    def cancel_shipment(self, pickings):
+        if self._is_roulier:
+            return NotImplementedError()
```
hbrunn: shouldn't you raise this?

delivery_roulier/models/stock_move_line.py
```diff
@@ -0,0 +1,33 @@
...
+            product = self.product_id
+            ato = self.env['account.tax']
+            price_unit = ato._fix_tax_included_price_company(
+                product.list_price, product.taxes_id, ato,
+                self.picking_id.company_id)
+        return price_unit
+
+    def get_sale_order_line(self):
+        self.ensure_one()
+        return self.move_id and self.move_id.sale_line_id or False
```
hbrunn: this is very v7, why not just use the expression directly?

delivery_roulier/models/stock_picking.py
```diff
@@ -0,0 +1,249 @@
...
+
+        packages = self._get_packages_from_picking()
+        if len(packages) == 0:
+            raise UserError(_('No packages found for this picking'))
+        elif len(packages) == 1:
+            return packages.open_website_url()  # shortpath
+
+        # display a list of pickings
+        action = self.env.ref('stock.action_package_view').read()[0]
+        action['res_id'] = packages.ids
```
hbrunn: this doesn't do anything

delivery_roulier/models/stock_picking.py
```diff
@@ -0,0 +1,249 @@
...
+        action['domain'] = "[('id', 'in', [%s])]" % (
+            ",".join(map(str, packages.ids))
+        )
```
hbrunn: why don't you just pass the domain as python list?

delivery_roulier/models/stock_picking.py
```diff
@@ -0,0 +1,249 @@
...
+        action['context'] = "{'picking_id': %s }" % str(self.id)
```
hbrunn: and the context as dict?

delivery_roulier/views/carrier_account.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="view_roulier_carrier_account_tree" model="ir.ui.view">
+        <field name="model">carrier.account</field>
+        <field name="inherit_id" ref="base_delivery_carrier_label.carrier_account_view_tree"/>
+        <field name="arch" type="xml">
+            <field name="file_format" position="attributes">
+                <attribute name="invisible">1</attribute>
```
hbrunn, line 9: why this?

delivery_roulier/models/stock_picking.py
```diff
@@ -0,0 +1,249 @@
...
+        """
+        return self.company_id.partner_id
+
+    def _roulier_get_label_format(self, account):
+        """format of the label asked for carrier
+
+        Return:
+            label format (string)
+        """
+        return getattr(account, '%s_file_format' % self.delivery_type, None)
```
hbrunn, line 140: why don't you fall back to `account.file_format`?

hbrunn (changes requested):
this is nice, I'm going to add a few carriers to `roulier`.

@bealdav are you going to work on this any time soon? Otherwise, I could take over this PR and finish it by fixing my comments + CI and also add tests

delivery_roulier/models/carrier_account.py
florian-dacosta: I think you are right, this was before this was relevant before this commit : https://github.com/OCA/delivery-carrier/commit/f267d5bccdcade7125d79a5d5d0e056daeeaf3ee
I'll change this

delivery_roulier/models/delivery_carrier.py
florian-dacosta: Yes, I'll change this now

delivery_roulier/models/stock_move_line.py
florian-dacosta: No reason I guess, let's change this

delivery_roulier/models/stock_picking.py
florian-dacosta: I am not sure the `file_format` will be relevant for any specific carrier.
Maybe we could fallback on this, but we hide it in the view and I don't really see cases where the carrier will have the same choices.
For me this `file_type` may not be relevant on the `carrier.account`, because not generic enough.

delivery_roulier/models/stock_picking.py
florian-dacosta: Will remove this

delivery_roulier/models/stock_picking.py
florian-dacosta: no reason, I change it

delivery_roulier/models/stock_picking.py
florian-dacosta: No reason, I change this

delivery_roulier/views/carrier_account.xml
florian-dacosta: It is hard to have a generic selection of file_format. Each carrier has its own list, with its own labels, etc...
That is why, for each carrier, usually you should implement a `carrier_type_file_format` field to set your selection list and display it in the view.
But in this case, you don't want to see the `file_format` field, which is useless.

Of course, if we use the file_format as fallback,it becomes an issue, but once again, can we really use a standard list of multiple carriers?

delivery_roulier/models/stock_picking.py
florian-dacosta: Same remark as before, it was relevant before this commit : https://github.com/OCA/delivery-carrier/commit/f267d5bccdcade7125d79a5d5d0e056daeeaf3ee

Let's change it

florian-dacosta (commented):
Hi @hbrunn 

I'am gonna attend your comments today.
But yes, feels free to make a PR to fix CI or add tests, we'll try to review and merge it quickly.

I am not sure what you looked for, but, the roulier lib is only for python 2.7 for now.
We have worked for a migration to python 3+ and made some refactore to re-use some parts, etc...
So, the branch we are working on for python3 and with the refactore compatible with this version of `delivery_roulier` is : https://github.com/akretion/roulier/tree/py3-refactore-factory-system/roulier

We'll try to merge it soon and make tag for the new version, which will only work with python 3+
But i guess, that this is the branch you want to look if you want to add carriers.

florian-dacosta:
@hbrunn 

This odoo module should now really be working with master branch of roulier lib : https://github.com/akretion/roulier
It is still wip and need a lot of cleaning but it works with laposte France carrier.
Feels free to ask if you have any questions

hbrunn:
great, thanks a lot @florian-dacosta!

I'll provide my PR for the lib soon, and probably also a PR for the tests.

In the beginning my hope was that we just need `delivery_roulier`, and the Odoo addon supports any carrier supported by the library automatically. Now reading the latest version of the code, this seems more doable to me as input and output is more normalized.

What are your thoughts on this? I think if I contribute here some code that reads the available providers from the lib, and also the neccesary fields for authentication, we should be able to just have `delivery_roulier`, and not a specific module for every carrier.

So that in https://github.com/OCA/delivery-carrier/pull/214/files#diff-a31b22e2d839f6a61ecd7f240e3eef5dR161 we'd just say

```python
roulier.get(picking.carrier_id.roulier_carrier, 'get_label', payload)
```

assuming we'll add `carrier_type` ``roulier`` on `deliver.carrier` and when you select this, you also have to select `roulier_carrier`, which is filled from `roulier.get_carriers()`.

It's then still possible to override parts of the process in `delivery_roulier_some_carrier` if some carrier does something that doesn't fit this flow, they can still override the existing functions.

florian-dacosta:
> In the beginning my hope was that we just need delivery_roulier, and the Odoo addon supports any carrier supported by the library automatically. Now reading the latest version of the code, this seems more doable to me as input and output is more normalized.

This would be awsome of course and yes the goal is to normalize output/intput in order to minimize the specific carrier modules.
But it seems really complicated to have no odoo module at all for a specific carrier.

Here an example : https://github.com/akretion/delivery-carrier/tree/12-delivery-roulier-chronopost/delivery_carrier_chronopost
The module is small, but still, we have to pass more info to the library for it to be able to get the label because carriers has always specificities...
And this one is small, but for some others, there are more specifics...

So, make it possible to not have any specific odoo carrier module for it to work, I think it is a good idea.
But I think, only few carriers would be able to work this way.

> assuming we'll add carrier_type roulier on deliver.carrier and when you select this, you also have to select roulier_carrier, which is filled from roulier.get_carriers()

Yes, but we still would need to differenciate roulier carriers which need an Odoo module and those who does not.
If Odoo get the carrier type list from `roulier.get_carriers()` (`roulier.get_carriers_action_available()` now actually but we could add some method if needed). It will still fail if an extra module is necessary and this one is not available.And I think that the extra module will be needed in most cases.

Also, there are always specific data in carrier modules. We usually have multipe delivery methods, to ship internationally, nationally, in 2 days or more, etc...
So at least a data only module seems inevitable.

So I guess a list should be maintained somewhere in `delivery_roulier` with roulier carriers needing an extra module.

But anyway, I think we first  could try to have carrier modules with only this part https://github.com/akretion/delivery-carrier/blob/12-delivery-roulier-chronopost/delivery_carrier_chronopost/models/delivery_carrier.py and nothing else
If we succeed in having one or more of such small modules, of course we'll find a way to get rid of it

pedrobaeza:
That's why I stated at the beginning when @bealdav talked about the library of not doing that, as you are now spreading the maintenance across Odoo modules + python library, needing also specific test suite/infrastructure for such library. We have worked on some generalization, like

https://github.com/OCA/delivery-carrier/tree/12.0/delivery_state
https://github.com/OCA/delivery-carrier/tree/12.0/delivery_price_method
https://github.com/OCA/delivery-carrier/tree/12.0/delivery_package_number

But then, implement separate modules for each carrier using Odoo carrier "API":

https://github.com/OCA/l10n-spain/tree/12.0/delivery_seur
https://github.com/OCA/l10n-spain/tree/12.0/delivery_gls_asm

bealdav:
Thanks @hbrunn for your initiative.

hbrunn:
@pedrobaeza to me the advantage of an Odoo-agnostic library is that there's a chance to get input from other projects / be useful for other projects. Indeed this makes maintanance more complex, but that's worth the price in my opinion

@florian-dacosta I work on adding https://paazl.com to roulier, this is one of the meta carriers so this includes different carriers and their options, still I'm optimistic it's possible to have at least a few providers generically. We'd have to add a concept of options indeed to the library that this addon uses to generate the `delivery.carrier.options` relevant for the roulier carrier selected. To take care of carriers that need something specific I think we can have a list of known good (or bad) carriers, and only allow selecting the ones that work.

pedrobaeza:
I don't really such advantage, as that feedback from other places doesn't worth with the connection layer maintenance + dis-adjustments. We came from having SEUR integration in a Python library + connection in Odoo, and switch to the other one after seeing no advantage, but a lot of work.

bealdav:
@hbrunn I'll work on it

hbrunn:
fyi: I still think when we make the roulier library's input and output agnostic of the carrier and add some introspection, we should be able to handle most carriers only in the library, so that we don't have to touch the Odoo part at all to add a new carrier.

Unfortunately I lack the time to join the effort of doing so, so for v12 I'll provide some standalone carrier implementation as odoo modules.

hparfr:
>fyi: I still think when we make the roulier library's input and output agnostic of the carrier and add some introspection, we should be able to handle most carriers only in the library, so that we don't have to touch the Odoo part at all to add a new carrier.

That was and is the goal. In practice with the carriers we have implemented (Laposte Fr, DPD Fr (cargonet fr and geopost fr), Geodis fr, Kuhne & Nagel fr, and some improbable and offline fr carriers), they all have lot of service very different.

Between dpd (geopost) and laposte there is not that much differences, the specifics in odoo modules are tiny or simple. But when you look at geodis with all the complex edifact mechanisms you have a very different implementation.

The focus was also put in the harmonization of the label request but additional services likes address validation, state of shipments, pickup, customs declarations, quotes, not addressed today.

rousseldenis:
> > fyi: I still think when we make the roulier library's input and output agnostic of the carrier and add some introspection, we should be able to handle most carriers only in the library, so that we don't have to touch the Odoo part at all to add a new carrier.
> 
> That was and is the goal. In practice with the carriers we have implemented (Laposte Fr, DPD Fr (cargonet fr and geopost fr), Geodis fr, Kuhne & Nagel fr, and some improbable and offline fr carriers), they all have lot of service very different.
> 
> Between dpd (geopost) and laposte there is not that much differences, the specifics in odoo modules are tiny or simple. But when you look at geodis with all the complex edifact mechanisms you have a very different implementation.
> 
> The focus was also put in the harmonization of the label request but additional services likes address validation, state of shipments, pickup, customs declarations, quotes, not addressed today.

I think as said before the more we can do in a general library is a gain. That can attract other people to make it evolve.
@hparfr @sebastienbeau what about launching the request creating a repo for roulier library on contributors mailing list ?

PR #215, opened 2019-08-20
12 carrier gls WIP

bealdav:


chienandalu (commented):
Doesn't this belong to l10n-france repo?

pedrobaeza:
Yes, we have GLS in Spain, and have specific webservice, so it's hosted in OCA/l10n-spain: https://github.com/OCA/l10n-spain/pull/1392

bealdav:
Do you know if GLS are on the way to unify their web services or not ?

pedrobaeza:
We talked recently with ASM Spain, and they told us that ASM was incorporated to GSM, and they are using ASM's webservices for Spain for both with no intention on changing for now.

bealdav:
OK, thanks could you precise what is ASM and GSM ?

It's the same mess with GLS belgium it seems

pedrobaeza:
They were 2 different companies. ASM is a local Spanish carrier that was incorporated to GLS "group".

bealdav:
Thanks

PR #220, opened 2019-09-05
[12.0][base_delivery_carrier_label] Wait data encoded in base64 / write tracking number on package

florian-dacosta:
As there are no submodules migrated yet, I'd like to propose some small improvements in the module.

- Wait label file as base64 instead of encoding it.
=> Most of the time when using carrier webservice, we already receive base64. So it forces us to decode it, before sending it and the base_delivery_carrier_label encode it in base64 once more...
Example here : https://github.com/OCA/delivery-carrier/blob/10.0/delivery_carrier_label_postlogistics/models/stock.py#L87
It is not very good. If a carrier submodule have not-base64 data, it can easily encode it and it seems cleaner this way.

- Transform the field "file_type" on shipping labels from selection field to char field.
=> I am not sure this field is very usefull... But I think it should not be a selection field anyway.
Indeed, depending on the carrier, a lot of different format may be managed, and the selection fields force us to override the selection on each module to add the possible format.
Since the field is not even in any view, and it is always filled by a module, never in the UI, a char field seems fine, and more flexible.

- Write the tracking number on the package. 
Indeed, there is a field with this goal and when we have the information of package and tracking number, we should write it. It is also something done in every carrier submodules which could be done here. example : https://github.com/OCA/delivery-carrier/blob/10.0/delivery_carrier_label_postlogistics/models/stock.py#L111

@bealdav @hparfr @yvaucher @Timo17100-c2c 
Thanks for your feedback

### DIFF
--- a/base_delivery_carrier_label/models/shipping_label.py
+++ b/base_delivery_carrier_label/models/shipping_label.py
@@ -2,7 +2,7 @@
 # Copyright 2014 Akretion <http://www.akretion.com>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import api, fields, models
+from odoo import fields, models
 
 
 class ShippingLabel(models.Model):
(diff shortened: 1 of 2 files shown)


bealdav (approved):

hparfr (approved):
+1 for all the propositions.
I'm in favor of going further : remove file_type or mark it as deprecated.

yvaucher (approved):
:+1: for all changes

Setting the parcel_tracking seems indeed the way to go. I don't know why we weren't doing it.

PR #212, opened 2019-08-20
[12.0][MIG] base_delivery_carrier_label: Migration to 12.0

bealdav:
replace #205

PR #99, opened 2016-07-21
[9.0] port module delivery_carrier_label_dispatch renamed to delivery_carrier_label_batch

yvaucher:
- change name to batch to match with change of `picking_dispatch` name to
  `stock_batch_picking`
- mode description in README.rst
- move files in models and views directories
- change headers to small license header
- use api 8.0
- fix and improve tests

depends on:
- [x] https://github.com/OCA/stock-logistics-workflow/pull/241

PR #222, opened 2019-09-10
[12.0][MIG] delivery_carrier_label_batch

TDu:


rousseldenis:
I think the trend is to use queue jobs :

https://github.com/OCA/queue/tree/12.0/queue_job

@yvaucher @guewen What do you think as Camptocamp  was the original writer of this module ?

rousseldenis:
@TDu

tschanzt:
@rousseldenis I checked and the problem is he uses threads to generate each individual label, but he needs them to return the values for the merge. If we use queue job we don't have control to do this. There is a project by @guewen to add dependencies in queue jobs but it is stalled.

sebalix:
@rousseldenis @tschanzt I talked to @guewen and the future job dependencies API won't support a `join/await` operation (https://github.com/OCA/delivery-carrier/pull/222/files#diff-b755b9f4f2e7cdfb81fa56517273d8ecR161) of jobs from the main process (one difficulty is to get job records committed in the middle of a transaction to get them running...).
So we have to keep `threading` here for this use case.

sebalix:
@TDu some developments have been done on 9.0 since this PR has been created (check https://github.com/OCA/delivery-carrier/pull/99/commits)
The following commits are missing:

- aba3b156c9880d7509dd24b8943bf16802e56545
- c9ea6329284333bd86ada33c9419d9b3b8d598b5

EDIT: the PR https://github.com/OCA/stock-logistics-workflow/pull/570 has been merged

rousseldenis:
> @rousseldenis I checked and the problem is he uses threads to generate each individual label, but he needs them to return the values for the merge. If we use queue job we don't have control to do this. There is a project by @guewen to add dependencies in queue jobs but it is stalled.

Ok, I see.

The point that makes me questionning is why do we need to manage threads ? What is the bottleneck in generation ?

sebalix:
@rousseldenis we need threads to generate labels in parallel and not sequentially to speed up the process when batch owns a lot of pickings.

delivery_carrier_label_batch/wizard/generate_labels.py
```diff
@@ -86,9 +86,7 @@ def _do_generate_labels(self, group):
                 # generate the label of the pack
                 package_ids = [pack.id] if pack else None
                 try:
-                    picking.with_env(new_env).generate_labels(
-                        package_ids=package_ids
-                    )
+                    picking.with_env(new_env).action_generate_carrier_label()
```
sebalix, line 87: @tschanzt This is a performance bottleneck here, before we were generating the label for a given package (`package_ids` above) but now we are generating the labels for all the picking packages in a `for` loop.
I do not know all the code, it needs to be rethought probably.
<s>If you want I can work on it.</s>

tschanzt: @sebalix we stil have the threads per picking, so as we discussed in most cases the amount of packages per picking shouldn't be that high and doesn't require further measures.

sebalix (approved):

delivery_carrier_label_batch/wizard/generate_labels.py
```diff
@@ -0,0 +1,279 @@
...
+
+from ..pdf_utils import assemble_pdf
+from ..zpl_utils import assemble_zpl2
+
+_logger = logging.getLogger(__name__)
+
+
+class DeliveryCarrierLabelGenerate(models.TransientModel):
+
+    _name = 'delivery.carrier.label.generate'
```
sebalix, line 21: @tschanzt add `_description = "Generate labels from batch pickings"` to make runbot happy :)

delivery_carrier_label_batch/wizard/generate_labels.py
```diff
@@ -0,0 +1,280 @@
...
+        # create a new cursor to be up to date with what was created by workers
+        with self._do_in_new_env() as new_env:
+            # labels = new_env['shipping.label']
+            self_env = self.with_env(new_env)
+            labels = []
+            for pack, operations, label in self_env._get_packs(batch):
+                picking = operations[0].picking_id
+                if pack:
+                    label = self_env._find_pack_label(pack)
+                else:
```
i-vyshnevska:
```suggestion
                elif not label and picking:
```
@sebalix should this condition check for both cases?

delivery_carrier_label_batch/wizard/generate_labels.py
```diff
@@ -0,0 +1,280 @@
...
+                    label = self_env._find_picking_label(picking)
```
i-vyshnevska:
```suggestion
                    picking = operations[0].picking_id
                    label = self_env._find_picking_label(picking)
```

delivery_carrier_label_batch/wizard/generate_labels.py
```diff
@@ -0,0 +1,280 @@
...
+                        ) + '\n'.join(messages)
+            raise exceptions.UserError(message)
+
+        # create a new cursor to be up to date with what was created by workers
+        with self._do_in_new_env() as new_env:
+            # labels = new_env['shipping.label']
+            self_env = self.with_env(new_env)
+            labels = []
+            for pack, operations, label in self_env._get_packs(batch):
+                picking = operations[0].picking_id
```
i-vyshnevska:
```suggestion
```

TDu: Actually the whole for loop could be summarized as 
```
for pack, operations, label in self_env._get_packs(batch):
    picking = operations[0].picking_id
    if picking and not label:
        label = self_env._find_picking_label(picking)
    if not label:                                                                                                                                                                                                                                                                                            
        continue
     labels.append((label.file_type, label.attachment_id.datas))
```
Because the `_find_pack_label` is already called in `_get_packs`
As it is only a migration...

delivery_carrier_label_batch/wizard/generate_labels.py
yvaucher: IIRC we wants to enforce the use of packs and not try to print the labels if the packs are missings.

Testing locally having pickings without packs it wasn't working.

delivery_carrier_label_batch/wizard/generate_labels.py
```diff
@@ -0,0 +1,280 @@
...
+            key=lambda r: r.result_package_id.name or r.package_id.name
+        )
+        for pack, grp_operations in groupby(
+                operations,
+                key=lambda r: r.result_package_id or r.package_id):
+            pack_label = self._find_pack_label(pack)
+            yield pack, list(grp_operations), pack_label
+
+    @api.model
+    def _find_picking_label(self, picking):
```
yvaucher: Following the RFC here https://github.com/OCA/delivery-carrier/issues/192

And the commit
https://github.com/OCA/delivery-carrier/pull/212/commits/ffaab9796f157274c67ce78ee735e43a5128f36e

Pack must be enforced, thus the method `_find_picking_label` can be dropped.

delivery_carrier_label_batch/wizard/generate_labels.py
```diff
@@ -0,0 +1,280 @@
...
+
+        # create a cursor to be thread safe
+        with self._do_in_new_env() as new_env:
+            for pack, picking, label in group:
+                try:
+                    picking.with_env(new_env).action_generate_carrier_label()
+                except Exception as e:
+                    # add information on picking and pack in the exception
+                    picking_name = _('Picking: %s') % picking.name
+                    pack_num = _('Pack: %s') % pack.name if pack else ''
```
yvaucher, line 91: This most likely should be a single translated string. Depending on the language the result might not be correctly translatable. (Right-to-left languages)

An other option would be to enforce the use or line return instead of having it one line.

delivery_carrier_label_batch/wizard/generate_labels.py
```diff
@@ -0,0 +1,280 @@
...
+                    # pylint: disable=translation-required
+                    raise exceptions.UserError(
+                        ('%s %s - %s') % (picking_name, pack_num, e)
```
yvaucher, line 94:
```suggestion
                        ('%s\n%s\n\n%s') % (picking_name, pack_num, e)
```

delivery_carrier_label_batch/wizard/generate_labels.py
yvaucher: to remove

yvaucher (changes requested):
I'll propose a sub PR for the removal of find_picking_label.

The PR can be found here: https://github.com/TDu/delivery-carrier/pull/6


EDIT: added PR

delivery_carrier_label_batch/wizard/generate_labels.py
```diff
@@ -0,0 +1,314 @@
...
+                        'datas': codecs.encode(filedata, "base64"),
+                        'datas_fname': filename,
+                    }
+                    self.env['ir.attachment'].create(data)
+
+        return {
+            'type': 'ir.actions.act_window_close',
+        }
+
+
```
yvaucher: One line to remove

```suggestion

```

delivery_carrier_label_batch/wizard/generate_labels.py
```diff
@@ -0,0 +1,314 @@
...
+                            'name': filename,
+                            'res_id': batch.id,
+                            'res_model': 'stock.picking.batch',
+                            'datas': label['data'],
+                            'datas_fname': filename,
+                        }
+                        self.env['ir.attachment'].create(data)
+                else:
+                    labels_bin = [
+                        codecs.decode(label['data'], "base64") for label in labels if label
```
yvaucher:
```suggestion
                        codecs.decode(label['data'], "base64")
                        for label in labels if label
```

yvaucher:
tests of `delivery_auto_refresh` are broken, I don't see the reason here.

Investigating...

yvaucher:
unit tests of auto refresh fixed in #299

PR #226, opened 2019-09-27
[9.0] port module delivery_carrier_label_dispatch renamed to delivery_carrier_label_batch

sebalix:
Same than #99 but with support for ZPL labels merge + Travis/Lints fixed.

I tried to add tests for the ZPL case but it didn't work, and I think the current ones are also "broken" in some way (they are green but the labels used are not the ones created in the test case). It seems related to ir.attachments, didn't go further.


sebalix:
Commits included in https://github.com/OCA/delivery-carrier/pull/99

PR #228, opened 2019-10-02
[IMP] partner_delivery_zone: Sync delivery zone between order and picking

sergio-teruel:
cc @Tecnativa TT19894

### DIFF
--- a/partner_delivery_zone/models/sale_order.py
+++ b/partner_delivery_zone/models/sale_order.py
@@ -17,3 +17,16 @@ class SaleOrder(models.Model):
     def onchange_partner_shipping_id_delivery_zone(self):
         if self.partner_shipping_id.delivery_zone_id:
             self.delivery_zone_id = self.partner_shipping_id.delivery_zone_id
+
+    def write(self, vals):
+        # Update picking delivery zone if user update it in sale order that
+        # creates a picking,
+        res = super().write(vals)
+        if ('delivery_zone_id' in vals and not
+                self.env.context.get('skip_delivery_zone_update', False)):
+            self.mapped('picking_ids').with_context(
+                skip_delivery_zone_update=True
+            ).write({
+                'delivery_zone_id': vals['delivery_zone_id'],
+            })
+        return res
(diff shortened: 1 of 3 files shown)


partner_delivery_zone/models/sale_order.py
```diff
@@ -17,3 +17,16 @@ class SaleOrder(models.Model):
     def onchange_partner_shipping_id_delivery_zone(self):
         if self.partner_shipping_id.delivery_zone_id:
             self.delivery_zone_id = self.partner_shipping_id.delivery_zone_id
+
+    def write(self, vals):
+        # Update picking delivery zone if user update it in sale order that
+        # creates a picking,
+        res = super().write(vals)
+        if ('delivery_zone_id' in vals and not
```
pedrobaeza, line 25: IMO `delivery_zone_id` shouldn't be updateable in sales order once you have confirmed the sales order.

partner_delivery_zone/models/sale_order.py
```diff
@@ -17,3 +17,16 @@ class SaleOrder(models.Model):
     def onchange_partner_shipping_id_delivery_zone(self):
         if self.partner_shipping_id.delivery_zone_id:
             self.delivery_zone_id = self.partner_shipping_id.delivery_zone_id
+
+    def write(self, vals):
+        # Update picking delivery zone if user update it in sale order that
```
pedrobaeza, line 22: s/update/updates
s/sale/sales

partner_delivery_zone/models/sale_order.py
```diff
@@ -17,3 +17,16 @@ class SaleOrder(models.Model):
...
+        # creates a picking,
+        res = super().write(vals)
+        if ('delivery_zone_id' in vals and not
+                self.env.context.get('skip_delivery_zone_update', False)):
+            self.mapped('picking_ids').with_context(
```
pedrobaeza, line 27:
```suggestion
            self.picking_ids.with_context(
```

pedrobaeza (changes requested):
For me this is an error as is, as it leads to several inconsistencies:

1. You have a sales order with delivery zone A. Confirm and have a picking with zone A. Then transfer partially the picking, generating a new one picking. Then you change the zone to B in sales order. Both picking, the done one and the other will be changed to zone B!!
2. Then you can think on only writing the delivery zone on not done/cancel pickings, but then, you will have:

  * Sales order with zone B
  * One picking with zone A
  * Other picking with zone B

so the delivery zone in sales order doesn't represent the delivery zone on pickings.

For me, no sync should be done, and delivery zone mustn't be editable in sales order once it's confirmed.

As I know this is for reporting purposes, it should be accomplished other way.

sergio-teruel:
Ok. So If close It.

PR #229, opened 2019-10-04
[8.0][FIX] .travis: Update to latest template

rousseldenis:


pedrobaeza (approved):

.travis.yml
```diff
@@ -1,28 +1,56 @@
...
-
-python:
-  - "2.7"
-
 sudo: false
 cache: pip
 
+python:
+  # For branches <= 10.0, put `- "2.7.13"`
+  - "2.7.13"
```
pedrobaeza, line 7: It seems there's a problem with this specific version. Put only 2.7 again to see

rousseldenis: Nope that's the line with virtualenv

PR #231, opened 2019-10-17
[FIX] delivery_auto_refresh: Avoid error when confirming sales order

pedrobaeza:
As the used method now has a check for not being used when the order is confirmed, so it collapses when you want to add the delivery line in that moment.

@Tecnativa TT20031


rousseldenis (approved):
Code and functional review

rousseldenis:
@sergio-teruel @carlosdauden Maybe a review of this in order to merge?

carlosdauden (approved):

PR #233, opened 2019-10-23
[12.0][base_delivery_carrier_label] Add delivery type on carrier acount

florian-dacosta:
The goal is to be able to link a carrier account with a type of carrier.
Indeed, now we can only do this link with the name of the account if we put the same value than the delivery_type of the carrier, but it is not so intuitive and also too restrictive.
Indeed, we may need to have a unique name for carrier (for it to work with carrier_environment for instance) and this make it impossible to have multiple accounts for a carrier.

@bealdav @yvaucher @rousseldenis

### DIFF
--- a/base_delivery_carrier_label/models/carrier_account.py
+++ b/base_delivery_carrier_label/models/carrier_account.py
@@ -17,6 +17,10 @@ class CarrierAccount(models.Model):
                 ('XML', 'XML')]
 
     name = fields.Char(required=True)
+    delivery_type = fields.Selection(
+        selection=lambda self: self.env["delivery.carrier"]
+        ._fields["delivery_type"].selection,
+        help="This field may be used to link an account to a carrier")
     account = fields.Char(string='Account Number', required=True)
     password = fields.Char(string='Account Password', required=True)
     company_id = fields.Many2one(comodel_name="res.company", string="Company")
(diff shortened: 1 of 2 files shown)


bealdav (approved):

base_delivery_carrier_label/models/carrier_account.py
```diff
@@ -17,6 +17,10 @@ def _selection_file_format(self):
                 ('XML', 'XML')]
 
     name = fields.Char(required=True)
+    delivery_type = fields.Selection(
```
rousseldenis, line 20: @florian-dacosta Maybe this one should come with a related if you have a 'carrier_id' field ?

florian-dacosta: @rousseldenis 
Actually, if we had a `carrier_id` field, we would not even need this delivery type, it would be enought to find the carrier account during label generation.
But the thing is, a carrier account generally concern multiple `delivery.carrier`.

Indeed, for a same carrier, we may have multiple `delivery.method` : 
- Expedition for European Union
- Expedition for the country
-Expedition for the reste of the world
- Expedition in a replay point
- ...

All this delivery method use the same carrier account and have the same `delivery_type`

yvaucher (approved):
LGTM

rousseldenis (approved):

yvaucher:
@rousseldenis seems the merge failed silently. Retrying.

PR #236, opened 2019-11-28
[MIG][12.0] partner_delivery_schedule: Migration to v12.0

sergio-teruel:
cc @Tecnativa TT20121
@ernestotejeda @rafaelbn review please!!


partner_delivery_schedule/models/partner_delivery_schedule.py
```diff
@@ -0,0 +1,75 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class DeliverySchedule(models.Model):
+    _name = 'delivery.schedule'
```
pedrobaeza, line 8:
```
WARNING openerp_test odoo.models: The model delivery.schedule has no _description
```

pedrobaeza (approved):

sergio-teruel:
@pedrobaeza Changes done!!

ernestotejeda (commented):
Code review

partner_delivery_schedule/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    'name': 'Partner Delivery Schedule',
+    'summary': 'Set on partners a schedule for delivery goods',
+    'version': '12.0.1.0.1',
```
ernestotejeda, line 6:
```suggestion
    'version': '12.0.1.0.0',
```

ernestotejeda (commented):
Maybe you can implement this later by having in re.partner a Many2one with resource.calendar instead of a Many2many with delivery.schedule and so we will have one less model to worry about and maybe the allow_delivery_date method could be shorter.
If it is not done in the migration of this module it could be done as an improvement later. What do you think ?

PR #241, opened 2019-12-18
[11.0][ADD] stock_picking_report_delivery_cost: Delivery cost in Stock Picking Report

pedrobaeza:
This module shows in the delivery slip and picking operations reports the cost of the delivery method according the rate computed on the fly according same conditions as standard `delivery` module would do.

Usage
=====

To use this module, you need to:

* Go to *Inventory > Operations > Transfers* and pick one not in state *Done*  or *Cancelled*.
* Click on *Print > Delivery Slip* or *Print > Picking Operations*.
* You will see in the report the computed delivery cost.

@Tecnativa TT21061


stock_picking_report_delivery_cost/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import models
+from . import tests
```
sergio-teruel: import tests not needed

stock_picking_report_delivery_cost/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2019 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Delivery cost in Stock Picking Report',
```
sergio-teruel: The module shows cost info in delivery slip and picking operation reports... so the name and summary would be more generic...

stock_picking_report_delivery_cost/models/stock_picking.py
```diff
@@ -0,0 +1,27 @@
...
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    currency_id = fields.Many2one(
+        related="sale_id.currency_id",
+        readonly=True,
+        string="Currency",
+        related_sudo=True,  # for avoiding access problems
+    )
```
sergio-teruel, line 15: Why do you create this field if you don't use it later in the reports?

sergio-teruel (changes requested):

stock_picking_report_delivery_cost/models/stock_picking.py
pedrobaeza: `fields.Monetary` uses it automatically. I will put a comment about that.

sergio-teruel: Ohh!! Ok, I do not know it

stock_picking_report_delivery_cost/__init__.py
pedrobaeza: True!

stock_picking_report_delivery_cost/__manifest__.py
pedrobaeza: Well, it's generic right now, but not plural. Let's put `Delivery cost in Picking Reports` (any way, my original idea was to put it only in delivery slip, but end it up putting it on both as sometimes the other is the preferred one).

chienandalu (approved):
Looks good. Although I couldn't test it properly in runbut due to a PdfWriter error that looks completly unrelated
image

pedrobaeza:
@sergio-teruel changes done

sergio-teruel (approved):

PR #243, opened 2020-01-10
[12.0] Add .gitattributes for merge automatically OCA_dependencies.txt file

jcoux:


gurneyalex (approved):

rousseldenis:
@jcoux @gurneyalex I'm not so much in favor of this as it could lead to duplicate entries (e.g.: two lines for sale-workflow and a branch on it if contributor added it to make PR green).

Moreover, if we do so, we should generalize it.

yvaucher:
Just one remark there, this will be useful only in case of manual merging, github doesn't do it automatically. Not sure about the OCA bot. Maybe something to define on the OCA bot?

We could sort this out with a little more complexe merge policy. As merge policy could be extended with scripts. We would need something like add all, sort and keep unique.

@sbidoul any opinion on this, do you think it's the role of the ocabot try to make the rebase on it's own? Thus needing us to give it strong directives?
Personally I would say that as long as we can identify something that will be repeated often and for which we identify the workflow/boundaries we want to automate it.

sbidoul:
The ocabot merge strategy could be improved to cope with that, but...

I still think `oca_dependencies.txt` is redundant, as well as `requirements.txt` because all dependencies are expressed in the manifests. And we can easily get rid of both of them by installing addons to test with pip: https://github.com/OCA/maintainer-quality-tools/pull/343

So I'd rather address the root cause of the issue (redundant files) than add automation complexity to cope with them.

yvaucher (approved):
I think the incentive here was to be able to merge multiple pending merge in local project without the hassle of having an unnecessary conflict when adding various PR that adds modules that each adds dependencies.

So in the mean time I'm ok with this change

sbidoul:
@yvaucher FYI, installation of dependencies is now [implemented](https://github.com/OCA/maintainer-quality-tools/blob/d44f30e6a29a7dd21735aceab018ada9d1592133/travis/travis_install_nightly#L119-L136) in MQT. It is opt-in: to activate it you need to set [MQT_DEP=PIP](https://github.com/OCA/oca-addons-repo-template/blob/3fa9c3924aa38e5dd9db3670009c7843a5db7e50/.travis.yml.jinja#L73) in .travis.yml environment variables.

PR #245, opened 2020-01-20
[FIX] delivery_auto_refresh: cost calculation on fixed carrier_type

gbaumgartner-callino:
When module `delivery_auto_refresh` ist installed and you try to confirm a stock picking with a fixed price carrier you'll get the following message:` "No price rule matching this order; delivery cost cannot be computed."`

This happens because `_get_price_from_picking` can only get costs if there are rules defined in the carrier. Fixed cost carriers don't have any rules hence cost calculation will fail.

So before getting costs via rules it should check if this is a fixed type carrier - if it's the case we can simply write  the `fixed_price` to `so_line`.

### DIFF
--- a/delivery_auto_refresh/i18n/de.po
+++ b/delivery_auto_refresh/i18n/de.po
@@ -6,7 +6,7 @@
 # OCA Transbot <transbot@odoo-community.org>, 2018
 msgid ""
 msgstr ""
-"Project-Id-Version: Odoo Server 10.0\n"
+"Project-Id-Version: Odoo Server 12.0\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2018-03-13 03:45+0000\n"
 "PO-Revision-Date: 2018-03-13 03:45+0000\n"
--- a/delivery_auto_refresh/models/stock_picking.py
+++ b/delivery_auto_refresh/models/stock_picking.py
@@ -37,7 +37,10 @@ class StockPicking(models.Model):
         total = sale_order.currency_id._convert(
             total, sale_order.company_id.currency_id, sale_order.company_id,
             sale_order.date_order)
-        so_line.price_unit = self.carrier_id._get_price_from_picking(
-            total, weight, volume, quantity,
-        )
+        if self.carrier_id.carrier_type == "fixed":
+            so_line.price_unit = self.carrier_id.fixed_price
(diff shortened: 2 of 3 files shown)


pedrobaeza:
Thanks for the patch. Indeed, there's that problem.

Can you check Travis?

pedrobaeza:
Can you add a test for avoiding the regression in the future or is it too much to ask?

gbaumgartner-callino:
Started making the test right after I saw the codecov messages. Will be added soon.

pedrobaeza:
Great, thanks!

pedrobaeza:
Any news about this?

gbaumgartner-callino:
Yes, sorry for the delay. Tests are added now, i hope they'll be fine - first tests i wrote for odoo tbh.

rousseldenis:
@gbaumgartner-callino Could you check Travis ?

gbaumgartner-callino:
> TypeError: _auto_refresh_delivery() got an unexpected keyword argument 'force'

seems to be the major issue, seems unrelated to this issue.

https://github.com/OCA/delivery-carrier/commit/6a01ca619eb00825b3ba2a96655e8adb3c3fd1a7 adds the force parameter to the `_auto_refresh_delivery` call, yet the method itself has no such param defined.

Seems the current 12.0 branch is failing as well.

rousseldenis:
@pedrobaeza Is this commit comes from a PR or just pushed in 12 branch ?

https://github.com/OCA/delivery-carrier/commit/6a01ca619eb00825b3ba2a96655e8adb3c3fd1a7

pedrobaeza:
Cherry picked from v11

rousseldenis:
Ok, a PR could be good as implementation has changed apparently. Do you take car of fixing this ?

pedrobaeza:
Uhm, I don't remember that change, but I'll check.

pedrobaeza:
You can rebase now that I have pushed 31fc996cd7d6e748e9de08b41079872b82164a24

pedrobaeza:
Travis is not notifying me now about failing CIs in pushed branches...

gbaumgartner-callino:
Alright, looks good!

pedrobaeza:
There's a conflict. Can you rebase and fix it for getting this finally working? Also please squash all your commits.

PR #254, opened 2020-02-20
[12.0][FIX+IMP] delivery_multi_destination: View + available carriers + shipping

pedrobaeza:
* [IMP] Improve view for hiding undesired elements

  On a multi-destination carrier, that fields shouldn't be shown.

* [FIX] available carriers must return the main one
   
  When testing available carriers, we must return the main one if one of the children matches, not returning the children itself.

* [FIX] Make delivery shipping properly

  Handle the step of sending the shipping (triggered on picking validation) for multi destination delivery. This means to manually check for fixing prices (as the implementation doesn't check this part and always take the main carrier price, not the subcarrier one.

  Test for this use case done, although the rest of the cases are not covered by tests yet.

cc @Tecnativa TT22325

### DIFF
--- a/delivery_multi_destination/models/delivery_carrier.py
+++ b/delivery_multi_destination/models/delivery_carrier.py
@@ -2,7 +2,8 @@
 # Copyright 2017 Luis M. Ontalba <luis.martinez@tecnativa.com>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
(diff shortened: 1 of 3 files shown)


pedrobaeza:
@carlosdauden @ernestotejeda please review this

ernestotejeda (approved):

pedrobaeza:
Let's move this

/ocabot merge minor

PR #255, opened 2020-03-01
[FIX] partner_delivery_zone: Fix compute methods for all records

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/partner_delivery_zone/models/sale_order.py
+++ b/partner_delivery_zone/models/sale_order.py
@@ -19,5 +19,9 @@ class SaleOrder(models.Model):
     @api.depends("partner_shipping_id")
     def _compute_delivery_zone_id(self):
         for so in self:
-            if so.partner_shipping_id.delivery_zone_id:
-                so.delivery_zone_id = so.partner_shipping_id.delivery_zone_id
+            partner = (
+                so.partner_shipping_id
+                if so.partner_shipping_id.type == "delivery"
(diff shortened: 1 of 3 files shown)


partner_delivery_zone/models/sale_order.py
```diff
@@ -19,5 +19,4 @@ class SaleOrder(models.Model):
     @api.depends("partner_shipping_id")
     def _compute_delivery_zone_id(self):
         for so in self:
-            if so.partner_shipping_id.delivery_zone_id:
-                so.delivery_zone_id = so.partner_shipping_id.delivery_zone_id
+            so.delivery_zone_id = so.partner_shipping_id.delivery_zone_id
```
pedrobaeza: That `if` was done on purpose if I don't remember bad, so you may need:
```suggestion
            if so.partner_shipping_id.delivery_zone_id:
                so.delivery_zone_id = so.partner_shipping_id.delivery_zone_id
            else:
                so.delivery_zone_id = so.delivery_zone_id  # reassign old value
```
(not sure this works)

Tardo:
Check Travis

victoralmau:
I think that everything it's OK

victoralmau (approved):

pedrobaeza (approved):
Not sure about not preserving old value if new delivery address has no delivery zone, but let's see.

PR #259, opened 2020-03-25
[12.0] new module carrier_deutsche_post

jans23:
This module allows to purchase shipping label from Deutsche Post (German Post) directly in Odoo. Several "products" of Deutsche Post can be configured.


carrier_deutsche_post/models/stock.py
```diff
@@ -0,0 +1,88 @@
+from odoo import _, api, fields, models
+from odoo.exceptions import Warning
+
+
+class Picking(models.Model):
+    _inherit = 'stock.picking'
+
+    label_de_attach_id = fields.Many2one('ir.attachment', 'Label Deutsche Post', copy=False)
+    carrier_type = fields.Selection(related='carrier_id.type', string='Deutsche Post Carrier Type', readonly=True)
```
rousseldenis, line 9: IMHO It has to be moved in a more generic module

PR #262, opened 2020-04-23
[9.0][IMP] delivery_carrier_label_batch: ZPL2 single images

grindtildeath:
If using ZPL2 file format, multiple *.zpl can be merged either "as is" or with
"single image definition" to spare file size.
Single image definition can be used by defining `ir.config_parameter` with name
`zpl2.assembler.single.images` to `True`.

### DIFF
--- a/delivery_carrier_label_batch/data/ir.config_parameter.xml
+++ b/delivery_carrier_label_batch/data/ir.config_parameter.xml
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<odoo>
-    <record id="zpl2_assembler_single_images" model="ir.config_parameter">
-        <field name="key">zpl2.assembler.single.images</field>
-        <field name="value">False</field>
-    </record>
-</odoo>
+<openerp>
+    <data>
+        <record id="zpl2_assembler_single_images" model="ir.config_parameter">
+            <field name="key">zpl2.assembler.single.images</field>
+            <field name="value">False</field>
+        </record>
+    </data>
+</openerp>
--- a/delivery_carrier_label_batch/data/ir.config_parameter.xml
+++ b/delivery_carrier_label_batch/data/ir.config_parameter.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <record id="zpl2_assembler_single_images" model="ir.config_parameter">
+        <field name="name">zpl2.assembler.single.images</field>
+        <field name="value">False</field>
+    </record>
+</odoo>
--- a/delivery_carrier_label_batch/data/ir.config_parameter.xml
+++ b/delivery_carrier_label_batch/data/ir.config_parameter.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <odoo>
     <record id="zpl2_assembler_single_images" model="ir.config_parameter">
-        <field name="name">zpl2.assembler.single.images</field>
+        <field name="key">zpl2.assembler.single.images</field>
         <field name="value">False</field>
     </record>
 </odoo>
--- a/delivery_carrier_label_batch/data/ir.config_parameter.xml
+++ b/delivery_carrier_label_batch/data/ir.config_parameter.xml
@@ -1,9 +1,14 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <openerp>
-    <data>
+    <data noupdate="1">
         <record id="zpl2_assembler_single_images" model="ir.config_parameter">
             <field name="key">zpl2.assembler.single.images</field>
             <field name="value">False</field>
         </record>
+        <record id="zpl2_batch_merge" model="ir.config_parameter">
+            <field name="key">zpl2.batch.merge</field>
+            <field name="value">False</field>
+        </record>
+
     </data>
 </openerp>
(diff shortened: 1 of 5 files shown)


delivery_carrier_label_batch/zpl_utils.py
```diff
@@ -3,6 +3,77 @@
...
 
 
+def assemble_zpl2_single_images(files):
+    """Assemble a list of ZPL2 files with single image definition."""
+    zpl_images = dict()
+    zpl_labels = dict()
+    # Extract all the images and labels from different files
+    for file_nr, zpl_file in enumerate(files):
+        # ^XA stands for label start
+        images_def, label_def = zpl_file.split("^XA")
```
gurneyalex, line 13: we are sure to always get exactly 1 "^XA" string in the file?

delivery_carrier_label_batch/wizard/generate_labels.py
```diff
@@ -277,6 +277,12 @@ def _concat_files(self, file_type, files):
         if file_type == 'pdf':
             return assemble_pdf(files)
         if file_type == 'zpl2':
-            return assemble_zpl2(files)
+            zpl2_single_images = self.env['ir.config_parameter'].get_param(
```
sebalix: No need to call `safe_eval` on it? I can't see a call of it in https://github.com/odoo/odoo/blob/9.0/openerp/addons/base/ir/ir_config_parameter.py#L57-L75 so I suppose that even if `"False"` is set we are evaluating the value to true.

delivery_carrier_label_batch/zpl_utils.py
```diff
@@ -3,6 +3,77 @@
...
+    #  3. Existing image with this ref is not the same image > Change the ref
+    #     for image and update refs in related label
+    #  4. There is a same existing image with different ref > Update the refs
+    #     in related label
+    to_add_images = dict()
+    for file_nr, file_zpl_images in zpl_images.items():
+        for zpl_img in file_zpl_images:
+            img_header, img_content = zpl_img.split(".GRF")
+            img_number = img_header.lstrip("~DGR:")
+            if img_content not in to_add_images.keys():
```
gurneyalex:
```suggestion
            if img_content not in to_add_images:
```

delivery_carrier_label_batch/zpl_utils.py
```diff
@@ -3,6 +3,77 @@
...
+                    to_add_images[img_content] = img_number
+                else:
+                    # Case 3.
+                    # Handle different image for same image numbers
+                    max_img_number = max(to_add_images.values())
+                    # Remove IMG form max_number_name in order to increment
+                    img_prefix = max_img_number[:3]
+                    digit = max_img_number[3:]
+                    # Define new img number
+                    new_digit = str(int(digit + 1))
```
gurneyalex: bug here: digit is a string, and you cannot add a number to a string. 
```suggestion
                    new_digit = str(int(digit) + 1)
```

delivery_carrier_label_batch/zpl_utils.py
```diff
@@ -3,6 +3,77 @@
...
+                    new_img_number = img_prefix + new_digit
+                    # change img number reference in label
+                    to_add_images[img_content] = new_img_number
+                    file_label = zpl_labels.get(file_nr)
```
gurneyalex: if you are garanteed to have file_nr as key in the dictionary, then use zpl_labels[file_nr]. Otherwise be prepared to handle file_label being None. 
```suggestion
                    file_label = zpl_labels[file_nr]
```

delivery_carrier_label_batch/zpl_utils.py
```diff
@@ -3,6 +3,77 @@
...
+                    file_label.replace(img_number, new_img_number)
+                    zpl_labels[file_nr] = file_label
+            elif img_number != to_add_images.get(img_content):
+                # Case 4
+                # Handle different image numbers for same image
+                file_label = zpl_labels.get(file_nr)
+                file_label.replace(img_number, to_add_images.get(img_content))
```
gurneyalex: same remark as above regarding get() and the possible None value.

gurneyalex (commented):
I would like to see an automated test, this will ease further evolution of the module and the code is quite complex.

delivery_carrier_label_batch/zpl_utils.py
grindtildeath: It must be so, as "^XA" defines the start of a label: https://www.zebra.com/content/dam/zebra/manuals/printers/common/programming/zpl-zbi2-pm-en.pdf (page 347)

However, looking into this I'm realizing that graphics (or images) could be defined in many different ways (cf page 158 in PDF above), and this code is based on the zpl labels we get from the postlogistics webservice...

So looking at this it seems I should rework the feature to target graphics through "DGx" instead of "~DGR:IMG" :sob: 
Or we could add a hook in `delivery_carrier_label_batch` and move this into the postlogistics module, I'm not sure what's best, but I don't think we can manage all the cases defined by zpl specs anyway...

guewen:
fix of lint + squash required

delivery_carrier_label_batch/zpl_utils.py
guewen: Let's start with the case we know, can be adapted if another need comes later.

sebalix (approved):

delivery_carrier_label_batch/data/ir.config_parameter.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
```
grindtildeath:
```suggestion
    <data noupdate="1">
```

PR #263, opened 2020-05-19
[ADD]base delivery carrier extension

zahravelasco:
This module adds a button on delivery orders to generate a label as an
attachement. This module doesn't do anything by itself, it serves as a
base module for other carrier-specific modules.

### DIFF
--- a/base_delivery_carrier_extensions/__manifest__.py
+++ b/base_delivery_carrier_extensions/__manifest__.py
@@ -1,0 +1,19 @@
+# © 2020 FactorLibre - Zahra Velasco <zahra.velasco@factorlibre.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Base delivery carrier extension',
+    'version': '11.0.1.0.0',
+    'depends': [
+        'delivery',
+        'base_delivery_carrier_label'
+    ],
+    'category': 'Stock',
+    'author': 'FactorLibre',
+    'license': 'AGPL-3',
+    'website': 'http://www.factorlibre.com',
+    'data': [
+        'views/stock_view.xml'
+    ],
+    'installable': True,
+    'application': False
+}
--- a/base_delivery_carrier_extensions/models/stock_picking.py
+++ b/base_delivery_carrier_extensions/models/stock_picking.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+# © 2020 FactorLibre - Zahra Velasco <zahra.velasco@factorlibre.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import api, fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = 'stock.picking'
+
+    date_shipped = fields.Date('Shipment Date', readonly=True)
+    date_delivered = fields.Datetime('Delivery Date', readonly=True)
+    tracking_status = fields.Char('Tracking status', readonly=True, index=True)
+    tracking_status_history = fields.Text(
+        'Tracking status history',
+        readonly=True)
+    carrier_status_picking = fields.Selection([
+        ('shipping_recorded_in_carrier', 'shipping recorded in carrier'),
+        ('in_transit', 'In transit'),
+        ('canceled_shipment', 'Canceled shipment'),
+        ('incidence', 'incidence'),
+        ('customer_delivered', 'Customer delivered'),
+        ('warehouse_delivered', 'Warehouse delivered')
+    ], string="Carrier State", readonly=True)
+
+    @api.multi
+    def carrier_tracking_status(self):
+        pass
+
+    @api.multi
+    def carrier_delivery_create(self):
+        self.ensure_one()
+        self.carrier_status_picking = 'shipping_recorded_in_carrier'
+
+    @api.multi
+    def carrier_delivery_cancel(self):
+        self.ensure_one()
+        self.carrier_status_picking = 'canceled_shipment'
--- a/base_delivery_carrier_extensions/views/stock_view.xml
+++ b/base_delivery_carrier_extensions/views/stock_view.xml
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- © 2020 Zahra Velasco <zahra.velasco@factorlibre.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). -->
+<odoo>
+    <record id="view_picking_withcarrier_out_form" model="ir.ui.view">
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="base_delivery_carrier_label.view_picking_withcarrier_out_form"/>
+            <field name="arch" type="xml">
+                <xpath expr="//page//group[@name='carrier_data']/.." position="after">
+                    <group name="tracking_status_group" colspan="4" col="3"
+                            attrs="{'invisible': [('carrier_tracking_ref', '=', False)]}">
+                    <field name="date_shipped"/>
+                    <field name="date_delivered"/>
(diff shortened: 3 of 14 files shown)


base_delivery_carrier_extensions/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Base delivery carrier extension',
+    'version': '11.0.1.0.0',
+    'depends': [
+        'delivery',
+        'base_delivery_carrier_label'
+    ],
+    'category': 'Stock',
+    'author': 'FactorLibre',
```
rlizana, line 11: Añade como autor a  "Odoo Community Association (OCA)",

base_delivery_carrier_extensions/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# © 2020 FactorLibre - Zahra Velasco <zahra.velasco@factorlibre.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Base delivery carrier extension',
+    'version': '11.0.1.0.0',
+    'depends': [
+        'delivery',
+        'base_delivery_carrier_label'
```
rlizana, line 8: falta una coma al final

base_delivery_carrier_extensions/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# © 2020 FactorLibre - Zahra Velasco <zahra.velasco@factorlibre.com>
```
pedrobaeza, line 1:
```suggestion
# Copyright 2020 FactorLibre - Zahra Velasco <zahra.velasco@factorlibre.com>
```
It is stated in copyright laws that the whole word is needed, so we have changed this all around

base_delivery_carrier_extensions/__manifest__.py
pedrobaeza, line 8: Why this dependency?

base_delivery_carrier_extensions/__manifest__.py
pedrobaeza, line 11: Include `OCA (Odoo Community Association)` for index purposes

base_delivery_carrier_extensions/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    'name': 'Base delivery carrier extension',
+    'version': '11.0.1.0.0',
+    'depends': [
+        'delivery',
+        'base_delivery_carrier_label'
+    ],
+    'category': 'Stock',
+    'author': 'FactorLibre',
+    'license': 'AGPL-3',
+    'website': 'http://www.factorlibre.com',
```
pedrobaeza, line 13: It should be the OCA GitHub repository one (you can put FL website in CONTRIBUTORS)

```suggestion
    'website': 'http://github.com/OCA/delivery-carrier',
```

base_delivery_carrier_extensions/models/stock_picking.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
```
pedrobaeza, line 1: This line is no longer needed for P3

base_delivery_carrier_extensions/models/stock_picking.py
```diff
@@ -0,0 +1,37 @@
...
+# © 2020 FactorLibre - Zahra Velasco <zahra.velasco@factorlibre.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import api, fields, models
```
pedrobaeza, line 4:
```suggestion
from odoo import api, fields, models
```

base_delivery_carrier_extensions/models/stock_picking.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class StockPicking(models.Model):
+    _inherit = 'stock.picking'
+
+    date_shipped = fields.Date('Shipment Date', readonly=True)
+    date_delivered = fields.Datetime('Delivery Date', readonly=True)
+    tracking_status = fields.Char('Tracking status', readonly=True, index=True)
```
pedrobaeza, line 12:
```suggestion
    tracking_state = fields.Char('Tracking status', readonly=True, index=True)
```

base_delivery_carrier_extensions/models/stock_picking.py
```diff
@@ -0,0 +1,37 @@
...
+    tracking_status_history = fields.Text(
```
pedrobaeza, line 13: Should we provide an extra model `delivery.tracking.status` that acts similarly to cost price in products (auto-saving a line when this is changed)?

base_delivery_carrier_extensions/models/stock_picking.py
```diff
@@ -0,0 +1,37 @@
...
+        'Tracking status history',
+        readonly=True)
+    carrier_status_picking = fields.Selection([
```
pedrobaeza, line 16:
```suggestion
    carrier_state = fields.Selection([
```

base_delivery_carrier_extensions/views/stock_view.xml
```diff
@@ -0,0 +1,22 @@
...
+<!-- © 2020 Zahra Velasco <zahra.velasco@factorlibre.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). -->
+<odoo>
+    <record id="view_picking_withcarrier_out_form" model="ir.ui.view">
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="base_delivery_carrier_label.view_picking_withcarrier_out_form"/>
+            <field name="arch" type="xml">
+                <xpath expr="//page//group[@name='carrier_data']/.." position="after">
+                    <group name="tracking_status_group" colspan="4" col="3"
+                            attrs="{'invisible': [('carrier_tracking_ref', '=', False)]}">
```
pedrobaeza, line 11: Isn't better to hide it if no `carrier_state`?

pedrobaeza (changes requested):

pedrobaeza:
Apart from the module name we have talked about in the chat.

base_delivery_carrier_extensions/__manifest__.py
rlizana: It's not necesary

base_delivery_carrier_extensions/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    'data': [
+        'views/stock_view.xml'
+    ],
+    'installable': True,
```
azucenatrey, line 17: 'installable' value is True by default in 12.0 version, you can delete it:
https://www.odoo.com/documentation/12.0/reference/module.html

base_delivery_carrier_extensions/models/stock_picking.py
rlizana: it's complex, the state lines can be rewritten by the carrier

base_delivery_carrier_extensions/__manifest__.py
pedrobaeza: This is one of the cases where I prefer to apply the Python zen "Explicit better than implicit", but it's only a matter of taste

base_delivery_carrier_extensions/models/stock_picking.py
pedrobaeza: OK, it was just an idea. Maybe we are talking about different things... I'm talking about saving the moment when `carrier_state` changes from one state to another. We can get something similar putting to the field `track_visibility="onchange"`, but it's less accessible. Let's do at least this.

zahravelasco:
could you check if the PR it's ok?

PR #265, opened 2020-05-20
[12.0][IMP] partner_delivery_zone: Get delivery zone from commercial partner for non delivery contacts

sergio-teruel:
cc @Tecnativa TT23898


rafaelbn (approved):
👍

chienandalu (approved):

PR #270, opened 2020-06-30
[12.0][ADD] delivery_price_method

rlizana:
This module adds additional functions that will be necessary for developments carriers integratión.

It provides a system to rate the shipping price based on a fixed price or based on rules.


delivery_price_method/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,4 @@
+Tis module adds additional functions that will be necessary for the operator
```
chienandalu:
```suggestion
This module adds additional functions that will be necessary for the operator
```

delivery_price_method/readme/USAGE.rst
```diff
@@ -0,0 +1,3 @@
+Only available in carriers that have been developed using this module, please consult yours carriers addons.
```
chienandalu:
```suggestion
Only available in carriers that have been developed using this module, please consult your carriers addons.
```

chienandalu (approved):
Thanks @rlizana Tested and code review. Minor remarks:

delivery_price_method/models/delivery_carrier.py
```diff
@@ -0,0 +1,32 @@
...
+from odoo import fields, models
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = 'delivery.carrier'
+
+    # You must add next method to your carrier model if want use this price
+    # method
+    #
+    # def [CARRIER]_rate_shipment(self, order):
```
pedrobaeza: The only point that I don't see is this. Why do you need to add that method? I think it can be override in the proper place for avoiding to define the method.

pedrobaeza (changes requested):
@rlizana do you want me to do the needed changes for what I state?

delivery_price_method/models/delivery_carrier.py
rlizana: Some providers compute the shipping price by webservice, this method avoids overwriting core functionality (https://github.com/OCA/OCB/blob/12.0/addons/delivery/models/delivery_carrier.py#L109), allows the developer to get the control of compute method, and allows providers coexist with webservice compute and static methods (fixed price or based on rules) with the same functionality.

pedrobaeza: But if you select one of these options, carrier webservice compute method doesn't make sense. That's why I want to override the method (in a compatible way), for not needing to query webservice and not needing this extra method at all.

rlizana: It's not an additional method, the <my_provider>_rate_shipment method is proposed by the core, it's one of the "magic methods" of delivery.carrier model  (https://github.com/OCA/OCB/blob/12.0/addons/delivery/models/delivery_carrier.py#L25).

If your provider allows the compute by webservice, you will have to consume the service in this method. 
Otherwise, you have to create the <my_provider>_rate_shipment method with the logic to compute the method indicated in price_method.

pedrobaeza: Yes, but that method should be added for querying the webservice, not for calling other methods. Imagine you have SEUR carrier, and you select that one for a delivery method. My mental idea is:

- You don't select any price method (although I would prefer to add a new selection value called `Webservice` or similar for being more explicit): the magic method is called and the webservice fetches data.
- You select fixed or based on rules, then magic method is ignored, and the price is returned from its base methods.
- The same applies if you have another delivery carrier without magic method implemented, because there's no webservice: you shouldn't need to add it pointing to the other one.

Do you get me? If you want, I can mount on top of your code my idea.

rlizana: I understand you perfectly, something like this.
Do you think this is more simple and clean?

```python
from odoo import _, fields, models


    _inherit = 'delivery.carrier'

    price_method = fields.Selection(
        selection=[
            ('fixed', 'Fixed price'),
            ('base_on_rule', 'Based on Rules'),
        ],
        string='Price method',
        default='fixed',
    )

    def rate_shipment(self, order):
        self.ensure_one()
        if hasattr(self, '%s_rate_shipment' % self.delivery_type):
            return super().rate_shipment(order)
        res = getattr(self, '%s_rate_shipment' % self.price_method)(order)
        res['price'] = float(
            res['price']) * (1.0 + (float(self.margin) / 100.0))
        order_total = order._compute_amount_total_without_delivery()
        if res['success'] and self.free_over and order_total >= self.amount:
            res['warning_message'] = _(
                'Info:\nThe shipping is free because the order amount exceeds '
                '%.2f.\n(The actual shipping cost is: %.2f)') % (
                    self.amount, res['price'])
            res['price'] = 0.0
        return res

    def send_shipping(self, pickings):
        res = super().send_shipping(pickings)
        rates = getattr(self, '%s_send_shipping' % self.price_method)(pickings)
        for index, rate in enumerate(rates):
            res[index]['exact_price'] = rate['exact_price']
        return res
```

pedrobaeza: Not exactly, but also when coding for showcasing you my idea, I have slightly changed the idea for being even more compatible, allowing extension modules to be compatible with this one. Please see the result in:

https://github.com/Tecnativa/delivery-carrier/commit/6ccad80494f4ebaa3ee01729f38fffda0f5a4a2f

I didn't want to push it here for not abusing, but if you see it OK, I do it and finally merge this.

With this code,  `delivery_seur` doesn't have to depend in this module for working the way you want.

chienandalu (commented):
I like @pedrobaeza 's approach. Thumbs up for it :+1:

delivery_price_method/models/delivery_carrier.py
rlizana: It's perfect!! clean and simple, thank @pedrobaeza  :+1: 
Please, push the code, and I remove depends in delivery_seur

pedrobaeza (approved):

pedrobaeza:
Thanks all!

/ocabot merge nobump

PR #271, opened 2020-07-10
[10.0][ADD] delivery_carrier_category

rousseldenis:


rousseldenis:
@Cedric-Pigeon

delivery_carrier_category/models/delivery_carrier.py
```diff
@@ -0,0 +1,22 @@
...
+from odoo import api, fields, models
+
+
+class DeliveryCarrier(models.Model):
+
+    _inherit = 'delivery.carrier'
+
+    @api.model
+    def _get_default_category_id(self):
+        return self.env['delivery.carrier.category'].search([], limit=1)
```
legalsylvain, line 14: don't understand this function, that seems to return a random category.

delivery_carrier_category/security/delivery_carrier_category.xml
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" encoding="utf-8"?>
```
legalsylvain, line 1: csv file seems the norm in Odoo and OCA modules for the ir.model.access records. (more synthetic, easy to edit).
(Just a remark / not a blocking point).

setup/delivery_carrier_category/.eggs/README.txt
```diff
@@ -0,0 +1,6 @@
+This directory contains eggs that were downloaded by setuptools to build, test, and run plug-ins.
```
legalsylvain, line 1: I think that such file in ``.eggs`` folder, should not be added on github. Don't you think ?

delivery_carrier_category/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    'category': 'Warehouse',
+    'maintainers': ['rousseldenis'],
+    'depends': [
+        'delivery',
+    ],
+    'data': [
+        'views/delivery_carrier.xml',
+        'security/delivery_carrier_category.xml',
+        'views/delivery_carrier_category.xml',
+        'data/delivery_carrier_category.xml',
```
legalsylvain, line 23: this file seems to be a demo data file.

legalsylvain (commented):
some minor remarks. LGTM Otherwise. thanks.

delivery_carrier_category/__manifest__.py
rousseldenis: Nope. I did that first but that seems logical to have it in data. Can be archived if user wants.

delivery_carrier_category/models/delivery_carrier.py
rousseldenis: As categories are ordered by sequence, by default it will load the first (but indeed, in case of equal sequences, it will take the first by id)

delivery_carrier_category/security/delivery_carrier_category.xml
rousseldenis: @legalsylvain Tastes and colors... :smiley: I'd prefer this. Don't like inline editing, here columns are explicitly written (and I think it's not explicit in OCA as it let dev free to choose one or another).

victoralmau (approved):

acsonefho (approved):
LGTM

PR #274, opened 2020-07-20
[12.0][ADD] delivery_package_number: New module

chienandalu:
This module adds a number of packages field that can be either be set manually by the user or computed when the delivery packages flow is used.

This field can be used by delivery carrier extensions that need such info to rate the shipment, print the labels, etc.

cc @Tecnativa TT20983


delivery_package_number/README.rst
```diff
@@ -0,0 +1,91 @@
...
+    :target: https://github.com/OCA/delivery/tree/12.0/delivery_package_number
+    :alt: OCA/delivery
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/delivery-12-0/delivery-12-0-delivery_package_number
+    :alt: Translate me on Weblate
+
+|badge1| |badge2| |badge3| |badge4| 
+
+This module adds a number of packages field that can be either be set manually
+by the user or computed when the delivery packages flow is used.
```
pedrobaeza:
```suggestion
This module recovers the number of packages field that was present in previous
versions of Odoo, that can be either be set manually by the user, but also allows
to be computed when the delivery packages flow is used.
```

delivery_package_number/README.rst
```diff
@@ -0,0 +1,91 @@
...
+
+If no delivery packages are used:
+
+#. Go to an open picking and click in the *Additional Info* tab.
+#. In the *Delivey Information* section you'll find a **Number of packages**
+   field that you can edit.
+
+If delivey packages are used:
+
+#. The field will be computed depending on the delivery packages used in the
```
pedrobaeza: I think the field should always remain writable. The only change using delivery packages is that the field gets auto-filled when you change these packages, but you should be able to change the number of packages anyway. Example of use case: imagine you set packages for an specific product, but the rest are put randomly in other packages and you want to set this manually.

delivery_package_number/models/stock_picking.py
```diff
@@ -0,0 +1,20 @@
...
+from odoo import api, fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    number_of_packages = fields.Integer(
+        string="Number of Packages",
+        store=True,
+        readonly=False,
```
pedrobaeza: This is not valid for v12, so you need to regular field + onchanges/writes.

pedrobaeza (changes requested):

pedrobaeza:
@rlizana what do you think about this new "extraction" for handling properly package number without heavy dependencies and giving more features to it?

delivery_package_number/models/stock_picking.py
chienandalu: The thing is package_ids is not a regular field, but a computed one, so a regular onchange wasn't working but this way it does (as you can see in the tests as well)

pedrobaeza: I don't think this works properly in UI, as this is not supported till v13. I'm talking about onchange + write with a regular one.

chienandalu: I didn't dive deep into the causes... but it worked both in tests and UI :sweat_smile: 

Changed anyway :)

pedrobaeza (approved):

pedrobaeza:
@rlizana do you give your blessings?

rlizana:
I seem to me a good idea to divide the compute of volume and package numbers into two addons.
A useful and new feature is add the number of packages field in the picking validation wizard, what you think?

yes, you have my blessings :yum:

:+1:

pedrobaeza:
Yes it can be good. @chienandalu can you do it?

chienandalu:
The problem is that there are multiple possible wizards when you validate the picking: the backorder wizard, the inmediate transfer wizard, the overprocessed qty wizard... and even no wizard when the initial demand fits the done qtys...

pedrobaeza:
I suggest the following:

- Put the field in a mixin and a related to the picking.
- Add code for writing back the value to the corresponding related the field if filled and different.
- Inherit stock.picking and the transients from the mixin.
- Add the relateds
- Add the field in views.

Is it too much work or does it worth? If not, this can go to roadmap.

rlizana:
I think it is simpler. Only the immediate transfer wizard is needed. 
If the user has create done lines, modify the  number of packages in the picking will not be much trouble.

The reality is that if everything works well, the user will go through the immediate transfer wizard. If there is a pending amount, the backorder wizard is shown later, and it is not necessary to add the field.

pedrobaeza:
OK, agreed

chienandalu:
Ok then, I'll do so :)

chienandalu:
@pedrobaeza @rlizana Changes done!

pedrobaeza:
Great! Let's go for the last PRs to merge:

/ocabot merge nobump

PR #269, opened 2020-06-22
[10.0][delivery_carrier_label_postlogistics] Street1 printed before S…

ajaniszewska-dev:
…treet2

Impacting module: delivery_carrier_label_postlogistics

Aim of this PR is to print street1 before street2 on shipping label. To do that, keys in a recipient dictionary inside __prepare_recipient_() method were exchanged.

partner:
image

Result:
[shipping_label.pdf](https://github.com/OCA/delivery-carrier/files/4812624/shipping_label.pdf)

PR #275, opened 2020-07-21
[10.0][delivery_carrier_label_postlogistics] Street1 printed before S…

santostelmo:
Street1 printed before Stree2

This PR intends to fix error message introduced by the following PR:
https://github.com/OCA/delivery-carrier/pull/269

"[E2001] A valid recipient address must be stated."

This happens when `Street` is send with empty string.

### DIFF
--- a/delivery_carrier_label_postlogistics/postlogistics/web_service.py
+++ b/delivery_carrier_label_postlogistics/postlogistics/web_service.py
@@ -166,15 +166,15 @@ class PostlogisticsWebService(object):
         partner_name = partner.name or partner.parent_id.name
         recipient = {
             'Name1': partner_name,
-            'Street': partner.street,
+            'Street': partner.street2 or partner.street,
             'ZIP': partner.zip,
             'City': partner.city,
             'Country': partner.country_id.code,
             'EMail': partner.email or None,
         }
 
-        if partner.street2:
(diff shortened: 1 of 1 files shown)


delivery_carrier_label_postlogistics/postlogistics/web_service.py
```diff
@@ -166,15 +166,15 @@ def _prepare_recipient(self, picking):
         partner_name = partner.name or partner.parent_id.name
         recipient = {
             'Name1': partner_name,
-            'Street': partner.street,
+            'Street': partner.street2 or partner.street,
```
yvaucher, line 169: I have some strong doubt about the order of those two fields.

yvaucher (commented):
Isn't it more logic to have street as address and street2 as address suffix?

It's a suffix not a prefix.

yvaucher (changes requested):
@santostelmo this seems too specific to go in there. If I understand correctly, you have case of partners where `street2` is filled but not `street` ? Which goes again the core logic of Odoo

This requires more information to be accepted in the OCA module.

You might want to have this as a specific module instead. As we need to keep the comunity version generic.

santostelmo:
It can be closed, it was a dev done with a very specific purpose.

PR #276, opened 2020-07-27
[12.0][ADD] New addon delivery_carrier_sale_label_note

victoralmau:
It allows defining at the stock_picking level notes for the carrier that will be used for integrations with carriers.
Has dependency on base_delivery_carrier_label (https://github.com/OCA/delivery-carrier)

Proposed in the issue: OCA/stock-logistics-workflow#684


rousseldenis:
Hi @victoralmau, thanks for your contribution. IMHO, 'sale' should appear in module name as the note is really stored in sale order.  Maybe delivery_carrier_sale_label_note

victoralmau:
> Hi @victoralmau, thanks for your contribution. IMHO, 'sale' should appear in module name as the note is really stored in sale order. Maybe delivery_carrier_sale_label_note

Ok thanks, I've already renamed the addon

PR #277, opened 2020-08-12
[IMP] delivery_state: automated action

chienandalu:
- Add automated action for checking the delivery state in providers with
the proper methods configured.
- Added docstrings
- Improved tracking state views usability toggling visibility just when
it's needed.
- Improved usage description.
- Removed demo data in tests that can lead to unexpected CI issues.

cc @Tecnativa TT20983

### DIFF
--- a/delivery_state/README.rst
+++ b/delivery_state/README.rst
@@ -35,11 +35,46 @@ ones carrier gives us.
 .. contents::
    :local:
 
+Configuration
+=============
+
+A scheduled action for automating the tracking update for these pickings can be
+configured going to *Settings > Technical > Scheduled Actions* and then choosing
+*Update deliveries states*. It will update the pending delivery states for the
+pickings with service providers with tracking methods configured, and in pending
+state (not delivered or cancelled).
+
 Usage
 =====
 
-* Go to Inventory / Operations and open an out picking
-* In "Additional Info" tab you can see delivery state
+Depending on the delivery service provider, the state tracking could be more or
+less complete, since it could have or not the necessary API calls implemented.
+
+With regular methods (fixed, based on rules):
+
+  #. Go to Inventory / Operations and open an outgoing pending picking.
+  #. In the *Additional Info* tab, assign it a delivery carrier which is fixed or
+     based on rules.
+  #. Validate the picking and you'll see in the same tab the delivery state
+     info with the shipping date and the shipping state.
+
+When service provider methods are implemented, we can follow the same steps as
+described before, but we'll get additionally:
+
+  #. In the *Additional Info* tab, we'll see button *Update tracking state* to
+     manually query the provider API for tracking updates for this expedition.
+  #. Depending on the stated returned by the provider, we could get these
+     states (field *Carrier State*):
+
+        * Shipping recorded in carrier
+        * In transit
+        * Canceled shipment (finished)
+        * Incidence
+        * Warehouse delivered
+        * Customer delivered (finished)
+  #. In the field *Tracking state* we'll get the tracking state name given by
+     the provider (which is mapped to the ones in this module)
+  #. In the field *Tracking history* we'll get the former states log.
 
 Bug Tracker
 ===========
@@ -59,6 +94,7 @@ Authors
 
 * Trey (www.trey.es)
 * FactorLibre
+* Tecnativa
 
 Contributors
 ~~~~~~~~~~~~
@@ -70,6 +106,10 @@ Contributors
 * `FactorLibre <https://www.factorlibre.com>`_:
 
   * Zahra Velasco <zahra.velasco@factorlibre.com>
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+  * Pedro M. Baeza
+  * David Vidal
 
 Maintainers
 ~~~~~~~~~~~
--- a/delivery_state/models/stock_picking.py
+++ b/delivery_state/models/stock_picking.py
@@ -1,7 +1,8 @@
 # Copyright 2020 Trey, Kilobytes de Soluciones
 # Copyright 2020 FactorLibre
+# Copyright 2020 Tecnativa - David Vidal
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class StockPicking(models.Model):
@@ -40,7 +41,35 @@ class StockPicking(models.Model):
     )
 
     def tracking_state_update(self):
-        for picking in self.filtered(lambda p: p.delivery_type):
+        """Call to the service provider API which should have the method
+        defined in the model as:
+            <my_provider>_tracking_state_update
+        It can be triggered manually or by the cron."""
+        for picking in self.filtered("carrier_id"):
             method = '%s_tracking_state_update' % picking.delivery_type
             if hasattr(picking.carrier_id, method):
+                getattr(picking.carrier_id, method)(picking)
+
+    @api.model
+    def _update_delivery_state(self):
+        """Automated action to query the delivery states to the carriers API.
+        every carrier should implement it 's own method. We split them by
+        delivery type so only those carries with the method update"""
+        pickings = self.search([
+            ("state", "=", "done"),
+            ("delivery_state", "not in",
+                ["customer_delivered", "canceled_shipment"]),
+            # These won't ever autoupdate, so we don't want to evaluate them
+            ("delivery_type", "not in", [False, "fixed", "base_one_rule"]),
+        ])
+        delivery_types = self.mapped("delivery_type")
+        # Split them by delivery type so we can ignore those without the
+        # proper method.
+        for delivery_type in delivery_types:
+            method = "%s_tracking_state_update" % delivery_type
+            if not hasattr(self.env["delivery.carrier"], method):
+                continue
+            delivery_type_pickings = pickings.filtered(
+                lambda x: x.delivery_type == delivery_type)
+            for picking in delivery_type_pickings:
                 getattr(picking.carrier_id, method)(picking)
--- a/delivery_state/tests/test_delivery_state.py
+++ b/delivery_state/tests/test_delivery_state.py
@@ -21,15 +21,29 @@ class TestDeliveryState(TransactionCase):
             'product_id': product_shipping_cost.id,
             'fixed_price': 99.99,
         })
+        self.product = self.env["product.product"].create({
+            "name": "Test product",
+            "type": "product",
+        })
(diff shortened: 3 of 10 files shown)


chienandalu:
ping @rlizana

delivery_state/models/stock_picking.py
```diff
@@ -40,7 +41,35 @@ class StockPicking(models.Model):
...
     def tracking_state_update(self):
-        for picking in self.filtered(lambda p: p.delivery_type):
+        """Call to the service provider API wich should have the method
+        definedin the manner. Can be triggered manually or by cron:
+        <my_provider>_tracking_state_update"""
+        for picking in self.filtered("delivery_type"):
             method = '%s_tracking_state_update' % picking.delivery_type
             if hasattr(picking.carrier_id, method):
-                getattr(picking.carrier_id, method)(picking)
+                return getattr(picking.carrier_id, method)(picking)
```
rlizana: if you use return you interrupt the loop

delivery_state/models/stock_picking.py
```diff
@@ -40,7 +41,35 @@ class StockPicking(models.Model):
...
+        """Automated action to query the delivery states to the carriers API.
+        every carrier should implement it 's own method. We split them by
+        delivery type so only those carries with the method update"""
+        pickings = self.search([
+            ("state", "=", "done"),
+            ("delivery_state", "not in",
+                ["customer_delivered", "canceled_shipment"]),
+            # These won't ever autoupdate, so we don't want to evaluate them
+            ("delivery_type", "!=", False),
+            ("delivery_type", "not in", ["fixed", "base_one_rule"]),
```
rlizana: you can combine the two conditions into one 
`("delivery_type", "not in", [False, "fixed", "base_one_rule"]),`

delivery_state/models/stock_picking.py
```diff
@@ -40,7 +41,35 @@ class StockPicking(models.Model):
...
+        # Split them by delivery type so we can ignore those without the
+        # proper method.
+        for delivery_type in delivery_types:
+            method = "%s_tracking_state_update" % delivery_type
+            if not hasattr(self.env["delivery.carrier"], method):
+                continue
+            delivery_type_pickings = pickings.filtered(
+                lambda x: x.delivery_type == delivery_type)
+            for picking in delivery_type_pickings:
+                return getattr(picking.carrier_id, method)(picking)
```
rlizana: if you use return you interrupt the loop

delivery_state/tests/test_delivery_state.py
```diff
@@ -21,15 +21,20 @@ def setUp(self):
...
+        self.partner = self.env["res.partner"].create({
+            "name": "Mr. Odoo",
+        })
 
     def test_delivery_state(self):
-        product = self.env.ref('product.product_delivery_01')
-        partner = self.env.ref('base.res_partner_12')
         sale = self.env['sale.order'].create({
-            'partner_id': partner.id,
+            'partner_id': self.partner.id,
```
rlizana, line 42: just of curiosity...  any reason for this change?

rlizana (changes requested):

delivery_state/models/stock_picking.py
chienandalu: you're absolutly right :+1:

delivery_state/tests/test_delivery_state.py
chienandalu: Well, we detected a problem in CIs of environments with default localizations. In this case, an EUR conversion from dollars is being made. So we try to avoid unexpected side effects creating the test data from scratch.  And I think it will be needed to create a pricelist as well, as it still makes the conversion...

delivery_state/README.rst
```diff
@@ -38,8 +38,40 @@ ones carrier gives us.
...
 
-* Go to Inventory / Operations and open an out picking
-* In "Additional Info" tab you can see delivery state
+Depending on the delivery service provider, the state tracking could be more or
+less complete, since it could have or not the necessary API calls implemented.
+
+With regular methods (fixed, based on rules):
+
+  #. Go to Inventory / Operations and open an outgoing pending picking.
+  #. In the *Additional Info* tab, assign it a delivery carrier wich is fixed or
```
pedrobaeza:
```suggestion
  #. In the *Additional Info* tab, assign it a delivery carrier which is fixed or
```

delivery_state/README.rst
```diff
@@ -38,8 +38,40 @@ ones carrier gives us.
...
+     states (field *Carrier State*):
+
+        * Shipping recorded in carrier
+        * In transit
+        * Canceled shipment (finished)
+        * Incidence
+        * Warehouse delivered
+        * Customer delivered (finished)
+  #. In the field *Tracking state* we'll get the tracking state name given by
+     the provider (wich is mapped to the ones in this module)
```
pedrobaeza:
```suggestion
     the provider (which is mapped to the ones in this module)
```

delivery_state/README.rst
```diff
@@ -38,8 +38,40 @@ ones carrier gives us.
...
+  #. In the field *Tracking history* we'll get the former states log.
+
+A scheduled action for automate the tracking update for these pickings can be
```
pedrobaeza:
```suggestion
A scheduled action for automating the tracking update for these pickings can be
```

And you should put this in CONFIGURE.rst section

delivery_state/models/stock_picking.py
```diff
@@ -40,7 +41,34 @@ class StockPicking(models.Model):
     )
 
     def tracking_state_update(self):
-        for picking in self.filtered(lambda p: p.delivery_type):
+        """Call to the service provider API wich should have the method
```
pedrobaeza:
```suggestion
        """Call to the service provider API, which should have the method
```

delivery_state/models/stock_picking.py
```diff
@@ -40,7 +41,34 @@ class StockPicking(models.Model):
...
+        definedin the manner. Can be triggered manually or by cron:
```
pedrobaeza:
```suggestion
        defined in the model. Can be triggered manually or by the cron:
```

delivery_state/models/stock_picking.py
```diff
@@ -40,7 +41,34 @@ class StockPicking(models.Model):
...
+        <my_provider>_tracking_state_update"""
+        for picking in self.filtered("delivery_type"):
```
pedrobaeza: Avoid to navigate to a related field that is required in carrier.
```suggestion
        for picking in self.filtered("carrier_id"):
```

delivery_state/models/stock_picking.py
```diff
@@ -40,7 +41,34 @@ class StockPicking(models.Model):
...
+    def _update_delivery_state(self):
+        """Automated action to query the delivery states to the carriers API.
+        every carrier should implement it 's own method. We split them by
+        delivery type so only those carries with the method update"""
+        pickings = self.search([
+            ("state", "=", "done"),
+            ("delivery_state", "not in",
+                ["customer_delivered", "canceled_shipment"]),
+            # These won't ever autoupdate, so we don't want to evaluate them
+            ("delivery_type", "not in", [False, "fixed", "base_one_rule"]),
```
pedrobaeza, line 63: Field is required, so no need to test `False` value:
```suggestion
            ("delivery_type", "not in", ["fixed", "base_one_rule"]),
```

delivery_state/models/stock_picking.py
```diff
@@ -40,7 +41,34 @@ class StockPicking(models.Model):
...
+        ])
+        delivery_types = self.mapped("delivery_type")
```
pedrobaeza, line 65: Instead of this, that requires several filters that are heavier than checking each time `hasattr`, I prefer a `read_group` grouped by `delivery_type`, and then process each group.

delivery_state/views/stock_picking_views.xml
```diff
@@ -2,20 +2,25 @@
...
-                <field name="tracking_state_history" colspan="3"/>
+            <xpath expr="//group[@name='carrier_data']" position="after">
+                <group name="delivery_tracking" attrs="{'invisible': ['|', ('state', '!=', 'done'), ('carrier_id', '=', False)]}">
+                    <field name="date_shipped"/>
+                    <field name="date_delivered"/>
+                    <field name="delivery_state"/>
+                    <field name="tracking_state" class="oe_inline"/>
+                    <button name="tracking_state_update" string="Update tracking state"
+                            type="object" class="oe_inline"
+                            attrs="{'invisible': [('state', 'in', ['customer_delivered', 'canceled_shipment'])]}" />
```
pedrobaeza: Make it invisible also for fixed or based on rules.

pedrobaeza (changes requested):

delivery_state/models/stock_picking.py
chienandalu: It can be `False` if no carrier is defined in the picking (i.e.: internal, receptions)

delivery_state/views/stock_picking_views.xml
chienandalu: One thing the module does is to put those pickings in state _Shipping recorded in carrier_, wich maybe it's a bit misleading, as there's no possibility to push the state further... maybe for those, the state when shipped should be simply `customer_delivered`. What do you think, @rlizana ?

delivery_state/models/stock_picking.py
chienandalu: The problem with that approach `delivery_type` is a related field, so a group_by can't be made with it :/  Anyway, the real time cost will be in the API queries, IMO. Also, we won't get much records (hundreds in an optimistic scenario, but dozens in the average one).

delivery_state/models/stock_picking.py
pedrobaeza: OK, right.

delivery_state/models/stock_picking.py
pedrobaeza: It depends on the delivery volume you have, but OK, I see that is a non stored, so you can't group.

delivery_state/views/stock_picking_views.xml
pedrobaeza: I think it's OK to be on that state, as we don't know really if they are delivered or not.

chienandalu: Ok then :)

pedrobaeza (approved):

delivery_state/tests/test_delivery_state.py
rlizana: Thanks for the explanation. Good to know.

rlizana (approved):

PR #282, opened 2020-08-25
[13.0] Add delivery_package_fee

guewen:
A list of Package Fees can be added on shipping methods.
When a outgoing transfer is done, for each package fee configured on the
shipping method, a new sale order line is created with:

* The product selected on the Package Fee
* The product name with the number of the transfer in the line's description
  (e.g. "Service Fee (WH/OUT/00036)")
* The quantity equal to the number of packages in the transfer
* The unit price equal to the price set on the product's pricelist (so it can be
  different per customer and even have different pricing depending on the number
  of packages)
* The taxes configured on the product, fiscal position applies if any.

Package Fee lines are added only if their quantity and price is above zero.

### DIFF
--- a/delivery_package_fee/tests/test_package_fee.py
+++ b/delivery_package_fee/tests/test_package_fee.py
@@ -1,0 +1,397 @@
+# Copyright 2020 Camptocamp
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import Form, SavepointCase
+from odoo.tools import mute_logger
+
+
+class TestPackageFee(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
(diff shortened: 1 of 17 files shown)


jgrandguillaume (approved):
Quick look seems good to me

sebalix (approved):

delivery_package_fee/tests/test_package_fee.py
```diff
@@ -0,0 +1,396 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import Form, SavepointCase
+from odoo.tools import mute_logger
+
+
+class TestPackageFee(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
simahawk, line 11: tracking_disable :pray:

simahawk: done

PR #284, opened 2020-08-30
WIP FIXME TESTME Add delivery_routing

Tonow-c2c:
# WIP :warning:

### DIFF
--- a/delivery_routing/__init__.py
+++ b/delivery_routing/__init__.py
@@ -1,0 +1,1 @@
+from . import wizard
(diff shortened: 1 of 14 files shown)


pedrobaeza:
@max3903 I think you are looking to achieve something similar, so you should join efforts IMO.

yvaucher:
@pedrobaeza @max3903 

@Tonow-c2c don't work anymore at camptocamp nor on Odoo projects, thus this PR is dead. I'll close it, if we have the need to fullfil it we will do it in an other PR (not sure which internal project it was about)

PR #286, opened 2020-09-10
12.0 FIX base_deliv_label: use valid qty to compute weight of move.line

bealdav:
small fix

cc @hbrunn @florian-dacosta @hparfr @sebastienbeau @pedrobaeza

### DIFF
--- a/base_delivery_carrier_label/models/stock_move_line.py
+++ b/base_delivery_carrier_label/models/stock_move_line.py
@@ -40,8 +40,10 @@ class StockMoveLine(models.Model):
                     'Type conversion not implemented for product %s' %
                     product.id)
                 cant_calc_total = True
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I wonder if instead this change, the problem is when this is called (before being validated instead of after).

bealdav:
Hi @pedrobaeza good question. We have many clients using this module with several versions. Some of them have real good reasons (I don't remember exactly which) to generate labels before or after picking done. Odoo exists for that. Make everybody happy with its marginal behavior ;-)

pedrobaeza (approved):
OK then :)

hbrunn (approved):

marcelsavegnago (approved):

PR #288, opened 2020-09-10
[ADD] base_delivery_carrier_label: active flag for options

hbrunn:
For carriers with a lot of options that aren't all relevant for a specific deployment, it's helpful to be able to hide uninteresting options

### DIFF
--- a/base_delivery_carrier_label/views/delivery.xml
+++ b/base_delivery_carrier_label/views/delivery.xml
@@ -35,6 +35,7 @@
     <field name="arch" type="xml">
       <form string="delivery_carrier_option">
         <group col="4">
+          <field name="active" />
           <field name="mandatory" attrs="{'readonly': [('readonly_flag','=',True)]}"/>
           <field name="by_default" attrs="{'invisible': [('mandatory', '=', True)]}"/>
           <field name="readonly_flag" attrs="{'invisible': True}"/>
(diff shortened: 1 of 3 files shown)


StefanRijnhart:
Can we show inactive options on the carrier form by default, but in gray color? Otherwise I see no way to configure this flag.

hbrunn:
good one

StefanRijnhart (approved):

StefanRijnhart:
I think you will want to squash the fixup commit. The autosquash feature has not yet made it into the ocabot https://github.com/OCA/oca-github-bot/pull/100/files

base_delivery_carrier_label/views/delivery.xml
```diff
@@ -52,7 +53,8 @@
     <field name="model">delivery.carrier.option</field>
     <field name="type">tree</field>
     <field name="arch" type="xml">
-      <tree string="delivery_carrier_option">
+      <tree string="delivery_carrier_option" decoration-warning="not active">
```
apollet: "decoration-muted" should be more appropriate for not active flag ?

hbrunn: thanks, that's better indeed

yvaucher (approved):

PR #289, opened 2020-09-10
[ADD] delivery_carrier_label_paazl

hbrunn:
support for carriers supported by [paazl](https://paazl.com)


yvaucher (approved):
LGTM

Quick code review only.

hbrunn:
thanks for the review. this PR will be updated probably this week with the push api support mentioned in the roadmap, we figured this should come with version 1

hbrunn:
@yvaucher @StefanRijnhart I implemented the push part in https://github.com/OCA/delivery-carrier/pull/289/commits/fdc5b1b2a6bc1845b38b128fc4a93e9df7e133a5#diff-4af1e2d72925310a9760a989c8595d07R9 and added a test for this in https://github.com/OCA/delivery-carrier/pull/289/commits/fdc5b1b2a6bc1845b38b128fc4a93e9df7e133a5#diff-48760e728612b656f2e7db5c602757a1R118

You find the documentation for this on https://support.paazl.com/hc/en-us/articles/360021024954-Push-API-reference

delivery_carrier_label_paazl/controllers/main.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def _push_api_get_picking(self, account, data):
+        """Return the shipping that matches the data we got passed"""
+        return (
+            http.request.env["stock.picking"]
+            .sudo()
+            .search(
+                [
+                    ("name", "=", data["orderReference"]),
+                    ("company_id", "=?", account.company_id.id),
```
StefanRijnhart, line 69: This one is going into my notes. Nice one.

StefanRijnhart (approved):

delivery_carrier_label_paazl/models/stock_picking.py
```diff
@@ -0,0 +1,239 @@
...
+        result = {
+            "addresseeLine": partner.name,
+            "street": partner.street,
+            "housenumber": "",
+            "zipcode": partner.zip,
+            "city": partner.city,
+            "province": partner.state_id.code,
+            "country": partner.country_id.code,
+        }
+        # support base_address_extended if installed
```
StefanRijnhart, line 207: I was told Paazl does not accept addresses without a house number at all. Maybe base_address_extended has to be a hard dependency of this module?

hbrunn: in my tests it happily accepted the empty string as house number, so it seems to me if the empty string breaks things, that's carrier specific. That plus the fact that `base_address_extended` has some nasty side effects if not installed with care ([the street format issue](https://github.com/OCA/OpenUpgrade/pull/2408)) makes me prefer not to set a dependency

StefanRijnhart: OK!

StefanRijnhart (approved):

rousseldenis:
@hbrunn Do you keep commits as is or will you squash them ?

hbrunn:
I try to keep commits that touch general things like requirements apart, so if you agree, I'll leave them like that. If not, I'll squash, no problem

rousseldenis:
So, this ok ?

StefanRijnhart:
In this case the commits are tightly linked so I think squashing is preferred.

hbrunn:
now it's just one commit

PR #294, opened 2020-09-23
[12.0][FIX] update delivery with multiple lines fails

sergiocorato:
Add only a check to avoid possible errors not managed.

### DIFF
--- a/delivery_auto_refresh/models/sale_order.py
+++ b/delivery_auto_refresh/models/sale_order.py
@@ -33,6 +33,8 @@ class SaleOrder(models.Model):
         res = super(SaleOrder, self).write(vals)
         for order in self:
             delivery_line = order.order_line.filtered('is_delivery')
+            if len(delivery_line) > 1:
+                continue
             order.with_context(
                 delivery_discount=delivery_line.discount,
             )._auto_refresh_delivery()


pedrobaeza:
But this is not possible in standard AFAIK...

sergiocorato:
> But this is not possible in standard AFAIK...

A delivery line is created for every OUT related to SO (backorder), is it not normal?

pedrobaeza:
Right, but then the thing should be to catch in the context all the discounts.

delivery_auto_refresh/models/sale_order.py
```diff
@@ -33,8 +34,12 @@ def write(self, vals):
         res = super(SaleOrder, self).write(vals)
         for order in self:
             delivery_line = order.order_line.filtered('is_delivery')
+            digits = self.env.user.company_id.currency_id.decimal_places
```
rousseldenis: No, you have to use decimal defined on the discount field: dp.get_precision('Discount')

sergiocorato: Done, thanks!

sergiocorato:
@pedrobaeza @rousseldenis I've modified this PR adding only a clear check, as it's not a common bug probably, so this can be mergeable.

pedrobaeza:
Please check Travis

sergiocorato:
@pedrobaeza done, thanks

pedrobaeza (approved):

marcelsavegnago (approved):

chienandalu (approved):

PR #300, opened 2020-10-13
[12.0] delivery_carrier_label_batch - Remove broken icon

yvaucher:
gtk-execute is not available anymore in 12.0.


yvaucher:
merging as this one is trivial

PR #301, opened 2020-10-13
[12.0] postlogistics sanitize string

yvaucher:
sanitize strings

Some characters are not handled by the REST API.

### DIFF
--- a/delivery_carrier_label_postlogistics/postlogistics/web_service.py
+++ b/delivery_carrier_label_postlogistics/postlogistics/web_service.py
@@ -5,6 +5,7 @@ import base64
 import re
 import logging
 import json
+import unidecode
 import requests
 from PIL import Image
 from io import StringIO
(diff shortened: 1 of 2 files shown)


i-vyshnevska (approved):

delivery_carrier_label_postlogistics/postlogistics/web_service.py
```diff
@@ -17,6 +17,12 @@
 _compile_itemnum = re.compile(r'[^0-9]')
 
 
+def _sanitize_string(string):
+    if not string:
+        return False
```
guewen: is the original string a boolean in this case? I think it makes more sense to return `""`

delivery_carrier_label_postlogistics/postlogistics/web_service.py
```diff
@@ -17,6 +17,12 @@
...
+    return string.replace("\u2018", "'").replace("\u2019", "'")
```
guewen: Do they have a list of characters they support? I guess that's not the only ones they don't support?

yvaucher:
closing in favor of PR #309

yvaucher:
reopening it as #309 doesn't seems to work

delivery_carrier_label_postlogistics/postlogistics/web_service.py
yvaucher: I asked for a list but I got only a vague answer about using utf-8 encoding.

PR #304, opened 2020-10-15
[ADD] delivery_carrier_label_ups

hbrunn:
This adds carrier and label integration for [UPS](ups.com)


StefanRijnhart (approved):
Approving after passing review on the internal PR on our customer project.

delivery_carrier_label_ups/models/product_packaging.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2020 Hunki Enterprises BV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ProductPackaging(models.Model):
+    _inherit = "product.packaging"
```
rousseldenis, line 7: Do you really use product packaging this way ? Why don't you use stock.quant.package ?

rousseldenis: Ok, understood, you use packaging_id on stock.quant https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_quant.py#L354

rousseldenis (approved):
Code review

PR #307, opened 2020-10-21
[13.0] Improve handling of free fee SO lines removal for 13.0

### LINKED ISSUE #306: [13.0] delivery_free_fee_removal behavior not coherent with standard behavior
The module `delivery_free_fee_removal` allows to ignore the creation of delivery lines when they have a zero amount.
I migrated it in 13.0 (#280) but overlooked changes in version 13.0 which bring new problems.

The delivery line becomes more important in 13.0 (probably to simplify the UI/UX):

* the carrier field is not shown, as the delivery line contains the delivery method's name
* the button "add shipping" becomes "update shipping" only when the computed field "delivery_set" is True, which is True when a delivery line exists
* for users, removing the delivery method is done by deleting the delivery line

Summarizing: the delivery line is now the indicator of the delivery method. If we remove the free lines, we have these side-effects:

* we don't know if and which method has been selected
* the button "update shipping" is never shown, only "add shipping"
* no way to remove a delivery method

One way to solve this would be to change the `delivery_set` computed field and show the carrier field: I don't like it because it brings back complexity and different ways for users to work with delivery methods.

IMO a better solution, since 13.0, would be to:

* keep the 0 lines but maybe show them in gray
* ignore the 0 lines when creating the invoice (set `qty_to_invoice` to 0)
* hide delivery lines in sale reports

Any thoughts? 
@cubells @pedrobaeza ?

Also, should I rename the module to `delivery_free_fee_hide` if I make these changes?

guewen:
Description of the issue on #306.
In short: the delivery sale line is required in 13.0 to be able:

* see which delivery method is used
* change the delivery method or update its price
* remove the delivery method

Removing entirely the line when the fee is free is therefore an issue.
The free delivery line is kept, although greyed, in the sale order.
However, it is considered as non-invoiceable, so it doesn't create an
invoice line, and it is hidden from the sale report.

Closes #306


simahawk (approved):
LG

sebalix (approved):

delivery_free_fee_removal/models/sale_order_line.py
```diff
@@ -0,0 +1,24 @@
...
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    is_free_delivery = fields.Boolean(compute="_compute_is_free_delivery", store=True,)
+
+    @api.depends("is_delivery", "currency_id", "price_unit")
+    def _compute_is_free_delivery(self):
+        for line in self:
+            line.is_free_delivery = line.is_delivery and line.currency_id.is_zero(
+                line.price_unit
```
pedrobaeza: Maybe `line.price_reduce`  for taking into account 100% discount?

pedrobaeza (commented):
The rest seems OK

delivery_free_fee_removal/models/sale_order_line.py
guewen: Much better yes! I'll update my PR

guewen: Actually `price_total` should cover all the possible cases.

pedrobaeza:
Can you please squash both commits together and I'll merge? I think they are not significant enough for being kept separate.

guewen:
> Can you please squash both commits together and I'll merge? I think they are not significant enough for being kept separate.

I agree. Done

PR #309, opened 2020-10-23
[FIX] delivery_carrier_label_postlogistic char encoding

gurneyalex:
It seems Postlogistics REST web service is unhappy with the normal JSON encoding
of non ASCII characters and requires these to be encoded in utf-8 -> by using ensure_ascii=False in the call to json.dumps() we try obtain this. 

From the `json` documentation:

    If `ensure_ascii` is false, then the return value can contain non-ASCII
    characters if they appear in strings contained in `obj`. Otherwise, all
    such characters are escaped in JSON strings.

### DIFF
--- a/delivery_carrier_label_postlogistics/postlogistics/web_service.py
+++ b/delivery_carrier_label_postlogistics/postlogistics/web_service.py
@@ -415,7 +415,7 @@ class PostlogisticsWebService(object):
                 'accept': 'application/json',
                 'content-type': 'application/json',
             },
-            data=json.dumps(data, ensure_ascii=False),
+            data=json.dumps(data, ensure_ascii=False).encode("utf-8"),
             timeout=60,
         )
 
--- a/delivery_carrier_label_postlogistics/postlogistics/web_service.py
+++ b/delivery_carrier_label_postlogistics/postlogistics/web_service.py
@@ -4,7 +4,6 @@
 import base64
 import re
 import logging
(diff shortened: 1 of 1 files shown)


gurneyalex:
This is an alternate fix for the issue that #301 tries to tackle.

yvaucher:
Testing it I get this answer from the API:

> Error when communicating with swisspost API: [ {
> "error" : "InvalidJSON"
> } ]

yvaucher (changes requested):
doesn't seems to do the trick, ~~#301 will be the way to go~~

Something is fishy with the right single quotation, I cannot make it work.

With following data:

```
data = {'e-cute': "é", 'Right single quotation': "’"}
```

Fails on right single quotation with "Bad request"
```
>>> json.dumps(data, ensure_ascii=True)
'{"e-acute": "\\u00e9", "Right single quotation": "\\u2019"}'
```

Fails always with "InvalidJSON"
```
>>> json.dumps(data, ensure_ascii=False)
'{"e-acute": "é", "Right single quotation": "’"}'
```

Also gets "InvalidJSON"
```
>>> json.dumps(data, ensure_ascii=False).encode("utf-8")
b'{"e-acute": "\xc3\xa9", "Right single quotation": "\xe2\x80\x99"}'
```

yvaucher:
The API seemingly just had an update that makes this fix obsolete.

PR #311, opened 2020-10-27
[12.0] postlogistics - Fix writing of tracking_number

yvaucher:
When there was a single pack tracking_number was written then emptied with postlogistics labels.
Refactor slightly the writing of reference to make it extendable.

### DIFF
--- a/base_delivery_carrier_label/models/stock_picking.py
+++ b/base_delivery_carrier_label/models/stock_picking.py
@@ -116,6 +116,13 @@ class StockPicking(models.Model):
                 package = self.env['stock.quant.package'].create({})
                 move_lines.write({'result_package_id': package.id})
 
+    def _set_carrier_tracking_ref(self, shipping_labels):
+        if len(shipping_labels) == 1:
+            label = shipping_labels[0]
+            self.write(
+                {"carrier_tracking_ref": label.get("tracking_number")}
+            )
+
     def action_generate_carrier_label(self):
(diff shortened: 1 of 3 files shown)


base_delivery_carrier_label/models/stock_picking.py
```diff
@@ -116,6 +116,13 @@ def _set_a_default_package(self):
                 package = self.env['stock.quant.package'].create({})
                 move_lines.write({'result_package_id': package.id})
 
+    def _set_carrier_tracking_ref(self, shipping_labels):
+        if len(shipping_labels) == 1:
+            label = shipping_labels[0]
+            self.write(
+                {"carrier_tracking_ref": label.get("tracking_number")}
+            )
```
gurneyalex, line 124: the code looks like there could be more than 1 shipping label per picking and that in that case we will not get a carrier_tracking_ref. 

Should the test be changed to `len(shipping_labels) >= 1`?

gurneyalex: @bealdav do you have an opinion on this?

yvaucher: For this PR I propose we keep the current behavior of both modules and we propose an other change if we want to redesign it.

yvaucher:
rebased on top of #325 which has been merged lastly

sebalix (approved):

PR #314, opened 2020-11-04
[12.0] Postlogistics - Recycle access token

yvaucher:
The REST API have limits on the number of requests to access token.
To tackle this and generate hundred of labels the access token is
saved for later uses.

Made threadsafe on the token request to be usable with
deliver_carrier_label_batch module.

### DIFF
--- a/delivery_carrier_label_postlogistics/postlogistics/web_service.py
+++ b/delivery_carrier_label_postlogistics/postlogistics/web_service.py
@@ -6,8 +6,10 @@ import re
 import logging
 import json
 import requests
+import threading
 from PIL import Image
 from io import StringIO
+from datetime import datetime, timedelta
 
 from odoo import _, exceptions
 
@@ -28,6 +30,10 @@ class PostlogisticsWebService(object):
     Allows to generate labels
 
     """
+
+    access_token = False
+    access_token_expiry = False
+    _lock = threading.Lock()
 
     def __init__(self, company):
         self.default_lang = company.partner_id.lang or 'en'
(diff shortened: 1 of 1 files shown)


delivery_carrier_label_postlogistics/postlogistics/web_service.py
```diff
@@ -330,40 +336,52 @@ def _prepare_data(self, lang, frankingLicense, post_customer,
...
+        """Threadsafe access to token"""
+
+        with self._lock:
+            now = datetime.now()
+
+            if self.access_token:
+                # keep a safe margin on the expiration
+                expiry = self.access_token_expiry - timedelta(seconds=5)
+                if now < expiry:
+                    return
```
jcoux: In case of you have a token not expired, I think return no value will not work, because later in the file you take the token as returned value of `get_access_token`:
`access_token = self.get_access_token(picking.env)`
https://github.com/OCA/delivery-carrier/blob/12.0/delivery_carrier_label_postlogistics/postlogistics/web_service.py#L387

jcoux (changes requested):
Are you sure to use the same instantiate class when you print several pickings?
And then the `self.access_token` is still set for each call.

If I correctly understand, the class is instantiate here:
https://github.com/OCA/delivery-carrier/blob/12.0/delivery_carrier_label_postlogistics/models/stock.py#L77
each time you want to generate label, then finally `self.access_token` will be rest each time?

delivery_carrier_label_postlogistics/postlogistics/web_service.py
yvaucher: rewritten to actually write on the class.

It's ok to have each worker creating a new class and having a separate access_token in parallel.

jcoux (approved):

yvaucher:
This has also been validated functionaly

PR #316, opened 2020-11-05
[13.0][PORT] base_delivery_carrier_label (port missing commit into 13.0)

sebastienbeau:
Port #283 #295 #288 ##286 from 12.0 to 13.0.

And fix migration

PR #317, opened 2020-11-05
[14.0] mig base delivery carrier label

sebastienbeau:
Migration of base_delivery_carrier_label
Based on PR https://github.com/OCA/delivery-carrier/pull/316

Please review and merge V13 PR before

We plan to do some refactor before merging

@florian-dacosta


sebastienbeau:
V13 have been merged.

Let's start the refactor and the discussion here (we plan to work on it on 10 november)

florian-dacosta:
@yvaucher @bealdav @sebastienbeau @hbrunn 
I've done some changed based on https://github.com/OCA/delivery-carrier/issues/322
The main goal beeing to be able to generate labels from the same entry point than in native odoo, so, removing `action_generate_carrier_label` method mainly.
You can check this commit for more detail : https://github.com/OCA/delivery-carrier/pull/317/commits/d7437b4e72f01687ebfb041cfcdbd8035a1245ad

florian-dacosta:
@StefanRijnhart FYI since I see you contribute in v12!

base_delivery_carrier_label/models/delivery_carrier.py
```diff
@@ -28,6 +30,16 @@ def send_shipping(self, pickings):
...
+        # We could want to generate labels calling a method that does not depend
+        # on one specific delivery_type.
+        # For instance, if we want to generate a default label in case there are not
+        # carrier
+        # we may want to call another method not based on any delivery_type.
+        # or at the contrary, we could call a common method for multiple delivery_type
+        # for instance, in delivery_roulier, the same method is always called for any
+        # carrier implemented in roulier lib.
+        if result is None:
+            result = self.alternative_send_shipping(pickings)
```
yvaucher, line 42: Does it deprecate https://github.com/OCA/delivery-carrier/tree/12.0/delivery_carrier_label_default ?

base_delivery_carrier_label/models/stock_quant_package.py
```diff
@@ -12,28 +12,25 @@ class StockQuantPackage(models.Model):
...
-    total_weight = fields.Float(
-        digits="Stock Weight",
-        help="Total weight of the package in kg, including the "
-        "weight of the logistic unit.",
-    )
 
-    @api.depends("total_weight")
+    @api.depends("shipping_weight")
     def _compute_weight(self):
         """Use total_weight if defined
```
yvaucher: Doc string to update as it still mention `total_weight`

base_delivery_carrier_label/readme/USAGE.rst
```diff
@@ -1,12 +1,26 @@
...
+In picking UI a button "Send to shipper" trigger label generation
+calling `send_to_shipper()` in models/stock.picking.py
 
 
 ** How to implement my own carrier ? **
 
 
-Override `generate_shipping_labels()` which is called by previous method
-in the same file.
+Define a method `carrier_delivery_type_send_shipping()` which is called by _send_shipping native method.
```
yvaucher: `carrier_delivery_type_send_shipping()`

If I get it right here `carrier_delivery_type` is to be replaced by some other string?
(By the way it's delivery.carrier and not carrier.delivery)

Would it make it clearer if we wrote something like: `{carrier}_send_shipping()`

```
Define a method `{carrier}_send_shipping()` which is called by _send_shipping native method.
Where `{carrier}` is a `delivery.carrier` type.
```

(I think `carrier` is enough in this context)

yvaucher (approved):
Looks good

Just a couple of doc and docstring remarks.

Already approving.

(Reviewed d7437b4)

alexis-via (approved):

base_delivery_carrier_label/models/stock_quant_package.py
florian-dacosta: Fixed

base_delivery_carrier_label/readme/USAGE.rst
florian-dacosta: Fixed
I guess it is clearer indeed, thanks

base_delivery_carrier_label/models/delivery_carrier.py
florian-dacosta: No, it does not deprecate delivery_carrier_label_default module.
Actually, I don't use it but I guess it does not.
I had to add this `alternative_send_shipping` method so it is still possible to achieve what is done in `delivery_carrier_label_default` (and also in `delivery_roulier`).

Because, natively, Odoo will call a method `{carrier}_send_shipping()`, meaning we have to define one method by implemented carrier.
But, in `delivery_carrier_label_default` and in `delivery_roulier`, we want to implement one method that will work for multiple different carrier. In these cases, implementing something like `{carrier}_send_shipping()` is not possible. I added this `alternative_send_shipping` hook, so these modules can override it to implement their logic, that will work for multiple carriers.

PR #319, opened 2020-11-11
[13.0][IMP] delivery_free_fee_removal fix invoicing

TDu:
The free delivery line stop the order from being fully invoiced.
To fix that they are set as invoiced on order confirmation.

### DIFF
--- a/delivery_free_fee_removal/models/sale_order.py
+++ b/delivery_free_fee_removal/models/sale_order.py
@@ -1,0 +1,18 @@
+# Copyright 2020 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def action_confirm(self):
+        super().action_confirm()
+        # Invoice all the free delivery line on order confirmation
+        # Or the order will never be fully invoiced.
+        delivery_lines = self.order_line.filtered(
+            lambda line: line.order_id.state == "sale" and line.is_free_delivery
+        )
+        for line in delivery_lines:
+            line.qty_invoiced = line.product_uom_qty
--- a/delivery_free_fee_removal/models/sale_order_line.py
+++ b/delivery_free_fee_removal/models/sale_order_line.py
@@ -15,10 +15,3 @@ class SaleOrderLine(models.Model):
             line.is_free_delivery = line.is_delivery and line.currency_id.is_zero(
                 line.price_total
             )
-
-    @api.depends("is_free_delivery")
-    def _get_to_invoice_qty(self):
-        free_delivery_lines = self.filtered("is_free_delivery")
-        free_delivery_lines.qty_to_invoice = 0
-        other_lines = self - free_delivery_lines
-        super(SaleOrderLine, other_lines)._get_to_invoice_qty()
--- a/delivery_free_fee_removal/tests/test_delivery_free_fee_removal.py
+++ b/delivery_free_fee_removal/tests/test_delivery_free_fee_removal.py
@@ -48,7 +48,14 @@ class TestDeliveryFreeFeeRemoval(SavepointCase):
         delivery_line = self.sale.mapped("order_line").filtered(lambda x: x.is_delivery)
         self.sale.action_confirm()
         self.assertRecordValues(
-            delivery_line, [{"is_free_delivery": False, "qty_to_invoice": 1}]
+            delivery_line,
+            [
+                {
+                    "is_free_delivery": False,
+                    "qty_to_invoice": 1,
+                    "invoice_status": "to invoice",
+                }
+            ],
         )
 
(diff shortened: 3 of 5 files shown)


delivery_free_fee_removal/models/sale_order.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def action_confirm(self):
+        super().action_confirm()
+        # Invoice all the free delivery line on order confirmation
+        # Or the order will never be fully invoiced.
+        delivery_lines = self.order_line.filtered(
+            lambda line: line.order_id.state == "sale" and line.is_free_delivery
+        )
+        for line in delivery_lines:
+            line.qty_invoiced = line.product_uom_qty
```
guewen, line 18: I think that you should remove the override that makes the "qty_to_invoice" to 0: https://github.com/OCA/delivery-carrier/blob/8574fee184ab2f01b2df349ed8ab7f8b61777d2f/delivery_free_fee_removal/models/sale_order_line.py#L21-L22

delivery_free_fee_removal/tests/test_delivery_free_fee_removal.py
```diff
@@ -58,3 +80,25 @@ def test_delivery_free_fee_removal_free_fee(self):
...
         )
+        self.deliver_and_invoice_order(self.sale)
+
+    def deliver_and_invoice_order(self, order):
+        """Check delivery line state for invoicing and delivery."""
+        # Deliver the order
+        picking = order.picking_ids
+        self.assertEqual(picking.state, "assigned")
+        picking.move_line_ids[0].qty_done = 1.0
+        picking.action_done()
```
guewen: do you actually need to deliver the picking for the test?

delivery_free_fee_removal/tests/test_delivery_free_fee_removal.py
```diff
@@ -58,3 +80,25 @@ def test_delivery_free_fee_removal_free_fee(self):
...
+        wizard = self.wizard_obj.with_context(
+            active_ids=order.ids, active_model=order._name
+        ).create({})
+        wizard.create_invoices()
+        self.assertTrue(order.invoice_ids)
+        # Check delivery lines are properly invoiced and delivered
+        fee_line = order.order_line.filtered("is_delivery")
+        self.assertEqual(fee_line.invoice_status, "invoiced")
+        # Check the order is fully invoiced
+        self.assertEqual(self.sale.invoice_status, "invoiced")
```
guewen: My reading of the code tells me that the fee line is supposed to have `invoice_status` equal to `invoiced` even before the invoice is created, as soon as we confirm the SO. If so, only testing this status in the `assertRecordValues` of the tests would have been enough (without creating the inventory, invoice, ...):

```
        self.assertRecordValues(
            delivery_line, [{"is_free_delivery": False, "qty_to_invoice": 1, "invoiced_qty": 1, "invoice_status": "invoiced"}]
        )
```

(The `invoice_status` of the sale being equal to "invoiced" when all the lines are "invoiced" is a property of the "sale" module, which should be tested (and maybe is) there.)

delivery_free_fee_removal/tests/test_delivery_free_fee_removal.py
TDu: I guess not, I wrote the test to understand what was required. They do look over the top.

delivery_free_fee_removal/models/sale_order.py
TDu: Removed

delivery_free_fee_removal/tests/test_delivery_free_fee_removal.py
TDu: Removing them

delivery_free_fee_removal/tests/test_delivery_free_fee_removal.py
TDu: Did that, thanks for your review

delivery_free_fee_removal/models/sale_order_line.py
```diff
@@ -19,6 +19,6 @@ def _compute_is_free_delivery(self):
     @api.depends("is_free_delivery")
     def _get_to_invoice_qty(self):
         free_delivery_lines = self.filtered("is_free_delivery")
-        free_delivery_lines.qty_to_invoice = 0
+        # free_delivery_lines.qty_to_invoice = 0
         other_lines = self - free_delivery_lines
         super(SaleOrderLine, other_lines)._get_to_invoice_qty()
```
guewen: you can actually remove the whole override of `_get_to_invoice_qty`: the only thing it was doing was to set `qty_to_invoice` to 0

jgrandguillaume (approved):
LGTM thx !

guewen (approved):

guewen:
Can you squash please?

TDu:
Done

PR #183, opened 2019-04-03
[10.0] delivery_carrier_label_postlogistics: inverse addresses on incoming shipments

grindtildeath:
Includes:
 - [x] #181

PR #320, opened 2020-11-12
[12.0] postlogistics - Fix return addresses on labels

yvaucher:
When pre-printing addresses for customers to return a product
the addresses were inverted. Goal is to have the current company
as the recipient.

Includes #183 

(https://github.com/OCA/delivery-carrier/pull/301 is pre-merged in there as currently being merged by ocabot)

### DIFF
--- a/delivery_carrier_label_postlogistics/models/stock.py
+++ b/delivery_carrier_label_postlogistics/models/stock.py
@@ -24,6 +24,24 @@ class StockPicking(models.Model):
         "Mobile", help="For notify delivery by telephone (ZAW3213)"
     )
 
+    def _compute_show_label_button(self):
+        """Determine if we should show the button to generate labels
+
+        Allows to print the label early, as for return pickings when
+        the label must be printed and sent to the customer.
+
+        """
+        visible_states = ['confirmed', 'assigned', 'done']
+        postlog_pickings = self.filetered(
+            lambda rec: rec.carrier_id.delivery_type == 'postlogistics'
+        )
+        for pick in postlog_pickings:
+            pick.show_label_button = (
+                pick.state in visible_states and pick.carrier_code
+            )
+        others = self - postlog_pickings
+        return super(StockPicking, others)._compute_show_label_button()
+
     @api.multi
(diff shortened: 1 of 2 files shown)


jcoux (approved):
Only review last commit, technically seems good :+1:

yvaucher:
Rewrote last commit message which was too generic on the current PR and not commit specific.

jcoux:
@yvaucher Created a PR on your branch with a fixup:
https://github.com/yvaucher/delivery-carrier/pull/3

delivery_carrier_label_postlogistics/models/stock.py
```diff
@@ -24,6 +24,24 @@ class StockPicking(models.Model):
...
+        the label must be printed and sent to the customer.
+
+        """
+        visible_states = ['confirmed', 'assigned', 'done']
+        postlog_pickings = self.filetered(
+            lambda rec: rec.carrier_id.delivery_type == 'postlogistics'
+        )
+        for pick in postlog_pickings:
+            pick.show_label_button = (
+                pick.state in visible_states and pick.carrier_code
```
grindtildeath, line 40: I'm just not sure about this, shouldn't the button be displayed only?
 - When state = done on outgoing pickings
 - When state = confirmed or assigned on incoming pickings

EDIT: This is how it was in #183

grindtildeath (commented):
Question

jcoux:
Additional fix: https://github.com/yvaucher/delivery-carrier/pull/4

delivery_carrier_label_postlogistics/models/stock.py
```diff
@@ -24,6 +24,24 @@ class StockPicking(models.Model):
...
 
+    def _compute_show_label_button(self):
+        """Determine if we should show the button to generate labels
+
+        Allows to print the label early, as for return pickings when
+        the label must be printed and sent to the customer.
+
+        """
+        visible_states = ['confirmed', 'assigned', 'done']
+        postlog_pickings = self.filetered(
```
yvaucher, line 35: Typo there

PR #324, opened 2020-12-15
[12.0][FIX] restore domain behaviour from v9

i-vyshnevska:


guewen:
It seems like there is a test to adapt

leemannd (approved):

PR #331, opened 2021-01-31
[IMP] better multihrending processing

i-vyshnevska:


ajaniszewska-dev (approved):
LG

yvaucher:
@i-vyshnevska please insert in the description / commit message in what this way of doing multi threading is better.

delivery_carrier_label_batch/wizard/generate_labels.py
```diff
@@ -150,14 +150,14 @@ def _get_all_files(self, batch):
...
-            t.start()
+        with concurrent.futures.ThreadPoolExecutor() as executor:
+            res = [
+                executor.submit(self._worker, args=(data_queue, error_queue))
+                for i in range(num_workers)
+            ]
 
         # wait for all tasks to be done
-        data_queue.join()
+        concurrent.futures.as_completed(res)
```
StefanRijnhart, line 160: I have no experience with this particular library, but if I read https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.as_completed correctly, `as_completed` does not wait for the tasks to be done, but returns an iterator that you have to go through to process results as they finish. Am I missing something? I see that https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.wait may be a more straight forward alternative.

PR #340, opened 2021-03-03
WIP [13.0][MIG] delivery_auto_refresh

acsonefho:
Migration of module `delivery_auto_refresh` from `12.0` to `13.0`.

During this migration, as the original code (about auto_refresh) doesn't work anymore, I update this part to use and simulate the standard wizard to automatically update the shipping cost (based on the carrier) but for now it doesn't work too.


pedrobaeza:
This was already migrated in other PR and merged, sorry.

PR #344, opened 2021-03-24
[12.0] delivery_carrier_label_postlogistics: Remove _sanitize_string

ajaniszewska-dev:
The Postlogistics API behaviour changed, it is now possible to send it 

With default encoding (`unicode-escape`) with `requests` it still fail but only on the first label queried after getting the token.

Use case 1:
ask for token
generate label without accents :heavy_check_mark: 
generate label with accents :heavy_check_mark: 

Use case 2:
ask for token
generate label with accents :negative_squared_cross_mark: 
generate label without accents :heavy_check_mark: 

But it will work when providing json content to `requests` with `json.dumps(data, ensure_ascii=False).encode("utf-8")`

EDIT: yvaucher: provide a description

### DIFF
--- a/delivery_carrier_label_postlogistics/postlogistics/web_service.py
+++ b/delivery_carrier_label_postlogistics/postlogistics/web_service.py
@@ -21,13 +21,6 @@ _compile_itemid = re.compile(r'[^0-9A-Za-z+\-_]')
 _compile_itemnum = re.compile(r'[^0-9]')
 
 
-def _sanitize_string(string):
-    """Temporary fix, remove all non ascii char as the API
-    weirdly doesn't support them in some cases.
-    """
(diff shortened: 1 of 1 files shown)


i-vyshnevska (commented):
check travis

ajaniszewska-dev:
> check travis

@i-vyshnevska  travis fails due to other modules.

yvaucher (changes requested):
This is missing a change on

https://github.com/OCA/delivery-carrier/blob/3c89a8418b71612a527cbff592d7ff9bc16a0503/delivery_carrier_label_postlogistics/postlogistics/web_service.py#L441

Without the proper encoding the API still fails.

Plus, can you put a meaningful commit message? (I provided this PR description)

PR #345, opened 2021-03-25
[12.0][RFR] base_delivery_carrier_label: show options as tags

StefanRijnhart:
Show carrier options as tags.

Before:
image

After:
image

### DIFF
--- a/base_delivery_carrier_label/models/delivery_carrier_option.py
+++ b/base_delivery_carrier_label/models/delivery_carrier_option.py
@@ -2,7 +2,7 @@
 # Copyright 2013-2016 Camptocamp SA
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class DeliveryCarrierOption(models.Model):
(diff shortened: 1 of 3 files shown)


StefanRijnhart:
I would very much like to remove the carrier_code field from the picking view as well. @bealdav , @yvaucher what do you think?

florian-dacosta (approved):

bealdav (approved):
Thanks a lot Stefan

StefanRijnhart:
@bealdav thanks for the review. Do you agree with removing the "Code" field here as well (column `carrier_code`). I think it adds very little here.

yvaucher (approved):
On the principle it make the view a bit lighter so I'm in favor, our UI is not really good with those options.

I just have concerns on loosing the mandatory field visibility which explains why an option was selected for the picking and can't be removed.
Would have been great if we could had the possibility to still show it somehow.

But maybe it's only useful for debugging thus useless on the UI and not a real issue and I didn't played with it since a while.

StefanRijnhart:
@yvaucher good point! I believe we can add a color field for that. I'll have a try tomorrow.

hbrunn:
@StefanRijnhart are you still planning to work on the colors or should I merge?

StefanRijnhart:
Still on my list!

yvaucher:
My comment is not blocking, if you want we can merge it as is, and add the colors later.

hbrunn:
I'm not in a hurry with merging, just came across this in my notification list. Will revisit when it shows up again

StefanRijnhart:
Sorry for the delay! Mandatory options show up in orange now. Screenshot updated.

hbrunn (approved):

PR #354, opened 2021-04-12
[13.0][ADD] New module 'stock_picking_delivery_link'

sebalix:
Moved from https://github.com/OCA/stock-logistics-workflow/pull/825

Adds link to the delivery on all intermediate operations (pick and pack with a pick/pack/ship setup).

It aims to replace https://github.com/OCA/stock-logistics-warehouse/pull/1115 to get a useful module on its own instead of a "helper" one.

image

Ref. 2356


stock_picking_delivery_link/models/stock_picking.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2021 Camptocamp SA (http://www.camptocamp.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+def get_first_move_dest(moves, done=False):
+    move_states = ("cancel", "done")
+    for move in moves.move_dest_ids:
+        if move.state in move_states if done else move.state not in move_states:
```
mmequignon: as discussed in "private", you could use `if not operator.xor(done, move.state in move_states)` instead of the double if.

stock_picking_delivery_link/models/stock_picking.py
```diff
@@ -0,0 +1,43 @@
...
+            picking.ship_carrier_id = ship.carrier_id
+
+    def _get_ship_from_chain(self, done=False):
+        """Returns the shipment related to the current operation."""
+        move_dest = get_first_move_dest(self.move_lines, done=done)
+        while move_dest:
+            picking = move_dest.picking_id
+            if picking.picking_type_id.code == "outgoing":
+                return picking
+            move_dest = get_first_move_dest(move_dest, done=done)
```
mmequignon, line 43:
```suggestion
        for move_dest in get_sorted_move_dest(self.move_lines, done=done):
            picking = move_dest.picking_id
            if picking.picking_type_id.code == "outgoing":
                return picking
```
could look like this if the function was yielding instead of returning the first move.

mmequignon (approved):
Costmetic questions / suggestions.
LG

stock_picking_delivery_link/models/stock_picking.py
simahawk: can't you simply change moves_states?
```py
move_states = ("cancel", "done")
if not done:
    move_states = (a, b, c)
if move.state in move_states:
```
sounds more clear and explicit

sebalix: As seen together, better to go with an "if box", much more readable.

sebalix: @simahawk I didn't want to list all possible states of a move, I think the code is OK now.

stock_picking_delivery_link/models/stock_picking.py
sebalix: We could transform `get_first_move_dest` into a generator yes, but it's OK like this.

jgrandguillaume (approved):
functional test ok !

PR #355, opened 2021-04-12
[13.0][ADD] Send delivery notification from any operation

sebalix:
Moved from https://github.com/OCA/stock-logistics-workflow/pull/821

<s>Based on #354 (first commit here)</s> merged

New module `delivery_send_to_shipper_at_operation` allowing to send the delivery notification from any operation (among chained operations) before the delivery.
With a pick/pack/ship setup this allows to send the delivery notification to the carrier when the pack operation is marked as done for instance (by default Odoo sends it on the ship/delivery).

Once the notification has been sent from a non-delivery operation, it won't be sent again when the delivery operation is validated.

**Screenshots:**

delivery_carrier1
delivery_carrier2

**Drawback:**

This way of sending delivery notifications is not compatible if we have several pack operations to process before the delivery (backorders?). In such a case, the notification will be generated from only part of the goods to deliver, which is not expected.

Ref. 2286


delivery_send_to_shipper_at_operation/models/stock_picking.py
```diff
@@ -0,0 +1,90 @@
...
+        # Override to hide the "Send to shipper" button if the delivery
+        # notification has already been sent
+        result = super().fields_view_get(
+            view_id=view_id, view_type=view_type, toolbar=toolbar, submenu=submenu
+        )
+        if result.get("name") == "stock.picking.form":
+            result["arch"] = self._fields_view_get_adapt_send_to_shipper_attrs(
+                result["arch"]
+            )
+        return result
```
mmequignon, line 87: Given how tricky this part is, it would be nice to make dovecot happy about this.

delivery_send_to_shipper_at_operation/models/stock_picking.py
```diff
@@ -0,0 +1,90 @@
...
+    _inherit = "stock.picking"
+
+    delivery_notification_sent = fields.Boolean(default=False)
+
+    def _send_confirmation_email(self):
+        for picking in self:
+            picking._handle_send_to_shipper_at_operation()
+            super(StockPicking, picking)._send_confirmation_email()
+
+    def _handle_send_to_shipper_at_operation(self):
```
simahawk, line 27: have you checked my comment regarding hooking to here in the old PR?

sebalix: Done, last fixup fixes this, now we add the delivery cost to the SO only on the ship validation like before.
This module will only send the delivery notice to the carrier, nothing more.

jgrandguillaume (approved):
Functional test ok

delivery_send_to_shipper_at_operation/tests/test_send_to_shipper.py
```diff
@@ -0,0 +1,146 @@
...
+                "product_id": cls.delivery_fee.id,
+                "integration_level": "rate_and_ship",
+                "send_delivery_notice_on": "custom",
+                "send_delivery_notice_picking_type_ids": [
+                    (6, 0, cls.wh.pack_type_id.ids)
+                ],
+                "invoice_policy": "real",
+            }
+        )
+        cls.carrier_on_pack.__class__.send_shipping = types.MethodType(
```
simahawk: why not mock or self.patch? this won't rollback you hack :thinking:

simahawk (commented):
LG just one remark on test mocking

delivery_send_to_shipper_at_operation/models/stock_picking.py
simahawk: seems a good point :)
Also, why not using the OCA module to extend existing domains?

delivery_send_to_shipper_at_operation/tests/test_send_to_shipper.py
sebalix: On another topic I tested mock+side_effect to test the call while keeping the underlying method executed, I catched one drawback by doing so: we lose the user's context (and I was calling the mocked method with a specific context key). By monkey-patching with `types` there was no issue (context was updated as expected), so I kept this idea here, even if I don't rely on the context in this case.

I could just set a `return_value` with `mock` here for sure.

delivery_send_to_shipper_at_operation/models/stock_picking.py
sebalix: @simahawk A matter of preference, I'm OK to use the OCA module for custom modules, but don't like to add it as dependencies in other OCA modules, when a module touches internal things I'm always afraid to bring regressions while here it's restricted to the delivery carrier form view.

Anyway, tests will be added, unless you prefer to switch on the view extension OCA module.

delivery_send_to_shipper_at_operation/tests/test_send_to_shipper.py
sebalix: I switched to `mock` here

sebalix:
Rebased on top of 13.0

simahawk (approved):

delivery_send_to_shipper_at_operation/models/stock_picking.py
```diff
@@ -0,0 +1,109 @@
...
+            related_ship = self.ship_picking_id
+            carrier = related_ship.carrier_id
+            if (
+                carrier.integration_level == "rate_and_ship"
+                and carrier.send_delivery_notice_on == "custom"
+                and self.picking_type_id
+                in carrier.send_delivery_notice_picking_type_ids
+            ):
+                self.carrier_id = carrier
+                self.with_context(skip_delivery_cost=True).send_to_shipper()
```
mmequignon, line 50: Why don't you call this when `self.carrier_id` is already defined?

PR #342, opened 2021-03-11
[13.0] Add delivery_postlogistics_dangerous_goods

mmequignon:
This modules overrides `Postlogistics Shipping` to declare dangerous goods
during postlogistics label generation.

Based on https://github.com/OCA/delivery-carrier/pull/357

First commit is shared with https://github.com/OCA/delivery-carrier/pull/358.
Those will need to be rebased when one will be merged.

PR #358, opened 2021-04-15
[13.0][FIX] Do not send disallowed chars on postlogistics api

mmequignon:
Some chars are disallowed from postlogistics API ("|", "\", "<", ">") and output json should be sanitized

Based on https://github.com/OCA/delivery-carrier/pull/367.
Those will need to be rebased when one will be merged.

### DIFF
--- a/delivery_postlogistics/postlogistics/web_service.py
+++ b/delivery_postlogistics/postlogistics/web_service.py
@@ -22,6 +22,8 @@ _compile_itemnum = re.compile(r"[^0-9]")
 AUTH_PATH = "/WEDECOAuth/token"
 GENERATE_LABEL_PATH = "/api/barcode/v1/generateAddressLabel"
 
+DISALLOWED_CHARS = ["|", "\\", "<", ">"]
+
 
 class PostlogisticsWebService(object):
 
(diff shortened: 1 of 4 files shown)


delivery_postlogistics/tests/test_sanitize_values.py
```diff
@@ -0,0 +1,135 @@
...
+                "delivery_type": "postlogistics",
+                "product_id": shipping_product.id,
+                "postlogistics_endpoint_url": ENDPOINT_URL,
+                "postlogistics_client_id": CLIENT_ID,
+                "postlogistics_client_secret": CLIENT_SECRET,
+                "postlogistics_label_layout": label_layout.id,
+                "postlogistics_output_format": output_format.id,
+                "postlogistics_resolution": image_resolution.id,
+            }
+        )
```
sebalix: All this code is the same than the one in `test_postlogistics.py` (as well as `ENDPOINT_URL` etc), it deserves a common class both tests should inherit from.

sebalix (approved):

sebalix:
Depends on https://github.com/OCA/delivery-carrier/pull/342 to get the refactoring of tests, waiting for it to be merged.

simahawk (approved):

delivery_postlogistics/postlogistics/web_service.py
```diff
@@ -434,6 +445,13 @@ def get_access_token(cls, picking_carrier):
             cls.access_token_expiry = now + timedelta(seconds=response["expires_in"])
             return cls.access_token
 
+    def _sanitize_string(self, value):
+        """Removes disallowed chars ("|", "\", "<", ">") from strings."""
+        if isinstance(value, str):
```
gurneyalex, line 453: if we have False in there, do we need to return False or an empty string?

mmequignon: The idea is to remove  disallowed chars from incoming strings, I don't think we should alter the value if it's something else than a string. That case should already be handled by the base implementation.

mmequignon: I mean, sending `{"whatever": null}` and `{"whatever": ""}` is not the same, and I don't want to modify something that is currently working.

simahawk:
@mmequignon pls rebase and drop the 1st commit

PR #360, opened 2021-04-15
[14.0][MIG] delivery_package_number: Migration to 14.0

hildickethan:
Standard migration to 14.0


delivery_package_number/models/stock_picking.py
```diff
@@ -0,0 +1,31 @@
...
+    _inherit = "stock.move.line"
+
+    @api.onchange("result_package_id")
+    def onchange_package_ids(self):
+        """TODO: in v13 Change field to compute readonly=False"""
+        picking = self.picking_id
+        picking.number_of_packages = len(picking.package_ids) or 1
+
+    def write(self, vals):
+        """TODO: in v13 Change field to compute readonly=False"""
```
gurneyalex: can we take the opportunity of this migration to handle this TODO?

hildickethan: I assume it means turning the field into a stored compute with readonly=False so its editable?

pedrobaeza: Yes, that's it

hildickethan: Do I need to preserve this funcionality? `Now we put in pack some quantities and the number of packages is set as the number of packs is added to the number already set`
It's a bit convoluted to do in a compute

pedrobaeza: OK, we can keep the TODO for me.

Jortolsa-S73 (approved):
Functional review

florian-dacosta:
Hello, I don't really get this module.
I did not understand why the "TODO: in v13 Change field to compute readonly=False" should not be done?

Indeed, from my tests, even a very simple case does not work, maybe I am missing something?
For instance, you have a delivery order with one line/one product ready to go.
By default package number is 1. If you add a package (with button "Put in pack"), then number of package is 2.

So a picking with 1 package 1 product has a number of package equal to 2.

Also, if you add some packages, made a mistake and then delete the `stock.move.line` and re do some stuff, the number of package will be wrong.

I understand the need to put a number of package manually, I guess when we do not manage package at all in Odoo, or maybe once the delivery is done, if we need to adjust the field of any reason. But except for this, I believe the field should be computed depending on the stock.move.lines in order to be reliable.

Is it not possible to have a field depending on `move_line_ids.result_package_id` which make a just something like `picking.number_of_package = len(picking.package_ids)`
Making the field readonly=False to be able to edit it?

Any explanation is welcome!

pedrobaeza:
Starting with, the commit history is not preserved following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0.

So we can't know what are the changes done from the previous version to this one. Anyways, @florian-dacosta seems correct.

hildickethan:
> Starting with, the commit history is not preserved following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0.

Sorry, it seems a colleague of mine force-pushed an update on a grafted branch and it nuked the history, I have re-built it.

> Is it not possible to have a field depending on `move_line_ids.result_package_id` which make a just something like `picking.number_of_package = len(picking.package_ids)`

I believe this is what I did when the TODO was suggested and it worked fine functionally, but one of the test cases failed and I asked if it was ok to drop that.

From what I remember the case that was lost was the possibility of manually adding in a number with a case of:

- Manually setting 3 packages
- Adding 1 via action_put_in_pack()
- It updates to 4

With the simple compute it updated to 1 as it was the amount of Odoo packages and failed the test case.

Personally I don't care if this case stays as it seems like patching bad habits, but I didn't want to just remove functionality someone somewhere might be using when I didn't need to. If everyone agrees that it's ok to remove it then I will change it to a compute of `len(picking.package_ids) or 1` as there needs to be 1 package minimum when doing deliveries with webservices (and it's also a current feature)

florian-dacosta:
Well, in the case I describe, it is a very normal case (add one package from native put in pack button) and it does not work.

The test from the case you describe seems really strange to me. 
Manually setting a number of pack and then add a real pack...I'd say it is kind of incompatible.

So, IMO, we should go to the computed mode (still keeping e readonly=False to be able to update it manually, when it won't change anymore) and drop the test/case you describe.

Or find any other solution, but it does not seems acceptable that a basic normal case is wrong !

If the case you described is really important (I am curious about the customer case), then maybe it could be done with an additional field.
manual_pack_number = classic integer invisible field, default to 0
pack_number = len(package_ids) + manual_pack_number or 0

pack number could be readonly=False with a inverse method which would get the number of manual pack and set it. manual_pack_number = (pack_number - len(package_ids))

Since this case don't seem very common, it could even be implemented in a submodule, only if really used?

hildickethan:
I agree that the compute would be better. The use case I described is not mine and I don't see it to be an important one either, at worst you have to input a manual number again.

Good point that if someone is using such an uncommon case, it can be handled in another module, so I think I will swap this over to compute and remove that test case

hildickethan:
@florian-dacosta @pedrobaeza PR updated with TODOs removed and using a computed version

pedrobaeza (approved):

pedrobaeza:
Not sure why it's failing. Can you rebase to see if we get the failing here in the PR?

hildickethan:
it looks like its API tests in delivery_roulier_laposte_fr

hildickethan:
I rebased and it worked, so it was probably just an API issue last night

pedrobaeza:
Let's try again:

/ocabot merge nobump

hildickethan:
:thinking: I don't even see an actual error, just this message `changes detected (content changed, file is created, or file is deleted):\n/home/travis/.cache/pip/http/0/0/e/1/b/00e1b74348bc37914bf3f2986bf03408e30a581c91b36ee93aab7c01`

dreispt:
It seems that `delivery_roulier_laposte_fr` s blocking the merge
https://app.travis-ci.com/github/OCA/delivery-carrier/jobs/530733142#L1549

@florian-dacosta can you please check?

dreispt:
Update: https://github.com/OCA/delivery-carrier/pull/404 might have solved the issue - retrying the merge.

hildickethan:
Thanks

PR #361, opened 2021-04-16
[13.0][FIX] delivery_package_fee: flag order lines

sebalix:
Flag package fee order lines with a new boolean `is_delivery_package_fee`.
Standard method `_is_delivery` will also consider this field to exclude
the lines from the computation of the `expected_date` in `sale` module.

Ref. 2438

### DIFF
--- a/delivery_package_fee/models/sale_order_line.py
+++ b/delivery_package_fee/models/sale_order_line.py
@@ -6,4 +6,9 @@ from odoo import fields, models
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
-    package_fee_id = fields.Many2one(comodel_name="delivery.package.fee", default=False)
+    package_fee_id = fields.Many2one(
+        comodel_name="delivery.package.fee", default=False, ondelete="restrict"
+    )
+
+    def _is_delivery(self):
+        return super()._is_delivery() or bool(self.package_fee_id)
--- a/delivery_package_fee/models/delivery_carrier.py
+++ b/delivery_package_fee/models/delivery_carrier.py
@@ -7,5 +7,7 @@ class DeliveryCarrier(models.Model):
     _inherit = "delivery.carrier"
 
     package_fee_ids = fields.One2many(
-        comodel_name="delivery.package.fee", inverse_name="carrier_id",
+        comodel_name="delivery.package.fee",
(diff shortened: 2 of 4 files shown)


mmequignon (approved):

delivery_package_fee/models/sale_order_line.py
```diff
@@ -7,3 +7,9 @@ class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
     package_fee_id = fields.Many2one(comodel_name="delivery.package.fee", default=False)
+    is_delivery_package_fee = fields.Boolean(
+        string="Is a Delivery Package Fee", default=False,
```
gurneyalex: could we not compute this from the presence / absence of package_fee_id, and not resort on a field that has to be added in the database?

delivery_package_fee/models/sale_order_line.py
```diff
@@ -7,3 +7,9 @@ class SaleOrderLine(models.Model):
...
+    )
+
+    def _is_delivery(self):
+        return super()._is_delivery() or self.is_delivery_package_fee
```
gurneyalex: I don't know the context really but would this be equivalent?

```suggestion
        return super()._is_delivery() or bool(self.package_fee_id)
```

gurneyalex (commented):
question asked about the pertinence of the new field

delivery_package_fee/models/sale_order_line.py
sebalix: I thought about that but, if we update the package fees defined on the carrier (removing them for instance) we will get an empty `package_fee_id` field on current order lines. I wanted to avoid that but if you think it's overkill I can still update this.

delivery_package_fee/models/sale_order_line.py
sebalix: It is indeed (if the package fee configuration is not removed meanwhile)

delivery_package_fee/models/sale_order_line.py
gurneyalex: set the field ondelete = restrict and make sure we have an active field on that model ?

sebalix: done

simahawk (approved):
@sebalix squash pls

PR #362, opened 2021-04-19
[13.0] [MIG] delivery_auto_refresh: Migration to v13

CarlosRoca13:
cc @Tecnativa TT26914

please @pedrobaeza @joao-p-marques review this


delivery_auto_refresh/models/__init__.py
```diff
@@ -1,2 +1,3 @@
 from . import sale_order
+from . import sale_order_line
```
chienandalu: The file doesn't exists

pedrobaeza: And you shouldn't split in 2.

CarlosRoca13:
@chienandalu @pedrobaeza Ups sorry 😅 changes done

delivery_auto_refresh/models/sale_order.py
```diff
@@ -1,21 +1,49 @@
...
+    carrier_id = fields.Many2one(
+        compute="_compute_carrier_id", store=True, readonly=False
+    )
+    invoice_shipping_on_delivery = fields.Boolean(
+        string="Invoice Shipping on Delivery", copy=False
+    )
+
+    @api.depends("partner_id")
+    def _compute_carrier_id(self):
+        for rec in self:
```
pedrobaeza, line 20: As this module is not the one defining the field, you can add this compatibility layer:
```python
if hasattr(super(), "_compute_carrier_id"):
    super()._compute_carrier_id()
```

delivery_auto_refresh/models/sale_order.py
```diff
@@ -1,21 +1,49 @@
...
+                    rec.carrier_id = rec.partner_id.property_delivery_carrier_id
+                else:
+                    carrier_ids = self.env["delivery.carrier"].search(
+                        [
+                            "|",
+                            ("company_id", "=", False),
+                            ("company_id", "=", self.company_id.id),
+                        ]
+                    )
+                    rec.carrier_id = min(carrier_ids, key=lambda c: c.sequence)
```
pedrobaeza: The order is already on the model, so you don't need to reorder again. Just put `limit=1` in the search sentence. Anyway, if the order wouldn't be set, you can specify also in the search `order="field"` for auto-ordering on result (and also add `limit`).

pedrobaeza: Test the onchange by the way.

delivery_auto_refresh/models/sale_order.py
```diff
@@ -1,21 +1,49 @@
...
 from odoo.tools import safe_eval
 
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    carrier_id = fields.Many2one(
+        compute="_compute_carrier_id", store=True, readonly=False
+    )
+    invoice_shipping_on_delivery = fields.Boolean(
```
pedrobaeza: I'm not sure we should recover this behavior, but be based on the carrier field that specify the invoicing policy. What do you think?

delivery_auto_refresh/models/sale_order.py
```diff
@@ -29,14 +57,15 @@ def create(self, vals):
...
-            if len(delivery_line) > 1:
-                continue
-            order.with_context(
-                delivery_discount=delivery_line.discount,
-            )._auto_refresh_delivery()
+        res = super().write(vals)
+        if self._get_param_value():
+            for order in self:
+                delivery_line = order.order_line.filtered("is_delivery")
+                if len(delivery_line) > 1:
```
pedrobaeza: Why excluding this?

delivery_auto_refresh/models/sale_order.py
```diff
@@ -47,3 +76,35 @@ def _create_delivery_line(self, carrier, price_unit):
...
+    _inherit = "sale.order.line"
+
+    def unlink(self):
+        """ Trick to avoid infinite recursive calls to write method on sale.order
+        """
+        for line in self:
+            get_param = self.env["ir.config_parameter"].sudo().get_param
+            param = "delivery_auto_refresh.auto_add_delivery_line"
+            if line.is_delivery and safe_eval(get_param(param, "0")):
+                line.is_delivery = False
```
pedrobaeza: I think upstream should have another think for avoiding this (maybe a context key?)

delivery_auto_refresh/tests/test_delivery_auto_refresh.py
```diff
@@ -75,20 +75,21 @@ def setUp(self):
                 "partner_id": self.partner.id,
                 "partner_shipping_id": self.partner.id,
                 "order_line": [
-                    (0, 0, {"product_id": self.product.id, "product_uom_qty": 2,})
+                    (0, 0, {"product_id": self.product.id, "product_uom_qty": 2})
                 ],
             }
         )
         order = order_obj.new(order_vals)
+        order.onchange_partner_id()
```
pedrobaeza: Use `Form` for better resilience

delivery_auto_refresh/tests/test_delivery_auto_refresh.py
```diff
@@ -106,9 +107,18 @@ def test_auto_refresh_so(self):
         line_delivery = self.order.order_line.filtered("is_delivery")
         self.assertEqual(line_delivery.price_unit, 95)
 
-    def test_auto_refresh_picking(self):
+    def _test_auto_refresh_picking(self):
```
pedrobaeza: This test is not executed

pedrobaeza (changes requested):
I miss more coverage for certain parts.

delivery_auto_refresh/models/sale_order.py
```diff
@@ -1,21 +1,49 @@
...
+                    carrier_ids = self.env["delivery.carrier"].search(
+                        [
+                            "|",
+                            ("company_id", "=", False),
+                            ("company_id", "=", self.company_id.id),
+                        ]
+                    )
+                    rec.carrier_id = min(carrier_ids, key=lambda c: c.sequence)
+
+    def _get_param_value(self):
```
joao-p-marques: Maybe you could give a more significant name to this function? With a generic name like this, we have the risk of conflicting with another method in the future / another module.

Also, change the calls accordingly.

delivery_auto_refresh/models/sale_order.py
```diff
@@ -47,3 +76,35 @@ def _create_delivery_line(self, carrier, price_unit):
...
+        return res
+
+    def action_open_delivery_wizard(self):
+        if self._get_param_value():
+            raise UserError(
+                _(
+                    "The shipping cost is updated automatically by changing "
+                    "the Delivery Method field."
+                )
+            )
```
joao-p-marques: If this will launch this error in UI, why not simply hiding the button in the interface if the parameter is set?

delivery_auto_refresh/models/sale_order.py
```diff
@@ -47,3 +76,35 @@ def _create_delivery_line(self, carrier, price_unit):
...
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def unlink(self):
+        """ Trick to avoid infinite recursive calls to write method on sale.order
+        """
+        for line in self:
+            get_param = self.env["ir.config_parameter"].sudo().get_param
+            param = "delivery_auto_refresh.auto_add_delivery_line"
+            if line.is_delivery and safe_eval(get_param(param, "0")):
```
joao-p-marques: Can't you call the method from the associated object instead of rewriting the code?
```python
self.order_id._get_param()  # or whatever the name if changed
```

delivery_auto_refresh/models/sale_order.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def action_open_delivery_wizard(self):
+        if self._get_param_value():
+            raise UserError(
+                _(
+                    "The shipping cost is updated automatically by changing "
+                    "the Delivery Method field."
+                )
+            )
+        super().action_open_delivery_wizard()
```
joao-p-marques: I don't know if this is normal but, while on draft and with the parameter set to False, if I press this button nothing happens in the UI. Is it the intended behavior?

joao-p-marques (changes requested):
Code and functional review

delivery_auto_refresh/models/sale_order.py
CarlosRoca13: Ok I will use it

CarlosRoca13:
@pedrobaeza and @joao-p-marques changes done! Check it again please

delivery_auto_refresh/models/sale_order.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def _auto_refresh_delivery(self):
+        self.ensure_one()
+        # Make sure that if you have removed the carrier, the line is gone
+        if self.state in {"draft", "sent"}:
+            # Context added to avoid the recursive calls and save the new
+            # value of carrier_id
+            ctx = self.env.context.copy()
+            ctx["new_carrier_id"] = self.carrier_id.id
+            self.env.context = ctx
```
pedrobaeza: This can't be done this way, as you are tricking the system through Python internals.

CarlosRoca13:
@pedrobaeza ups! Changed

delivery_auto_refresh/models/sale_order.py
```diff
@@ -29,14 +57,16 @@ def create(self, vals):
...
-            delivery_line = order.order_line.filtered("is_delivery")
-            if len(delivery_line) > 1:
-                continue
-            order.with_context(
-                delivery_discount=delivery_line.discount,
-            )._auto_refresh_delivery()
+        res = super().write(vals)
+        if (
+            self._get_param_auto_add_delivery_line()
+            and "auto_refresh_delivery" not in self.env.context
```
joao-p-marques: This would be more resilient:
```suggestion
            and not self.env.context.get("auto_refresh_delivery")
```

delivery_auto_refresh/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2018 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Auto-refresh delivery",
+    "summary": "Auto-refresh delivery price in sales orders",
+    "version": "13.0.1.0.0",
+    "category": "Delivery",
+    "website": "https://github.com/OCA/delivery-carrier",
+    "author": "Tecnativa, " "Odoo Community Association (OCA)",
```
victoralmau:
```suggestion
    "author": "Tecnativa, Odoo Community Association (OCA)",
```

delivery_auto_refresh/models/sale_order.py
```diff
@@ -0,0 +1,90 @@
...
+        order = super().create(vals)
+        order._auto_refresh_delivery()
+        return order
+
+    def write(self, vals):
+        """Create or refresh delivery line after saving."""
+        res = super().write(vals)
+        if (
+            self._get_param_auto_add_delivery_line()
+            and "auto_refresh_delivery" not in self.env.context
```
victoralmau:
```suggestion
            and not self.env.context.get("auto_refresh_delivery")
```

victoralmau (changes requested):

CarlosRoca13:
@joao-p-marques @victoralmau changes done

victoralmau (approved):

delivery_auto_refresh/models/sale_order.py
```diff
@@ -0,0 +1,89 @@
...
+    @api.depends("partner_id")
+    def _compute_carrier_id(self):
+        if hasattr(super(), "_compute_carrier_id"):
+            super()._compute_carrier_id()
+        for rec in self:
+            if rec.partner_id:
+                if rec.partner_id.property_delivery_carrier_id:
+                    rec.carrier_id = rec.partner_id.property_delivery_carrier_id
+                else:
+                    rec.carrier_id = self.env["delivery.carrier"].search(
```
pedrobaeza: Why always adding one carrier?

delivery_auto_refresh/models/sale_order.py
```diff
@@ -0,0 +1,89 @@
...
+                )._auto_refresh_delivery()
+        return res
+
+    def _create_delivery_line(self, carrier, price_unit):
+        """Allow users to keep discounts to delivery lines. Unit price will
+           be recomputed anyway"""
+        sol = super()._create_delivery_line(carrier, price_unit)
+        discount = self.env.context.get("delivery_discount")
+        if discount and sol:
+            sol.discount = discount
```
pedrobaeza, line 68: Can you please test this use case?

delivery_auto_refresh/models/sale_order.py
```diff
@@ -0,0 +1,89 @@
...
+        return sol
+
+    def set_delivery_line(self, carrier, amount):
+        if self._get_param_auto_add_delivery_line() and self.state in {"draft", "sent"}:
+            self.carrier_id = carrier.id
```
pedrobaeza, line 73: This is also suspicious to not be covered by tests

delivery_auto_refresh/models/sale_order.py
```diff
@@ -0,0 +1,89 @@
...
+        else:
+            return super().set_delivery_line(carrier, amount)
+
+    def _remove_delivery_line(self):
+        new_carrier = self.carrier_id
```
pedrobaeza: A better nomenclature is to call it `current_carrier`

CarlosRoca13:
@pedrobaeza changes done

pedrobaeza (approved):

PR #363, opened 2021-04-20
[14.0][MIG] base_delivery_carrier_files

StefanRijnhart:


base_delivery_carrier_files/models/stock_move.py
```diff
@@ -0,0 +1,19 @@
...
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def write(self, values):
+        write_result = super(StockMove, self).write(values)
+        if values.get("state") == "done":
+            done_pickings = self.mapped("picking_id").filtered(
+                lambda sp: sp.state == "done"
+            )
```
hailangvn, line 16: May I know the benefit of using filtered+lambda over search+domain? Thanks

hailangvn (approved):
Code review: LGTM. Thank you.

base_delivery_carrier_files/models/stock_move.py
hailangvn: Example: https://github.com/odoo/odoo/blob/c08082d89782b7675e8e649cba15ca9e03e40077/addons/purchase_stock/tests/test_replenish_wizard.py#L51

StefanRijnhart: Good question! Because the processes of pickings and their stock moves are very much intertwined, it is likely that at this point we will already have the picking states in our cache so then we avoid the search as well as a potential read. Also, when setting the state of more than one stock move, it is unlikely that there are a large number of pickings involved. I would not prefer to use `filtered` if the result set was potentially large (but it also depends on whether the column used in the domain is indexed, which happens to be the case here).

silverf0x0:
Hi there. What is missing here. Any help needed?

StefanRijnhart:
@silverf0x0 this is a standard migration PR (and a tiny API change when storing the output, in the last commit). It's waiting for a second review.

silverf0x0:
@StefanRijnhart Understood. Thank you.

StefanRijnhart:
@silverf0x0 by the way, that review could come from you if you have the time!

silverf0x0:
@StefanRijnhart Nice to ear that, because in this very moment I'm testing your work. Any specifc procedure is needed for the review, or I just need to be sure that everything is working?

StefanRijnhart:
@silverf0x0 Great, thanks! If you successfully tested the module, please mention that in your approval comment. No specific procedure applies.

silverf0x0:
@StefanRijnhart Excellent, so I'll finish all the tests and I'll approve the review. Thanks to you!

silverf0x0:
@StefanRijnhart I've completed to test everything and for me work/look nice. Maybe the only thing we could add, as improvement, is a UserError (or ValidationError) withing models/stock_picking.py, within generate_carrier_files function:

- when a carrier is not present in an outgoing picking it would not generate a file, as per standard. In this case would be useful to let the user know why a file is not being generated (but in this case the raise could block the generation of other files, which is a little bit annoying). Maybe we can just leave a to-do in the comment?

Thank you in advance

silverf0x0 (approved):
Tested locally several times, without issues.
Code has been reviewed and is fine.

silverf0x0 (approved):

base_delivery_carrier_files/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+# Author: Guewen Baconnier
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Base Delivery Carrier Files",
+    "version": "14.0.1.0.0",
+    "category": "Generic Modules/Warehouse",
+    "author": "Camptocamp,Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "website": "https://github.com/OCA/delivery-carrier",
+    "depends": ["base", "stock", "delivery"],
```
gfcapalbo, line 11: in v14 delivery implies stock
delivery -> sale_stock ->  stock_account -> stock

base_delivery_carrier_files/models/stock_picking.py
```diff
@@ -0,0 +1,60 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def generate_carrier_files(self, auto=True, recreate=False):
+        """
```
gfcapalbo, line 12: fails silently when a picking with  no carrier or no carrier.carrier_file_id  is selected.
Perhaps some feedback (like a warning icon), or  just filtering out non-eligible stock.picking from wizard could be a usability improvement.

gfcapalbo (approved):
Tested locally, generated the CSV, all looks good , just some observations on usability, those can be discarded because the module will be used by other modules that can add usability niceties themselves.
LGTM.

PR #364, opened 2021-04-22
[13.0][FIX] partner_delivery_zone: Restore missing commits from v12

sergio-teruel:
cc @Tecnativa TT29350
ping @chienandalu @carlosdauden

### DIFF
--- a/partner_delivery_zone/models/partner_delivery_zone.py
+++ b/partner_delivery_zone/models/partner_delivery_zone.py
@@ -9,3 +9,4 @@ class PartnerDeliveryZone(models.Model):
 
     code = fields.Char()
     name = fields.Char(string="Zone", required=True)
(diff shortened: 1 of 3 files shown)


chienandalu (commented):
Aren't you missing some ot those changes? Why not cherry picking those commits instead?

https://github.com/OCA/delivery-carrier/commit/d43a3093a68c687efaf554fbbea23322a7385268
https://github.com/OCA/delivery-carrier/commit/277a6925666e6fdc7e1793d7f5f17e230ebe6c93

sergio-teruel:
About the smart-button, in v13 is not needed... you can do that from action button. Is the record is archived a ribbon band appears.
Added some changes

carlosdauden (approved):

pedrobaeza (approved):

PR #367, opened 2021-05-11
[13.0] Add delivery_postlogistics_dangerous_goods

mmequignon:
This modules overrides `Postlogistics Shipping` to declare dangerous goods
during postlogistics label generation.

Based on:
- https://github.com/OCA/delivery-carrier/pull/357

Supersedes:
- https://github.com/OCA/delivery-carrier/pull/342


sebalix (approved):

delivery_postlogistics_dangerous_goods/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "complexity": "normal",
+    "depends": [
+        # OCA/delivery-carrier
+        "delivery_postlogistics",
+        # OCA/community-data-files
+        "l10n_eu_product_adr",
+    ],
+    "website": "https://github.com/OCA/delivery-carrier",
+    "installable": True,
+    "auto_install": False,
```
TDu: `auto_install` False is the default, so not really needed, here.
But does it not make sense to have it set to `True` ?

simahawk: @mmequignon ping

simahawk (approved):
LG, one comment by @TDu needs answer

delivery_postlogistics_dangerous_goods/__manifest__.py
mmequignon: Yeah, I don't know in what case we should enable auto-install or not

TDu: With `auto_install` set to `True` your module will be automatically installed when both its dependencies are installed. Which in my opinion makes sense in this case.

sebalix: It makes sense to set it to `True` as this module doesn't break the base PostLogistics module.

grindtildeath:
@mmequignon Can you please fix the CI ?

sebalix (commented):
Don't mention `13.0` in the commit message as commits could be cherry-picked from one version to another.

mmequignon:
> Don't mention `13.0` in the commit message as commits could be cherry-picked from one version to another.

done

sebalix (approved):

sebalix:
Let's try again...
/ocabot merge nobump

yvaucher:
Is it intended to have that kind of output? It seems too verbose and not giving significant informations. The label binary gives a lot of noise.
It could reach some limit of the output or something like that.

Travis is still unhappy.


https://app.travis-ci.com/github/OCA/delivery-carrier/jobs/543251740#L1416

nilshamerlinck:
Hi @mmequignon we faced this issue before and a workaround was to mute vcr logger for the case involving the big response; I tried it [here](https://github.com/nilshamerlinck/delivery-carrier/commit/ffe84dde7239804792d7a9250a2d0d231610cf05), seems to work

mmequignon:
@dreispt thanks Daniel, but I think there's some action to do on my side before trying.
I have no time right now, but will fine to find some.
Sorry about that…

rousseldenis:
@mmequignon No news on this ?

simahawk:
@mmequignon ping

simahawk:
we keep only v14 https://github.com/OCA/wms/pull/334

PR #372, opened 2021-06-02
[12.0][IMP] delivery_carrier_label_paazl: split methods

astirpe:
This is a proposal of splitting methods to ease customizations.

CC @hbrunn

### DIFF
--- a/delivery_carrier_label_paazl/models/stock_picking.py
+++ b/delivery_carrier_label_paazl/models/stock_picking.py
@@ -28,44 +28,64 @@ class StockPicking(models.Model):
         """Register this picking and commit to paazl immediately,
         return a list of dicts {'exact_price': 'tracking_number':}
         suitable for delivery.carrier#send_shipping"""
+        self._paazl_check_picking_consistency()
+        self._set_a_default_package()
         result = []
-        api = self._paazl_api()
+        for this in self:
+            this._paazl_send_update_order()
+            result.append(this._paazl_generate_labels())
+        return result
+
+    def _paazl_check_picking_consistency(self):
+        """ Check consistency of picking data before sending to Paazl """
         for this in self:
             if len(this.option_ids) != 1:
                 raise UserError(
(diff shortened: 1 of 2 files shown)


hbrunn (approved):
I think this is okay as it is, but wouldn't it be more convenient for the user (=programmer) of this module when you do the looping in `_paazl_order_send` and the functions you introduce operate on self?

astirpe:
@hbrunn do you mean this?

```python
    def _paazl_send(self):
        result = []
        for this in self:
            this._paazl_check_picking_consistency()
            this._set_a_default_package()
            this._paazl_send_update_order()
            result.append(this._paazl_generate_labels())
        return result
```

hbrunn (approved):
exactly!

astirpe:
@hbrunn I think it's good to have `_paazl_check_picking_consistency()` and `_set_a_default_package()` outside the loop, so that it will be ensured that all the pickings have consistent data before starting to exchange data with Paazl.

astirpe:
@hbrunn I added this commit https://github.com/OCA/delivery-carrier/pull/372/commits/23f0206150e993faa243a1feb428d4f8e315dceb, to allow to update the order data. Do you agree with it? And if you agree, do you think it's worth a separate PR?

astirpe:
@hbrunn do you agree with https://github.com/OCA/delivery-carrier/pull/372/commits/49cb363f3c92a8cfd4efe79811c33e6fe91b22c1 ?

hbrunn:
@astirpe totally. If you keep working on this, would it make sense to mark your PR as draft? Then new commits don't generate a notification for subscribers

dkubiak789 (approved):

StefanRijnhart:
@astirpe can you squash commits?

astirpe:
@StefanRijnhart squashed, thanks!

StefanRijnhart:
@astirpe thanks!

/ocabot merge patch

PR #375, opened 2021-06-07
[14][base_delivery_carrier_label]Allow to have multiple carrier accounts for a same delivery type

florian-dacosta:
Sometimes we may have multiple delivery methods for a same carrier/api and some carriers may require to have different accounts depending on the delivery method used.
The goal of this small PR is to allow this case and make the carrier account choice a bit more flexible.
Still it has no impact if this feature is not needed as the new field on account is optional.

@bealdav @yvaucher @hparfr

### DIFF
--- a/base_delivery_carrier_label/models/stock_picking.py
+++ b/base_delivery_carrier_label/models/stock_picking.py
@@ -147,15 +147,22 @@ class StockPicking(models.Model):
         vals = self._values_with_carrier_options(vals)
         return super().create(vals)
 
+    def _get_carrier_account_domain(self):
+        return [
+            ("delivery_type", "in", self.mapped("delivery_type")),
+            "|",
+            ("company_id", "=", False),
+            ("company_id", "in", self.mapped("company_id.id")),
+            "|",
+            ("carrier_ids", "=", False),
+            ("carrier_ids", "in", self.mapped("carrier_id").ids),
+        ]
+
     def _get_carrier_account(self):
         """ Return a carrier suitable for the current picking """
+        domain = self._get_carrier_account_domain()
         return self.env["carrier.account"].search(
-            [
-                ("delivery_type", "in", self.mapped("delivery_type")),
-                "|",
-                ("company_id", "=", False),
-                ("company_id", "in", self.mapped("company_id.id")),
-            ],
+            domain,
             limit=1,
             order="company_id asc, sequence asc",
         )
--- a/base_delivery_carrier_label/models/carrier_account.py
+++ b/base_delivery_carrier_label/models/carrier_account.py
@@ -23,6 +23,18 @@ class CarrierAccount(models.Model):
         .selection,
         help="This field may be used to link an account to a carrier",
     )
+    carrier_ids = fields.Many2many(
+        "delivery.carrier",
+        "delivery_carrier_account_rel",
+        "account_id",
+        "carrier_id",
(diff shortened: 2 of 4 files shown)


base_delivery_carrier_label/models/stock_picking.py
```diff
@@ -147,15 +147,22 @@ def create(self, vals):
...
 
+    def _get_carrier_account_domain(self):
+        return [
+            ("delivery_type", "in", self.mapped("delivery_type")),
+            "|",
+            ("company_id", "=", False),
+            ("company_id", "in", self.mapped("company_id.id")),
+            "|",
+            ("carrier_ids", "=", False),
+            ("carrier_ids", "in", self.mapped("carrier_id").ids),
```
hparfr, line 158:
```suggestion
            ("company_id", "in", ('False' + self.mapped("company_id").ids)),
            "|",
            ("carrier_ids", "in", (False + self.mapped("carrier_id").ids)),
```

hparfr (commented):
Back in the days we had some `inversion of control` coding style to avoid to link something to a carrier account in order to allow a multi-env / multi-account.
Is it still possible with your patch ?

PS: the goal were:
- be able to manage multi-account with the same carrier (like you do dropshipping for someone else and use the customer account to ship)
- having different account for different environments (test / ci / prod)

base_delivery_carrier_label/models/stock_picking.py
florian-dacosta: Your proposal does not seem to work.
domain `('x2many', 'in', [False])` does not seem to work.

florian-dacosta:
> Back in the days we had some `inversion of control` coding style to avoid to link something to a carrier account in order to allow a multi-env / multi-account.
> Is it still possible with your patch ?
> 
> PS: the goal were:
> 
>     * be able to manage multi-account with the same carrier (like you do dropshipping for someone else and use the customer account to ship)
> 
>     * having different account for different environments (test / ci / prod)

Actually, the goal of the PR is to be able to have multiple account for a same type of carrier.
It does not allow, as it is to have multiple account for a same delivery method. But, the domain is now returned in a separeted method, which allow anyone to override it, and implementer a new field on carrier account, and then use it in the domain, overriding `_get_carrier_account_domain`.
I believe the dropshipping case is manageable this way.

About the different accounts per environment, actually, it should be managed with server_environment module and a carrier account specific module as in version 12 : https://github.com/OCA/server-env/tree/12.0/carrier_environment
@hparfr

hparfr (approved):
Thanks for the clarifications (and for the work :) )

yvaucher:
@florian-dacosta what we did in the past was simply to create multiple delivery methods. And then for each delivery method we were choosing which license was in use. But keep in mind that for licenses we are using a dedicated object and not `carrier.account`.

I don't know your usage but I don't see how `_get_carrier_account_domain` should be used, as to me the account to use is simply the one set on the `delivery.carrier`. Otherwise a match rule would have to be computed based on additional fields on the `carrier.account`. Well, you could still implement a function that makes a call to know the capacities of an account. But you better want to store that info instead of getting it on each call.

Last version of `postlogistics` module dropped the dependency on `base_delivery_carrier_label` to integrate more with last Odoo version and logic (for the good, and the bad).

yvaucher (approved):
Code review

PR #376, opened 2021-06-07
[MIG][REF] delivery_carrier_label_gls

len-foss:
It's actually essentially rewritten from scratch to use the REST API.
It's available in multiple countries, including Belux, Germany, Austria, excluding France.
You need to contact GLS to check if it is available in your country, as they seem unable to provide the complete list. 

Caution: 
The API does not offer any idempotent way to make the calls, so any lost call to "createParcel" will be invoiced, even if the parcel is never shipped (wew lad!). Therefore, a call to cancelShipment should be done in case there is a rollback.
However this was only done in https://github.com/odoo/odoo/pull/54321, so you need to pull it in your sources if you want that safety net.


pedrobaeza:
@chienandalu is this webservice the same as the Spanish one?

len-foss:
@pedrobaeza AFAIK this is supposed to be the international API, except for France (they use a Minitel-based protocol probably). Is there already an implementation?

pedrobaeza:
https://github.com/OCA/l10n-spain/tree/12.0/delivery_gls_asm

len-foss:
Oh thanks for the pointer.
It's definitely not the same API. I'll have to check in which countries this one can really be used.

pedrobaeza:
The one in Spanish localization is AFAIK inherited from the old carrier (ASM) that was absorbed by GLS.

delivery_carrier_label_gls/tests/common.py
```diff
@@ -0,0 +1,152 @@
...
+                                        {u"Name": u"Phone", u"Value": u""},
+                                    ],
+                                }
+                            ]
+                        },
+                        u"TrackID": u"ZMZE0SRO",
+                    }
+                ],
+                u"PickupLocation": u"BE7100",
+                u"PrintData": [{u"Data": u"JVeryLongString==", u"LabelFormat": u"PDF"}],
```
lmignon, line 123: @len-foss The LabelFormat should be configurable on the carrier info

lmignon: @len-foss see https://github.com/acsone/delivery-carrier/pull/1

rousseldenis:
@len-foss Could you check Travis errors (flake8 and pylint)

delivery_carrier_label_gls/models/stock_picking.py
```diff
@@ -0,0 +1,65 @@
...
+    _inherit = "stock.picking"
+
+    gls_parcel_shop = fields.Char(
+        "GLS Parcel Shop Identifier", help="Fill this for a delivery to a ParcelShop.",
+    )
+    gls_package_ref = fields.Char("GLS Package Identifiers", readonly=True, copy=False)
+
+    def _check_is_everything_packaged(self):
+        for picking in self:
+            if not all(o.result_package_id for o in picking.pack_operation_product_ids):
```
lmignon, line 29: @len-foss what about `pack_operation_pack_ids`?

PR #378, opened 2021-06-09
[13.0] delivery_postlogistics: Handle case where no label is returned by the API

mmequignon:
Sometimes, no sucess label is returned by the API.
In such case, we should just ignore it, and raise the errors.

### DIFF
--- a/delivery_postlogistics/models/stock_picking.py
+++ b/delivery_postlogistics/models/stock_picking.py
@@ -155,6 +155,11 @@ class StockPicking(models.Model):
         zpl_patch_string = self.carrier_id.zpl_patch_string
 
         labels = []
+
+        # It could happen that no successful label has been returned by the API
+        if not label_result:
+            return labels
+
         if not packages:
             label = label_result[0]["value"][0]
             self.carrier_tracking_ref = label["tracking_number"]


delivery_postlogistics/models/stock_picking.py
```diff
@@ -155,6 +155,11 @@ def write_tracking_number_label(self, label_result, packages):
         zpl_patch_string = self.carrier_id.zpl_patch_string
 
         labels = []
+
+        # It could happen that no successful label has been returned by the API
+        if not label_result:
+            return labels
```
sebalix, line 161: I would not place this check in `write_tracking_number_label` but perhaps in `_generate_postlogistics_label` where `success_label_results` (and `failed_label_results`) are initialized => https://github.com/OCA/delivery-carrier/blob/4a66d8ed7207468e007b411ea28914dabe86784a/delivery_postlogistics/models/stock_picking.py#L205-L217

Furthermore it would be interesting to know why the label hasn't been generated? Is the `failed_label_results` empty too? On the link above you can see that if `failed_label_results` is not empty, a rollback is performed but no exception is raised, so the flow continues as if nothing happened: why it has been done like this? Is it on purpose?
ping @phuctranerp @jbaudoux

mmequignon: If you ask a label for a single package and this one is errored (in the case raised by the customer, weight was >30kgs), then success_label_results is empty.

mmequignon: The reason why I placed this check here was to avoid code like this:
```python3
labels = []
if success_label_results:
    labels = self.write_tracking_number_label(success_label_results, packages)
```

sebalix: But you say that `success_label_results` is empty, ok, what about `failed_label_results`? Because `write_tracking_number_label` is called after the rollback, and maybe instead of a rollback we simply need to raise an exception?

mmequignon: failed_label_results isn't empty.
The exception is raised after, with the error message if `failed_label_results` is not empty, whether `success_label_result` is empty or not.
What I don't understand though, is that if failed_label_results is not empty, a `rollback()` is made and a `commit()` a few lines after.
We maybe want to attach valid labels anyway, and raise the exception for errored ones.

sebalix: Yes this is also an issue we encountered with @santostelmo the other day (related to this fix: https://github.com/OCA/delivery-carrier/pull/371): we were losing the packages generated in the same transaction with this rollback, even the good ones.

Ok then for your fix!

sebalix (approved):

simahawk (approved):

PR #380, opened 2021-06-10
[14.0][ADD] delivery_carrier_city

ivantodorovich:
Integrates `delivery` with `base_address_city`, to allow to use cities in the delivery methods' destination conditions.

PR #381, opened 2021-06-10
[14.0][ADD] delivery_carrier_location

ivantodorovich:
Depends on https://github.com/OCA/delivery-carrier/pull/380

Integrates `delivery` with `base_location`, to allow to use `res.city.zip` records in the delivery methods' destinations.

### DIFF
--- a/delivery_carrier_location/models/delivery_carrier.py
+++ b/delivery_carrier_location/models/delivery_carrier.py
@@ -1,0 +1,26 @@
+# Copyright 2021 Camptocamp SA - Iván Todorovich
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = "delivery.carrier"
+
+    zip_ids = fields.Many2many(
+        "res.city.zip",
+        relation="delivery_carrier_zip_rel",
+        column1="carrier_id",
+        column2="zip_id",
+        string="Zip codes",
+    )
+
+    def _match_address(self, partner):
+        # Override to account for city_ids and zip_ids
+        res = super()._match_address(partner)
+        # Fail quickly if super already rejected it
+        if not res or not self.zip_ids:
+            return res  # pragma: no cover
+        # Check partner's zip.
+        partner_zip = (partner.zip or partner.zip_id.name or "").upper()
+        return partner_zip in {r.name.upper() for r in self.zip_ids}
--- a/delivery_carrier_location/tests/test_delivery_carrier_location.py
+++ b/delivery_carrier_location/tests/test_delivery_carrier_location.py
@@ -1,0 +1,150 @@
+# Copyright 2021 Camptocamp SA - Iván Todorovich
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests import common
+
+
+class TestDeliveryCarrierCity(common.SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        # Countries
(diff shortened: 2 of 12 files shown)


delivery_carrier_location/models/delivery_carrier.py
```diff
@@ -0,0 +1,34 @@
...
+        column1="carrier_id",
+        column2="zip_id",
+        string="Zip codes",
+    )
+
+    def _match_address(self, partner):
+        # Override to account for city_ids and zip_ids
+        res = super()._match_address(partner)
+        # Fail quickly if super already rejected it
+        if not res:
```
simahawk:
```suggestion
        if not res or not self.zip_ids:
            return False
```

delivery_carrier_location/models/delivery_carrier.py
```diff
@@ -0,0 +1,34 @@
...
+        # If partner doesn't have zip_id try to match using the zip code.
+        if self.zip_ids:
+            if partner.zip_id:
+                if partner.zip_id not in self.zip_ids:
+                    return False
+            else:
+                allowed_zip_codes = {r.name.upper() for r in self.zip_ids}
+                if (partner.zip or "").upper() not in allowed_zip_codes:
+                    return False
+        return res
```
simahawk:
```suggestion
        partner_zip = partner.zip or partner.zip_id.name
        if not partner_zip:
            return res
        return partner_zip.uppper() in {r.name.upper() for r in self.zip_ids}
```

delivery_carrier_location/tests/test_delivery_carrier_location.py
```diff
@@ -0,0 +1,149 @@
+# Copyright 2021 Camptocamp SA - Iván Todorovich
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests import common
+
+
+class TestDeliveryCarrierCity(common.SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
simahawk, line 10:
```suggestion
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```

delivery_carrier_location/tests/test_delivery_carrier_location.py
```diff
@@ -0,0 +1,149 @@
...
+                "country_ids": [(4, cls.france.id)],
+            }
+        )
+
+    def _get_available_carriers(self, partner):
+        return self.env["delivery.carrier"].search([]).available_carriers(partner)
+
+    def test_00_partner_zip_id(self):
+        # Partner living in paris 03, with zip_id set
+        partner = self.env["res.partner"].create(
```
simahawk, line 120: can't you use existing demo partners and save some tree? :)

oca_dependencies.txt
```diff
@@ -1 +1,2 @@
 # See https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#oca_dependencies-txt
+partner-contact
```
simahawk, line 2: I'd propose to enable MQT=PIP here, but that's another story :)

simahawk (changes requested):

delivery_carrier_location/tests/test_delivery_carrier_location.py
ivantodorovich: I'd still need to write on them to set zip_id (not in core) so not sure it'll save anything :sweat_smile:

delivery_carrier_location/models/delivery_carrier.py
ivantodorovich: Thanks!
Took your suggestion but changed it a bit.
Empty partner zip shouldn't return True, it should fail

simahawk (approved):
the other PR is getting merged

simahawk:
@ivantodorovich can you rebase?

PR #382, opened 2021-06-16
[IMP] delivery_auto_refresh: void delivery lines on uninvoiced returns

chienandalu:
Added new config paramerter that automates de voiding of the delivery
line for the case of returning a picking that just didn't go away from
the warehouse (due to several causes like the customer cancelling the
sale order in the last minute).

This will only be performed if the order wasn't already invoiced and the
return was set to refund.

cc @Tecnativa TT30359

please review @pedrobaeza @CarlosRoca13

### DIFF
--- a/delivery_auto_refresh/README.rst
+++ b/delivery_auto_refresh/README.rst
@@ -52,6 +52,11 @@ Configuration
   or create a new one if not exists.
   Put a non Falsy value (1, True...) if you want to refresh delivery price
   after transferring.
+* Locate the setting with key "delivery_auto_refresh.auto_void_delivery_line"
+  or create a new one if it doesn't exists.
+  Put a non Falsy value (1, True...) if you want to void the delivery line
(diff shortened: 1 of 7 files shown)


CarlosRoca13 (approved):

joao-p-marques (commented):
Code and functional review

I found a little error and I don't know if it is related to this change:

https://user-images.githubusercontent.com/38977934/122343257-6dfa3180-cf3d-11eb-81de-884ac503ca91.mp4

Am I doing something wrong in the process?

chienandalu:
> I found a little error and I don't know if it is related to this change:

Looks like it's related to that provider (UPS), that maybe doesn't have the proper extension methods implemented. Check it with the standard ones.

pedrobaeza (approved):

pedrobaeza:
Needs to be forward-port it to v13:

/ocabot merge major

PR #383, opened 2021-06-21
[IMP] delivery_auto_refresh: void delivery lines on uninvoiced returns

chienandalu:
forward port of #382 

Added new config paramerter that automates de voiding of the delivery
line for the case of returning a picking that just didn't go away from
the warehouse (due to several causes like the customer cancelling the
sale order in the last minute).

This will only be performed if the order wasn't already invoiced and the
return was set to refund.

cc @Tecnativa TT30359

please review @pedrobaeza @CarlosRoca13

### DIFF
--- a/delivery_auto_refresh/README.rst
+++ b/delivery_auto_refresh/README.rst
@@ -52,6 +52,11 @@ Configuration
   or create a new one if not exists.
   Put a non Falsy value (1, True...) if you want to refresh delivery price
   after transferring.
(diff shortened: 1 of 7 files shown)


pedrobaeza (approved):
Code review

CarlosRoca13 (approved):

pedrobaeza:
/ocabot merge patch

Is this needed in v14? @chienandalu

pedrobaeza:
@chienandalu Please check Travis

chienandalu:
Now it should pass

PR #387, opened 2021-06-28
[12.0] Fix Postlogistics webservice error handling

guewen:
The webservice class returns a list of dictionaries.


leemannd (approved):

yvaucher (approved):
Thanks.

Ok, looks like this part of the code was 8 years old which is even older than the change from SOAP to REST.

`res` is indeed a list of dict labels which was introduced here

https://github.com/OCA/delivery-carrier/commit/6e1447a981e4c90ec644f4db8cbf769c9bc0c9cf#diff-d3c06dad2d58ebdeb558386909ed5b2ad9ed22d84b8476e5fb2658ae4308d61fR430

And the error handling was missing in that commit.

PR #388, opened 2021-06-28
[12.0] Fix postlogistics error when email contains a "realname"

guewen:
When the webservices receives a 2 parts email such as `Real Name
<email@example.com>`, it returns a 500 internal error.
Extract the email address only and send only this part to the service.

Note: parseaddr is pretty tolerant regarding the input, even if it
receives False or '', it returns ('', ''), so we don't need to be
very defensive.


leemannd (approved):

hparfr (approved):
/ocabot merge patch

PR #401, opened 2021-08-04
[13.0][MIG] delivery_ups_oca: Migration 13.0 + Refactor

victoralmau:
The following changes have been implemented:

- [x] Migration to 13.0
- [x] Refactor code
- [x] Added webservice documentation..
- [x] Rename `delivery_carrier_label_ups` to `delivery_ups_oca` + Rename in Openpupgrade. (https://github.com/OCA/OpenUpgrade/pull/2885)

Locked by:
- [x] https://github.com/OCA/delivery-carrier/pull/402

@Tecnativa TT30668


pedrobaeza:
Please do the renaming since the beginning, changing also the name on the commit history for not losing anything:
```
git filter-branch --tree-filter 'if [ -d old_module_name ]; then mv old_module_name new_module_name; fi' HEAD
```

delivery_ups_oca/README.rst
```diff
@@ -1,32 +1,34 @@
...
 .. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
-    :target: https://runbot.odoo-community.org/runbot/99/12.0
+    :target: https://runbot.odoo-community.org/runbot/99/13.0
     :alt: Try me on Runbot
 
 |badge1| |badge2| |badge3| |badge4| |badge5| 
 
 This module adds `UPS <https://ups.com>`_ to the available carriers.
 
+Register shipping, generate label, read shipping state and cancel shippment by webservices.
```
pedrobaeza: Remember to split lines with the same 80 cols criteria on text files for better diff reading:

```suggestion
It allows you to register shippings, generate labels, read shipping states and 
cancel shipments using UPS webservice, so no need of exchanging any kind of file.
```

delivery_ups_oca/README.rst
```diff
@@ -38,16 +40,17 @@ Configuration
 To configure this module, you need to:
 
 #. Add options to your UPS carrier
+#. Go to *Inventory > Configuration* and check "Delivery Packages" option.
```
pedrobaeza: Why do you need this?

delivery_ups_oca/README.rst
```diff
@@ -38,16 +40,17 @@ Configuration
...
 #. Add a carrier account with delivery type ``ups`` and fill in your credentials
 
 Usage
 =====
 
-To use this module, you need to:
+You have to set Ups carrier in the stock picking you want to ship:
```
pedrobaeza:
```suggestion
You have to set the created shipping method in the delivery order to ship:
```

delivery_ups_oca/README.rst
```diff
@@ -38,16 +40,17 @@ Configuration
...
 
-#. Select the ups carrier on your sale orders or pickings
-#. Set exactly one delivery option
-#. Press the button to generate labels for a done picking
+ * In the stock picking form go to *Additional Info* tab and choose Ups as carrier and the service and product code. You only be able to choose this if the state of the picking is 'Ready to Transfer.
```
pedrobaeza: Don't forget these are user instructions, not developer ones, so use UI terminology:
```suggestion
* In the delivery order form go to *Additional Info* tab and choose the UPS shipping
  method you created as carrier and the service and product code.
* You will be only able to choose this if the state of the delivery is "Ready to Transfer".
```
Another thing is that the choosing should be done also on state "Partially available".

delivery_ups_oca/README.rst
```diff
@@ -38,16 +40,17 @@ Configuration
...
+
+ * When the picking is 'Transferred', it appears a *Create Shipping Label* button. Just push it, and if all went well the label will be 'attached'
```
pedrobaeza: Put instructions for checking the delivery state (although I think there's a general cron for this, but it worths to mention all these aspects).

No automatic label generation? Isn't this an option on other carriers?

Mention also about cancelling.

Other thing: what about pricing?

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -1,22 +1,285 @@
...
+            ("86", "UPS Today Express Saver"),
+            ("96", "UPS Worldwide Express Freight"),
+        ],
+        default="11",
+        string="Service code",
+    )
+    ups_default_packaging_id = fields.Many2one(
+        comodel_name="product.packaging", string="Default Packaging Type"
+    )
+    ups_package_dimension_code = fields.Char(string="Package Dimension code")
```
pedrobaeza: What are these parameters? A big help string is needed here.

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -1,22 +1,285 @@
...
+    ups_package_weight_code = fields.Char(string="Package Weight code")
+
+    def _ups_set_url(self, ups_type):
```
pedrobaeza: This is not setting anything:

```suggestion
def _ups_get_url(self, ups_type):
```

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -1,22 +1,285 @@
...
+                self.env["res.currency"].search(
+                    [("name", "=", total_charges["CurrencyCode"])]
+                ),
+                picking.company_id,
+                fields.Date.today(),
+            )
+        picking.carrier_tracking_ref = ShipmentResults["ShipmentIdentificationNumber"]
+        # Add message to add .pdf file and auto-attachment
+        ShippingLabel = ShipmentResults["PackageResults"]["ShippingLabel"]
+        message = _("Tracking number: %s" % picking.carrier_tracking_ref)
```
pedrobaeza: Is this is a usual practice? There's already a field for putting this information, and that field can have `tracking=True`

delivery_ups_oca/data/product_packaging_data.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2021 Tecnativa - Víctor Martínez
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="product_packaging_ups" model="product.packaging">
+        <field name="name">UPS</field>
```
pedrobaeza:
```suggestion
        <field name="name">UPS default</field>
```

delivery_ups_oca/README.rst
pedrobaeza: Also service code is defined on the carrier and should be in CONFIGURE.rst.

delivery_ups_oca/views/delivery_carrier_view.xml
```diff
@@ -0,0 +1,41 @@
...
+        <field name="model">delivery.carrier</field>
+        <field name="inherit_id" ref="delivery.view_delivery_carrier_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//notebook" position="inside">
+                <page
+                    string="Integration"
+                    attrs="{'invisible': ['|', ('delivery_type', '!=', 'ups'), ('integration_level', '!=', 'rate_and_ship')]}"
+                >
+                    <group
+                        attrs="{'required': [('delivery_type', '=', 'ups'), ('integration_level', '=', 'rate_and_ship')]}"
```
pedrobaeza: required can't be set for `<group>`

delivery_ups_oca/views/delivery_carrier_view.xml
```diff
@@ -0,0 +1,41 @@
...
+License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+-->
+<odoo>
+    <record id="view_delivery_carrier_form" model="ir.ui.view">
+        <field name="model">delivery.carrier</field>
+        <field name="inherit_id" ref="delivery.view_delivery_carrier_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//notebook" position="inside">
+                <page
+                    string="Integration"
```
pedrobaeza: Are other carriers added as a new page?

```suggestion
                    string="UPS"
```

delivery_ups_oca/models/product_packaging.py
```diff
@@ -32,4 +32,6 @@ class ProductPackaging(models.Model):
             ("66", "BPM Flat"),
             ("67", "Standard Flat"),
         ],
+        default="02",
+        string="Ups code",
```
pedrobaeza:
```suggestion
        string="UPS code",
```

pedrobaeza (changes requested):

delivery_ups_oca/README.rst
victoralmau: Sorry, it's not necessary.

delivery_ups_oca/models/delivery_carrier.py
victoralmau: I have modified this so that it only create the attachment with the UPS generated label.

delivery_ups_oca/README.rst
victoralmau: Some explanation added to clarify some of these things.

I don't understand about "_what about pricing?_". Can you get a example in some addon to understand it? (Actually, creation shipping process return price although we don't use in anywhere).

pedrobaeza: On sales order, you are able to get the rate for the shipping. There's a method for that (not sure if `_rate_<carrier_code>` or similar)

delivery_ups_oca/README.rst
```diff
@@ -0,0 +1,109 @@
...
+
+.. contents::
+   :local:
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Add options to your UPS carrier
```
pedrobaeza: I don't get this sentence. And I don't see an explanation of which data is needed to be filled, among some explanation on how to get such credentials.

delivery_ups_oca/README.rst
```diff
@@ -0,0 +1,109 @@
...
+#. Add a carrier account with delivery type ``ups`` and fill in your credentials
+
+Usage
+=====
+
+You have to set the created shipping method in the delivery order to ship:
+
+* In the delivery order form go to *Additional Info* tab and choose the UPS shipping
```
pedrobaeza:
```suggestion
* In the delivery order form, go to *Additional Info* tab and choose the UPS shipping
```

delivery_ups_oca/README.rst
```diff
@@ -0,0 +1,109 @@
...
+  method you created as carrier and the service and product code.
```
pedrobaeza: Again `service and product code` is mentioned, and I didn't see such fields, but even more, this should be at carrier level, not delivery order level.

delivery_ups_oca/README.rst
```diff
@@ -0,0 +1,109 @@
...
+* You will be only able to choose this if the state of the delivery is "Ready to Transfer".
```
pedrobaeza: I put a comment about this before that has been ignored.

delivery_ups_oca/README.rst
```diff
@@ -0,0 +1,109 @@
...
+
+* When the picking is 'Transferred', it appears a *Create Shipping Label* button. Just push it, and if all went well the label will be 'attached'
```
pedrobaeza:
```suggestion
* When the picking is 'Transferred', a *Create Shipping Label* button appears. Just push it, and if all went well, the label will be 'attached'.
```

And I also put some comments about this not being automatic.

delivery_ups_oca/README.rst
```diff
@@ -0,0 +1,109 @@
...
+
+Usage
+=====
+
+You have to set the created shipping method in the delivery order to ship:
+
+* In the delivery order form go to *Additional Info* tab and choose the UPS shipping
+  method you created as carrier and the service and product code.
+* You will be only able to choose this if the state of the delivery is "Ready to Transfer".
+
```
pedrobaeza: No need to put here an empty line. It's only needed on RST when changing the indentation.

delivery_ups_oca/README.rst
```diff
@@ -0,0 +1,109 @@
...
+* When the picking is 'Transferred', it appears a *Create Shipping Label* button. Just push it, and if all went well the label will be 'attached'
+
+* When the picking is 'Cancel' automatically cancel too in UPS.
```
pedrobaeza:
```suggestion
* When the delivery order is cancelled, it's automatically cancelled too in UPS.
```

delivery_ups_oca/README.rst
```diff
@@ -0,0 +1,109 @@
...
+
+* According to delivery_state addon cron exists from auto state update according from UPS (only if checked "Tracking state update sync" in carrier).
```
pedrobaeza: Don't use too much technical words, and put a reference about "Tracking state update sync" in configure section:

```suggestion
* If you have "Tracking state update sync" checked in the shipping method, a periodical state check will be done querying UPS services.
```

delivery_ups_oca/README.rst
```diff
@@ -0,0 +1,109 @@
...
+
+Known issues / Roadmap
+======================
+
+* support international forms
```
pedrobaeza: First letter upper

delivery_ups_oca/README.rst
```diff
@@ -0,0 +1,109 @@
...
+Authors
+~~~~~~~
+
+* Hunki Enterprises BV
+
+Contributors
+~~~~~~~~~~~~
+
+* Holger Brunn <mail@hunki-enterprises.nl> (https://hunki-enterprises.nl)
+* `Tecnativa <https://www.tecnativa.com>`_:
```
pedrobaeza, line 94: Have we added enough code to be co-authors?

pedrobaeza (changes requested):
I still don't see anything about rating in the README.

delivery_ups_oca/README.rst
victoralmau: It's not necessary now.

delivery_ups_oca/README.rst
victoralmau: It's not necessary now.

delivery_ups_oca/README.rst
victoralmau: It's not necessary now.

delivery_ups_oca/README.rst
```diff
@@ -0,0 +1,107 @@
...
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/99/13.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module adds `UPS <https://ups.com>`_ to the available carriers.
+
+It allows you to register shippings, generate labels, get rates from order, read shipping states and
+cancel shipments using UPS webservice, so no need of exchanging any kind of file.
```
pedrobaeza: I see no mention here of getting rates for the shipping

delivery_ups_oca/README.rst
```diff
@@ -0,0 +1,107 @@
...
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Add a carrier account with delivery type ``ups`` and fill in your credentials
+#. Configure the required fields of the UPS tab with your account data* https://wwwapps.ups.com
+#. If yo have "Tracking state update sync" checked all delivery orders state check will be done querying UPS services.
+
+* You will be necessary to request access key from https://www.ups.com/upsdeveloperkit/manageaccesskeys?loc=es_ES
```
pedrobaeza:
```suggestion
**NOTE** You need to request an access key from https://www.ups.com/upsdeveloperkit/manageaccesskeys?loc=es_ES for using the webservice.
```

delivery_ups_oca/README.rst
```diff
@@ -0,0 +1,107 @@
...
+.. contents::
+   :local:
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Add a carrier account with delivery type ``ups`` and fill in your credentials
+#. Configure the required fields of the UPS tab with your account data* https://wwwapps.ups.com
```
pedrobaeza: Specify which fields and the URL seems incorrect. Another thing is that RST syntax doesn't allow `*` as footnotes. Use the proper one or don't put it as footnote (see my following suggestion).

I also miss something about the service code, and about packaging.

delivery_ups_oca/README.rst
```diff
@@ -0,0 +1,107 @@
...
+#. If yo have "Tracking state update sync" checked all delivery orders state check will be done querying UPS services.
+
+* You will be necessary to request access key from https://www.ups.com/upsdeveloperkit/manageaccesskeys?loc=es_ES
+
+Usage
+=====
+
+You have to set the created shipping method in the delivery order to ship:
+
+* When the picking is 'Transferred', a *Create Shipping Label* button appears. Just push it, and if all went well, the label will be 'attached'.
```
pedrobaeza: Remember to break lines on the line length. What happens if not `all went well`? Document it.

```suggestion
* When the picking is 'Transferred', a *Create Shipping Label* button appears. Just 
click on it, and if all went well, the label will be 'attached'.
```

delivery_ups_oca/tests/test_delivery_ups.py
```diff
@@ -0,0 +1,96 @@
...
+        cls.carrier = cls.env["delivery.carrier"].create(
+            {
+                "name": "UPS",
+                "delivery_type": "ups",
+                "product_id": product_shipping_cost.id,
+                "price_method": "fixed",
+                "ups_default_packaging_id": cls.env.ref(
+                    "delivery_ups_oca.product_packaging_ups"
+                ).id,
+                "ups_ws_username": "ups_odoo_test",
```
pedrobaeza, line 32: Put a comment here saying that this account was issued by Odoo S.A. and we are borrowing it from data seen on enterprise runbot.

pedrobaeza (changes requested):
Please check that no comments are left, as I have to repeat several times some of them.

delivery_ups_oca/README.rst
victoralmau: Sorry, but "`get rates from order`" is intended to say that in a simplified way.

pedrobaeza: I didn't see it, sorry. Then I think it's enough, but explain on the usage about how to do it (although being something not specific from UPS, this is user documentation, that can repeat certain parts if convenient).

delivery_ups_oca/README.rst
victoralmau: Ok, I'm going to change the URL to https://wwwapps.ups.com/ppc/ppc.html
According to these comment https://github.com/OCA/delivery-carrier/pull/401#discussion_r684116111 I don't know where the fields to fill in are because I don't have a UPS account to verify it.

pedrobaeza: I mean in Odoo, not in UPS. Mention the required fields, the "Track update..." field, etc.

delivery_ups_oca/README.rst
victoralmau: Ok, I have added more explanation about it.

delivery_ups_oca/README.rst
victoralmau: I have specified that it is in Odoo and that all the required fields in the UPS tab have to be filled in.

pedrobaeza (approved):
Although we can still improve the README, let's continue with this base.

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -1,22 +1,382 @@
...
+        price = self._ups_get_response_price(
+            ShipmentResults["ShipmentCharges"]["TotalCharges"],
+            picking.company_id.currency_id,
+            picking.company_id,
+        )
+        picking.carrier_tracking_ref = ShipmentResults["ShipmentIdentificationNumber"]
+        # Create attachment to add pdf label
+        self.env["ir.attachment"].create(
+            {
+                "name": "%s.pdf" % picking.carrier_tracking_ref,
```
florian-dacosta: Is it normal to force pdf here ? I though label could be in other format, like zpl?

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -1,22 +1,382 @@
...
+                "Weight": str(PackageWeight),
+            },
+        }
+
+    def _ups_quant_package_data_from_picking(self, package, picking, is_package=True):
+        NumOfPieces = picking.number_of_packages
+        PackageWeight = picking.shipping_weight
+        if is_package:
+            NumOfPieces = sum(package.mapped("quant_ids.quantity"))
+            PackageWeight = package.max_weight
```
florian-dacosta: Is not normal to use a max_weight here ? Shouldn't it be the package shipping_weight or weight ?
AFAIK max_weight is on `product.packaging` and here `package` is a `stock.quant.package` or am I mistaken?

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -1,22 +1,382 @@
...
+                "Height": str(package.height),
+            },
+            "PackageWeight": {
+                "UnitOfMeasurement": {"Code": self.ups_package_weight_code},
+                "Weight": str(PackageWeight),
+            },
+            "PackageServiceOptions": "",
+        }
+
+    def _ups_label_data(self, TrackingNumber):
```
florian-dacosta: I don't see where it is called

florian-dacosta (commented):
Code review only for now because I think one need an UPS account to test it.

On global remark though, I think it is a pity that it does not depend on base_delivery_carrier_label, which aims to add some helper, and reusable stuff among carrier.
For instance, the model `carrier_account`, which allow to re-use the same model and some fields between carriers implemntation, and mainly make the way to configure account a lot more configurable that on delivery carrier.
It may be find if you use one or 2 carrier, not if you use 5/6 IMO.

In v13, base_delivery_carrier_label is not well compatible with native Odoo way to generate labels, so it seems ok to do without, but I it has been refactored in v14 and I think it would be nice to implement this on top of it once again to make for this next version.
I believe that if base_delivery_carrier_label is too big/complex, some functionality (like `delivery.carrier.option`) could/should be extracted in a separated module, to keep base_delivery_carrier_label as a helper/unifomizing module for all carrier modules.

But I guess this is a question for the migration to 14.

pedrobaeza:
We don't want to depend on such module and have a non needed maintenance burden.

delivery_ups_oca/models/delivery_carrier.py
victoralmau: Thank you, it is already changed for the different formats.

delivery_ups_oca/models/delivery_carrier.py
victoralmau: Sorry, I had forgotten this, now label type is sent in the creation of the shipment.

StefanRijnhart:
I'm a little surprised to see base_delivery_carrier_label being refactored out here, but I guess it had its flaws. We definitely made good use of the generic option model that @florian-dacosta mentions.

/cc @hbrunn

pedrobaeza:
@StefanRijnhart `base_delivery_carrier_label` was providing very few to the module in these latest versions, but means a high maintenance burden + dangerous features that are not needed and confuse the user (all the `delivery.carrier.option` thing).

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -0,0 +1,387 @@
...
+import requests
+
+from odoo import _, fields, models
+from odoo.exceptions import UserError
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = "delivery.carrier"
+
+    delivery_type = fields.Selection(selection_add=[("ups", "UPS")])
```
chienandalu, line 12: Should we try to avoid conflict with the enterprise module or do they exclude each other?

delivery_ups_oca/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "version": "13.0.1.0.0",
+    "development_status": "Production/Stable",
+    "category": "Delivery",
+    "website": "https://github.com/OCA/delivery-carrier",
+    "author": "Hunki Enterprises BV, Tecnativa, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": [
+        "delivery",
+        "delivery_package_number",
+        "delivery_price_method",
```
chienandalu, line 17: As this carrier has ratings implemented, this shouldn't be a requisite, isn't it?

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -0,0 +1,387 @@
...
+            "cancel": "%s/ship/v1/shipments/cancel" % base_url,
+            "status": "%s/track/v1/details/" % base_url,
+            "rate": "%s/ship/v1807/rating" % base_url,
+        }
+        return urls[ups_type]
+
+    def _ups_raise_error(self, api_result):
+        """If UPS' API returned an error, raise it"""
+        errors = api_result.get("response", {}).get("errors")
+        if self.debug_logging and errors:
```
chienandalu: Also use the native method to log requests calls/responses. It's called `log_xml`, although you can store any text and it opens another cursor to save your logged stuff in `ir.logging` records.

You can use it in your API calls with a try/except/finally

delivery_ups_oca/models/product_packaging.py
```diff
@@ -0,0 +1,37 @@
...
+            ("62", "Irregulars"),
+            ("63", "Parcel Post"),
+            ("64", "BPM Parcel"),
+            ("65", "Media Mail"),
+            ("66", "BPM Flat"),
+            ("67", "Standard Flat"),
+        ],
+        default="02",
+        string="UPS code",
+    )
```
chienandalu: I think the proper aproach is to create the proper `product.packaging` records for this carrier with their `shipper_package_code` and name. This way, it's possible to easily document each packaging dimensions and use them to compute volumetric stuff in the shippings. An example: https://github.com/OCA/delivery-carrier/blob/66bc5b6526cc8ebe8afe2bb42311b1b9f88a9853/delivery_schenker/data/delivery_schenker_data.xml

```suggestion
    package_carrier_type = fields.Selection(
        selection_add=[("ups", "UPS")],
    )
```

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -0,0 +1,387 @@
...
+            "AccessLicenseNumber": self.ups_access_license,
+            "Username": self.ups_ws_username,
+            "Password": self.ups_ws_password,
+            "transactionSrc": "Odoo (%s)" % self.env.cr.dbname,
+            "Content-Type": "application/json",
+            "Accept": "application/json",
+        }
+
+    def _ups_request(
+        self, url, payload=None, method="post", query_parameters=None,
```
chienandalu: I personally prefer to abstract the API stuff in a separate class with some exposed methods to separate that connection logic from the Odoo flows logic (this just an opinion :slightly_smiling_face: )

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -0,0 +1,387 @@
...
+        # Create attachment to add pdf label
+        if carrier.ups_file_format != "GIF":
+            attachment_name = "%s.pdf" % picking.carrier_tracking_ref
+        else:
+            attachment_name = "%s-%s.%s" % (
+                picking.carrier_tracking_ref,
+                carrier.ups_file_format,
+                carrier.ups_file_format.lower(),
+            )
+        self.env["ir.attachment"].create(
```
chienandalu: A nice to have feature is to give the user the oportunity to re-download the shipping labels if by chance someone deletes them.

chienandalu (changes requested):
Good work! Code review (I'll try to rebuild the runbot and test with the credentials in the test case)

delivery_ups_oca/models/delivery_carrier.py
victoralmau: Personally, I don't think anyone would install both addons at the same time (enterprise + OCA). Do you think we should change the type to `ups_oca`?

pedrobaeza: If it doesn't take too much, let's do it for whole compatibility. Nobody knows what can happen (maybe a transitional period enterprise-OCA where you need both modules and there are some carriers of one type and others of the other type?)

chienandalu: Well, it implies to rename all the wildcard methods anyway

pedrobaeza: @victoralmau you tell us, but if not renamed, declare the incompatibility in the manifest (I think it's `excludes` key).

chienandalu:
> I'm a little surprised to see base_delivery_carrier_label being refactored out here, but I guess it had its flaws. We definitely made good use of the generic option model that @florian-dacosta mentions.

I understand the common ground that `base_delivery_carrier_label` aims to provide but IMO those features should be added *on top* of the base carrier modules which should be kept as simple and stuck to the core methods as possible (which is not always completely possible).

florian-dacosta:
Well, that is kind of what I said, `base_delivery_carrier_label` could be split into multiple modules to really be simpler and have more optional modules depending on the need.

But there are some common ground usefull for every carrier and I think it is a shame to let it go.
Doing some generic carrier stuff in each carrier module is fine when one use only 1 or 2 carriers. But as soon as one use more, it can become a mess.

An example is the `carrier_account` model.
Adding login, password fields on delivery.carrier directly as in this module has several problems : 
- Duplicate fields for every carrier
- if for a same carrier, you use multiple delivery methods, you have to duplicate the configuration
- Totally inflexible => if you share delivery method between company but not the carrier accounts, you are blocked ?
- If you want to use your carrier account with preprod/test environement (which is usually handy) with module `server_environment`, you would just do one generic module (that already exists) instead of doing one more additional module per carrier...
- ...

Kind of problems that were resolved by a separated/generic model which is droped here out of the blue.

I understand that for a simple case, it is a bit more of maintenance, but it also make it kind of impossible to use for more complex cases.
I have such a case on my side and this kind of refactore make the module UPS impossible to use for me.

I don't see such a big maintenance burden of base_delivery_carrier_label, which is already refactored and migrated to 14.
And it could be even smaller splitting its functionality.
If there were a consensus, I could take the time to do such another recactore for v14 (maybe it is a bit late for v13?)

We could have something like this : 
`base_delivery_carrier_label`
- carrier_account and helper to get the right account from a delivery method (overridable for flexibility)
- shipping label : class inheriting ir_attachment for the carrier labels => very usefull to be able to find the labels attachment from the other picking attachments.
add a link with the package 
Could be usefull to : reprint label later without re-generating a new one from the carrier webservice
delete it if you cancel the tracking/label...
- helpers to avoid re-coding the same in all carrier modules and have some flexibility (overidable methods)
  - to create labels, attach it on the pickingand fill the racking number (instead of re-coding it in every carrier modules)
  - to get the address of the sender (overiddable) (in this ups module it is the warehouse address, but one could want the company instead, better to have this choice in a base module, once again, to be able to override it for all carriers and not carrier per carrier

`delivery_carrier_option`
Possibility to choose specific options on a picking for a specific delivery carrier.

`delivery_carrier_label_before_shipping` (we'd have to find another name...)
- make it possible to generate the labels before the actual transfer of the picking (when the picking is actually at ready state)
(If necessary, on my side, I don't think I use this feature but base_delivery_carrier_label add some stuff to be able to do it)

@StefanRijnhart @chienandalu @pedrobaeza

delivery_ups_oca/models/delivery_carrier.py
victoralmau: I have changed it to leave only one log message.

delivery_ups_oca/__manifest__.py
victoralmau: I'm sorry but I don't understand what dependency is missing. 
The thing about `xxxx_rate_shipment()` is something added by the `delivey` addon, right?

delivery_ups_oca/models/delivery_carrier.py
victoralmau: Although it is a good idea, I think this change (and other things) would need to be implemented in a base addon with some things:
- Hook auto-create label (attachment) + define new field `is_picking_label=True` (for example).
- Define a button in the picking view that allows to create the label again if there is no attachment with `is_picking_label` field.

pedrobaeza:
I agree there's common ground for deliveries, and that's why we created `delivery_state` and `delivery_package_number`, but for me, `carrier_account` it's not. Sharing 2 fields that not always fit the needed fields for authenticating is not a big maintenance burden and there's not too much penalty on having all of them in the same model.

`delivery_carrier_option` shouldn't exist as well IMO, as you are deforming the nature of the carriers. If you want several options, you create different delivery methods for the same carrier, but with these options. If not, you have to adapt all the flows, including the e-commerce one, to this 2-step mechanism, when the user doesn't benefit from selecting first `UPS` and then `express`, than selecting at first between `UPS` and `UPS (express)` in the list.

Anyways, please move this debate to an issue instead of contaminating this pull request.

delivery_ups_oca/models/delivery_carrier.py
pedrobaeza: How is this done in other modules?

chienandalu: In Odoo implementations they simply post a message in the picking thread attaching the label during the shipping validation. In other modules of our own, we do so as well and add this possibility in a carrier method *`whatever_carrier`*`_get_label`

pedrobaeza: Then let's do it for now that way.

victoralmau: Ok, already added.

pedrobaeza (approved):

chienandalu (approved):

PR #406, opened 2021-08-17
[UPD] Also update service level on stock.picking

lfreeke:


delivery_carrier_service_level/models/stock_picking.py
```diff
@@ -14,10 +14,16 @@ class StockPicking(models.Model):
...
+        compute="_compute_carrier_service_level",
+        inverse=lambda x: None,
     )
 
-    @api.onchange("carrier_id")
-    def on_change_carrier_id(self):
-        """Set carrier_service_level_id to False"""
-        for this in self:
-            this.carrier_service_level_id = False
+    def _compute_carrier_service_level(self):
```
NL66278: Compute methods should be above the field definition (convention), More important: they can not be ensure_one, add a `for this in self:` loop. And you will always need to set the value. Set self.carrier_service_level_id to False if no self.carrier_id.

ntsirintanis (approved):

delivery_carrier_service_level/models/stock_picking.py
```diff
@@ -10,14 +10,20 @@ class StockPicking(models.Model):
     _name = "stock.picking"
     _inherit = "stock.picking"
 
+    def _compute_carrier_service_level(self):
+        carrier_model = self.env["carrier.service.level"]
```
NL66278: This is misleading. It is not a carrier model, but a service_level_model, please rename accordingly.

delivery_carrier_service_level/models/stock_picking.py
```diff
@@ -10,14 +10,20 @@ class StockPicking(models.Model):
...
+        for this in self:
+            this.carrier_service_level_id = False
+            if not this.carrier_id:
+                continue
+            this.carrier_service_level_id = carrier_model.search(
```
NL66278: Directly under for this in self (indented of course):
```
this.carrier_service_level_id = service_level_model,search(
    [("carrier_id", "=", this.carrier_id.id)],
    limit=1,
) if this.carrier_id else service_level_model.browse([])
```
So assign empty record_set to this.carrier_service_level_id if there is no carrier.
Or maybe even better:
```
this.carrier_service_level_id = (
    service_level_model,search([("carrier_id", "=", this.carrier_id.id)], limit=1)
    if this.carrier_id
    else service_level_model.browse([])
)
```

NL66278:
Looks nice now! Does it work?

lfreeke:
@NL66278 Yes, we have tested it.

NL66278:
The pre-commit is failing because the code is not blacked. Install the pre-commit hook in your repository clone using the OCA instructions: https://odoo-community.org/blog/the-oca-blog-1/post/black-isort-pre-commit-97

If you commit and you get a message, just do $ git add . to include the changes created by the pre-commit hook and then try again.

lfreeke:
@NL66278 Somehow it doesn't fail for me locally when running pre-commit.

NL66278:
Note that the line length has been increased to 88 in the black configuration, adjust flake8, pylint and vin configurations accordingly.
In .vimrc: `autocmd FileType text setlocal textwidth=88 `
In .pylintrd: `max-line-length=88 `

NL66278:
Basically pre-commit wants the limit=1 argument on the same line as the domain.

NL66278 (approved):
:+1: LGTM

PR #409, opened 2021-08-25
[13.0][ADD] delivery_schenker: New module

chienandalu:
This module links the [DB Schenker](https://www.dbschenker.com) booking and tracking APIs with Odoo delivery system.

To configure a Schenker delivery method:

1. Go to *Inventory > Configuration > Delivery > Shipping methods* and create a new one.
2. Choose *DB Schenker* as provider.
3. Configure the service parameters according to your contract considerations.
4. Choose a delivery product and a default packaging. This is mandatory for the booking
   request as it needs the packaging code.

To make tests, set the carrier environment to test from the smart button. Don't forget
to set it to production once you're ready to use the delivery method!

These are the operations possible with this module:

Place shipping bookings
---

1. When the picking is validated, the shipping will be booked at Schenker.
2. With the response, we'll receive the delivery tracking number and the pdf label in a
   chatter message and it will be kept as attachment to the document.
3. You can manage packages number either with the proper Odoo workflows or with the
   package number field available in the *Additional Info* tab. You'll get as many
   labels as declared packages.

Cancel bookings
---

1. As in other carriers, we can cancel the shipping after the picking is done. To do
   so, go to *Additional Info* tab and click on the *Cancel* action on the side of the
   tracking number.
2. We can generate a new shipping if necessary.

Get labels
---

1. If by chance we delete the generated labels, we can obtain them again hitting the
   *Schenker Label* buttons in the header of the picking form.

Tracking
---

1. The module is integrated with `delivery_state` to be able to get the tracking info
   directly from the DB Schenker API.
2. To do so, go to a picking shipped with Schenker. In the *Additional Info* tab you'll
   find an action button to *Update tracking state* so the state will be updated from
   the Schenker API.

Debugging
---

The API calls and responses are tracked in two special fields in the picking that can
be viewed by technical users. You can also log them in as `ir.logging` records setting
the carrier debug on from the smart button.

cc @Tecnativa TT31005


delivery_schenker/data/delivery_schenker_data.xml
```diff
@@ -0,0 +1,120 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <!-- Schenker Product Packagings. It'd be nice to have proper measures -->
+    <record id="schenker_packaging_01" model="product.packaging">
+        <field name="shipper_package_code">CI</field>
+        <field name="name">Canister</field>
```
pedrobaeza: As we can have other module adding similar packagings, put at least in the name between parenthesis the carrier name for distinguishing them.

chienandalu: Done

pedrobaeza (approved):

delivery_schenker/models/delivery_carrier.py
```diff
@@ -0,0 +1,543 @@
...
+        """Tracking state update"""
+        # TODO: To be implemented
+        self.ensure_one()
+        # if not picking.carrier_tracking_ref:
+        #     return
+        # # schenker_request = SchenkerRequest(
+        # #     **self._get_schenker_credentials(), service="tracking"
+        # # )
+        # Get response
+        return
```
victoralmau: This is pending or cannot be done through the API?

delivery_schenker/models/stock_picking.py
```diff
@@ -0,0 +1,20 @@
...
+        help="Used for issues debugging",
+        copy=False,
+        readonly=True,
+    )
+    schenker_last_response = fields.Text(
+        string="Last Schenker xml response",
+        help="Used for issues debugging",
+        copy=False,
+        readonly=True,
+    )
```
victoralmau: Finally this is not necessary, right?

victoralmau (changes requested):
Good work, I think with a little more I would be ready to merge, right?

delivery_schenker/models/delivery_carrier.py
hildickethan: It can be done, you need to use endpoint `https://eschenker.dbschenker.com/webservice/trackingWebServiceV2` and service `getPublicShipmentDetails` with a request similar to this 
```xml
<ns3:getPublicServiceShipmentDetails
    xmlns:ns2="http://www.schenker.com/SGI/v4_0"
    xmlns:ns3="http://www.schenker.com/CustomerServices/eBusiness/ShipmentService/v2">
    <AccessKey>xxxxxxxxxxxxxxxxxxxxx</AccessKey>
    <in>
        <ns2:referenceType>cu</ns2:referenceType>
        <ns2:referenceNumber>yyyyyyyy</ns2:referenceNumber>
        <ns2:transportNature>exp</ns2:transportNature>
    </in>
</ns3:getPublicServiceShipmentDetails>
```

I have something ready to handle tracking for Schenker and am planning to integrate it into this module, but I'd rather wait until it is merged in order to migrate to 14.0 and add the tracking which we can then cherry-pick back to 13.0

hildickethan (approved):
With the removal of `schenker_last_request` and `schenker_last_response` fields since they aren't necessary with the `log_xml`, LGTM

delivery_schenker/models/delivery_carrier.py
chienandalu: Thanks @hildickethan ! Yes, we got the docs, but as Schenker doesn't work in the test enviroment with the test API (there's no demo data available to test the responses) we left it until having production data. No it's in a lower priority queue, but if you already have the proper implementation, we can merge this already and cherry pick as you say!

chienandalu:
@victoralmau @hildickethan changes done

victoralmau (approved):

chienandalu:
@hildickethan already merge, you can make your PR. Ping us for review when you got it ready :wink:

hildickethan:
@chienandalu Will do, conveniently I was looking at starting to work on it today

delivery_schenker/views/stock_picking_views.xml
```diff
@@ -0,0 +1,20 @@
...
+                <button
+                    name="schenker_get_label"
+                    string="Schenker Label"
+                    type="object"
+                    attrs="{'invisible':['|', '|', ('delivery_type', '!=', 'schenker'), ('carrier_tracking_ref', '=', False), ('state', '!=', 'done')]}"
+                />
+            </xpath>
+        </field>
+    </record>
+</odoo>
```
hildickethan, line 20: @chienandalu This slipped through due to changes on an existing DB I imagine, the `schenker_get_label` function doesn't exist on `stock.picking` and it stops the module from even installing. We would need an extra function on `stock.picking` for this which calls the same function on `delivery.carrier` and adds the PDF to attachments or posts it in the chatter

chienandalu: Sure. That was meant to work with the tracking features. It should go away until they're implemented. I'll patch it tomorrow

chienandalu: Fix here: https://github.com/OCA/delivery-carrier/pull/447

PR #412, opened 2021-08-30
[ADD][13.0] delivery_extra_fee delivery_carrier_by_line

i-vyshnevska:
add two new addons to extend delivery logic

### DIFF
--- a/delivery_carrier_by_line/__manifest__.py
+++ b/delivery_carrier_by_line/__manifest__.py
@@ -1,7 +1,7 @@
 # Copyright 2021 Camptocamp
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 {
-    "name": "Delivery by per Carrier",
+    "name": "Delivery carrier by sales order line",
     "summary": "",
     "version": "13.0.1.0.0",
     "category": "Delivery",
--- a/delivery_carrier_by_line/__manifest__.py
+++ b/delivery_carrier_by_line/__manifest__.py
@@ -1,0 +1,14 @@
+# Copyright 2021 Camptocamp
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Delivery by per Carrier",
+    "summary": "",
+    "version": "13.0.1.0.0",
+    "category": "Delivery",
+    "website": "https://github.com/OCA/delivery-carrier",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "installable": True,
+    "license": "AGPL-3",
+    "depends": ["delivery", "sale", "stock_move_assign_picking_hook"],
+    "data": ["views/sale_order.xml"],
+}
--- a/delivery_carrier_by_line/models/stock_move.py
+++ b/delivery_carrier_by_line/models/stock_move.py
@@ -1,0 +1,30 @@
+# Copyright 2021 Camptocamp (https://www.camptocamp.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+
+from odoo import models
+
+_logger = logging.getLogger(__name__)
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _domain_search_picking_for_assignation_default(self):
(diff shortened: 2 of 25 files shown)


delivery_carrier_by_line/models/stock_move.py
```diff
@@ -0,0 +1,32 @@
...
+    _inherit = "stock.move"
+
+    def _domain_search_picking_for_assignation_default(self):
+        domain = super()._domain_search_picking_for_assignation_default()
+        domain.append(("carrier_id", "=", self.sale_line_id.carrier_id.id))
+        return domain
+
+    def _get_new_picking_values(self):
+        pass
+        # pdb.set_trace()
```
ajaniszewska-dev:
```suggestion
```

delivery_carrier_by_line/tests/test_delivery_carrier_by_line.py
```diff
@@ -0,0 +1,108 @@
+# Copyright 2020 Camptocamp
```
ajaniszewska-dev:
```suggestion
# Copyright 2021 Camptocamp
```

delivery_carrier_by_line/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2021 Camptocamp
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Delivery by per Carrier",
```
jbaudoux:
```suggestion
    "name": "Delivery carrier by sales order line",
```

delivery_carrier_by_line/models/stock_move.py
```diff
@@ -0,0 +1,30 @@
...
+    def _domain_search_picking_for_assignation_default(self):
+        domain = super()._domain_search_picking_for_assignation_default()
+        domain.append(("carrier_id", "=", self.sale_line_id.carrier_id.id))
+        return domain
+
+    def _get_new_picking_values(self):
+        vals = super()._get_new_picking_values()
+        carrier = self.mapped("sale_line_id.carrier_id")
+        if len(carrier) > 1:
+            _logger.warning("Moves belongs to different carriers %s" % carrier)
```
jbaudoux: According to me, the call to super will already fail on
`vals['carrier_id'] = self.mapped('sale_line_id.order_id.carrier_id').id`

I would move that test before the call to super and raise an error instead of logging a warning.
The else case is managed by the call to super and can be dropped

delivery_extra_fee/models/sale_order_line.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2021 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    is_delivery_extra = fields.Boolean(related="product_id.is_delivery_extra")
```
jbaudoux, line 10: add readonly=True

jbaudoux (commented):
LGTM. Just some small remarks

delivery_extra_fee/models/sale_order_line.py
i-vyshnevska: it is readonly by default, all computed and respectively related fields are readonly

delivery_carrier_by_line/models/stock_move.py
i-vyshnevska: the parent method check for carrier on order, here I am checking for carrier on order lines, these two checks are not equal

delivery_extra_fee/models/sale_order_line.py
jbaudoux: I don't think so for related

i-vyshnevska: it is, as related is computed by itself, for instance, in the settings all related fields are made explicitly related=False  
https://github.com/odoo/odoo/blob/c236087d126052415e62bd3a725d879d21732c96/addons/account/models/res_config_settings.py#L11

PR #413, opened 2021-09-02
[ADD] delivery_carrier_label_quickpac

StephaneMangin:
This is a new implementation for delivery carrier interaction with Quickpac.

This module has been based on the postlogistics. It uses an external python library that impleements the Quickpac API through its OpenAPI client generation (Swagger Codegen, see: https://pypi.org/project/quickpac/)


pedrobaeza:
Is Quickpac an international carrier or only a local one? If the second one, then this should be instead on the l10n-x repository IMO.

StephaneMangin:
> Is Quickpac an international carrier or only a local one? If the second one, then this should be instead on the l10n-x repository IMO.

I see the though and it's logical, but let's see what everyone think about it because it's a delivery behavior and this repo is intended for.

pedrobaeza:
Yeah, but the repo PSC/maintainers, if they are not from your country, won't be able to act in this module because the local things. We have done the same with other local providers like https://github.com/OCA/l10n-spain/tree/13.0/delivery_gls_asm for letting Spanish maintainers to be able to act on them.

StephaneMangin:
> Yeah, but the repo PSC/maintainers, if they are not from your country, won't be able to act in this module because the local things. We have done the same with other local providers like https://github.com/OCA/l10n-spain/tree/13.0/delivery_gls_asm for letting Spanish maintainers to be able to act on them.

Oh ok, didn't knew that. I'll check this out then. Thanks

StephaneMangin:
Canceled in favor of https://github.com/OCA/l10n-switzerland/pull/613

yvaucher:
@pedrobaeza this module is very similar to the postlogistics one which is also only for Switzerland. I would have prefered to have them grouped together.

pedrobaeza:
But both should be in that repo, not here.

yvaucher:
I'm a bit mixed about this. localization repo are usually more focused on Accounting, bank interconnection. IMO, adding delivery is not in the current scope of the PSC.

As an other example it would be difficult for me to move the module https://github.com/OCA/geospatial/tree/13.0/geoengine_swisstopo to https://github.com/OCA/l10n-switzerland as it is strongly linked to other modules in the same repository. And I have few people in OCA/l10n-switzerland that know a thing about Openlayers.

pedrobaeza:
Well, such focus you mention is the traditional one, as normally there weren't other features specific to the country, but imagine to populate here a bunch of carriers (in Spain we have now 4 national carriers, and increasing), and transfer to the PSCs of this repository the responsibility of handling all of them, issues reports and so. Having them also on the localization repo, allows to use your native language to report problems in the issue tracker (or put the README in such language), be closer to the people that can know how they work, etc, so for me it's a better option.

PR #414, opened 2021-09-06
[FIX] delivery_price_method: method permissions

chienandalu:
If a user don't have the necessary permissions for writing in the
`delivery.carrier` model (like a low range salesman) he won't be able to
choose a carrier wich uses the rate shipment override.

cc @Tecnativa TT31627

### DIFF
--- a/delivery_price_method/models/delivery_carrier.py
+++ b/delivery_price_method/models/delivery_carrier.py
@@ -24,10 +24,10 @@ class DeliveryCarrier(models.Model):
         previous_method = False
         if self.price_method in ("fixed", "base_on_rule"):
             previous_method = self.delivery_type
-            self.delivery_type = self.price_method
+            self.sudo().delivery_type = self.price_method
         res = super().rate_shipment(order)
         if previous_method:
-            self.delivery_type = previous_method
+            self.sudo().delivery_type = previous_method
         return res
 
     def send_shipping(self, pickings):


delivery_price_method/models/delivery_carrier.py
```diff
@@ -24,10 +24,10 @@ def rate_shipment(self, order):
         previous_method = False
         if self.price_method in ("fixed", "base_on_rule"):
             previous_method = self.delivery_type
-            self.delivery_type = self.price_method
+            self.sudo().delivery_type = self.price_method
```
victoralmau, line 27:
```suggestion
            _self = self
            if not _self.check_access_rights("write"):
                _self = _self.sudo()
           _self.delivery_type = self.price_method
```

delivery_price_method/models/delivery_carrier.py
```diff
@@ -24,10 +24,10 @@ def rate_shipment(self, order):
...
         res = super().rate_shipment(order)
         if previous_method:
-            self.delivery_type = previous_method
+            self.sudo().delivery_type = previous_method
```
victoralmau, line 30:
```suggestion
            _self.delivery_type = previous_method
```

victoralmau (changes requested):

pedrobaeza (approved):

delivery_price_method/models/delivery_carrier.py
pedrobaeza: Although it can avoid sudo in v12 where the uid is changed, I don't think in this specific case it's going to add too much.

victoralmau (approved):

pedrobaeza:
Let's fast-track the patch:

/ocabot merge patch

I suppose this is needed in v13/v14 as well.

chienandalu:
Can you try to merge it again?

pedrobaeza:
@chienandalu is this needed in v13?

chienandalu:
Yes, I just tested it

PR #415, opened 2021-09-16
[13.0][ADD] delivery_local_pickup: Adds a local pickup carrier

victoralmau:
Adds a local pickup carrier.

@Tecnativa TT30665


delivery_local_pickup/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Delivery Local pickup",
+    "version": "13.0.1.0.0",
+    "category": "Delivery",
+    "website": "https://github.com/OCA/delivery-carrier",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": ["delivery", "delivery_state", "website"],
```
chienandalu: What's `delivery_state` needed for?

delivery_local_pickup/models/delivery_carrier.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def local_pickup_get_tracking_link(self, picking):
+        config_parameter = self.env["ir.config_parameter"].sudo()
+        base_url = config_parameter.get_param(
+            "web.base.url", default="http://localhost:8069"
+        )
+        return "%s/local_pickup/%s" % (base_url, picking.id)
+
+    def local_pickup_rate_shipment(self, order):
+        raise NotImplementedError(_("Local Pickup not allow compute delivery rates"))
```
chienandalu: I'd rather return 0 and state in the readme that `delivery_price_method` can be used in order to charge for this carrier type.

delivery_local_pickup/models/product_packaging.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ProductPackaging(models.Model):
+    _inherit = "product.packaging"
+
+    package_carrier_type = fields.Selection(
```
chienandalu: I'm not sure this is very useful for this particular case

delivery_local_pickup/models/delivery_carrier.py
```diff
@@ -0,0 +1,28 @@
...
+    def local_pickup_send_shipping(self, pickings):
+        return [self.local_pickup_create_shipping(p) for p in pickings]
+
+    def local_pickup_create_shipping(self, picking):
+        return {"exact_price": False, "tracking_number": _("Local pickup")}
+
+    def local_pickup_get_tracking_link(self, picking):
+        config_parameter = self.env["ir.config_parameter"].sudo()
+        base_url = config_parameter.get_param(
+            "web.base.url", default="http://localhost:8069"
```
chienandalu: This is not necessarily true if the website doesn't have the same url, like for example multi company/multi website. Maybe a proper way would be to add a website_id field or at least default to the proper company website url. This could be added to the ROADMAP for now anyway

delivery_local_pickup/views/website_templates.xml
```diff
@@ -0,0 +1,29 @@
...
+                        <p><strong>Picking:</strong> <span t-esc="picking.name" /></p>
+                        <p t-if="picking.origin">
+                            <strong>Origin:</strong> <span t-esc="picking.origin" />
+                        </p>
+                    </div>
+                    <div id="information_right" class="col-6">
+                        <div class="address">
+                            <strong>Local Address:</strong>
+                            <div
+                                t-esc="picking.carrier_id.local_address_id"
```
chienandalu, line 20: Wouldn't it be convinient to render the carrier `website_description` as well if any?

chienandalu (commented):
Code review

delivery_local_pickup/__manifest__.py
victoralmau: This is necessary to use the functions it adds and to be able to define the `carrier_tracking_ref` value for picking and that information appears in the confirmation email.

delivery_local_pickup/models/delivery_carrier.py
victoralmau: I would personally remove the function, but I have seen `delivery_seur` add something similar.

chienandalu: It can't be removed due to the core design that mandates that every carrier has this methods defined.

delivery_local_pickup/__manifest__.py
chienandalu: I'd say that can be done in core without further extensions

delivery_local_pickup/models/delivery_carrier.py
victoralmau: I have changed this a bit to use the carrier's website if it is defined.

pedrobaeza: I don't think you should build the URL this way, but there should be a method that builds it. Take into account that the `website_id` field is only available when you install `website_delivery`, so we can't depend on it.

victoralmau: Ok, I had added the `website_delivery` dependency to use the `website_id` field and also for this https://github.com/OCA/delivery-carrier/pull/415#discussion_r710059907 but if you want better we can remove it.

Ok, I will try to find some odoo core method to build the URL although if you know any it would be great.

pedrobaeza: Please don't. If any, we will need to build a `website_delivery_local_pickup` glue module for that, that returns a different URL if a website is put. I don't know which is the method for the base to return the current URL, but `mail` uses for building the URL for the buttons `"View <object>"` that we see on the generated mails, so please search a bit.

victoralmau: Odoo does it this way: https://github.com/odoo/odoo/blob/13.0/addons/mail/models/mail_thread.py#L455
It is already changed to use something similar.

chienandalu: Yes, that's why I said it could be put on the roadmap. Anyway, we could check that the attribute exists and behave accordinly if so to avoid the hard dependency or a glue module. `website_sale_delivery` is auto installed with this module's dependencies after all: https://github.com/odoo/odoo/blob/13.0/addons/website_sale_delivery/__manifest__.py#L22

victoralmau: I think this function would be enough.

In the future, another addon `website_delivery_local_pickup` (delivery_local_pickup + website_sale_delivery) would implement the function to return the URL according to the `website_id` field and would add the `website_description` field of the carrier to the website.

pedrobaeza: Well, I like the idea of checking `if "website_id" in self._fields` for not having to have a glue module.

victoralmau:
Changes done.

delivery_local_pickup/models/delivery_carrier.py
```diff
@@ -0,0 +1,40 @@
...
+    delivery_type = fields.Selection(selection_add=[("local_pickup", "Local pickup")])
+    local_address_id = fields.Many2one(
+        comodel_name="res.partner", string="Local address"
+    )
+
+    def local_pickup_send_shipping(self, pickings):
+        return [self.local_pickup_create_shipping(p) for p in pickings]
+
+    def local_pickup_create_shipping(self, picking):
+        return {"exact_price": False, "tracking_number": _("Local pickup")}
```
chienandalu: Testing on runbot the other day, the mail stated something like "your trackink numbre is Local pickup...". Maybe it would be more coherent to simply put the picking name.

victoralmau: I think it is important that this text appears because it is what is shown in the email that is sent to the client when the stock picking is done.

chienandalu: Ok, but the resultant message is kinda weird. Anyway I don't come up with any clever string substitute right now :slightly_smiling_face: 

Selección_01064
:

victoralmau: Ok yeah, it's kind of weird but there is no other option in my opinion.
Can you check the PR again to approve it if it is OK?

chienandalu: I'd go for a picking name plus carrier name. For example, for the picking *WH/OUT/0001* and the carrier name *Store Pickup* we would get something like: "*Your tracking reference is [**WH/OUT/0001 (Store Pickup)**](#)*"

Which way more informative, don't you think?

victoralmau: Thanks, changes done.

chienandalu (approved):

delivery_local_pickup/demo/assets.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- License AGPL-3.0 (http://www.gnu.org/licenses/agpl.html). -->
+<odoo>
+    <template id="delivery_local_pickup_frontend_demo" inherit_id="web.assets_frontend">
```
CarlosRoca13: You can charge it at `web.assets_tests`
```suggestion
    <template id="delivery_local_pickup_frontend_demo" inherit_id="web.assets_tests">
```

delivery_local_pickup/tests/test_website.py
```diff
@@ -0,0 +1,21 @@
...
+    def test_local_pickup_tour(self):
+        self.picking.action_confirm()
+        self.picking.action_assign()
+        self.picking.send_to_shipper()
+        tour = "delivery_local_pickup_tour"
+        self.phantom_js(
+            "/local_pickup/%s" % self.picking.id,
+            "odoo.__DEBUG__.services['web_tour.tour'].run('%s')" % tour,
+            "odoo.__DEBUG__.services['web_tour.tour'].tours.%s.ready" % tour,
+        )
```
CarlosRoca13: In 13.0 exists this method that it's more readable. You can set too the user that runs the tour by setting `login="admin"` for example
```suggestion
        self.start_tour("/local_pickup/%s" % self.picking.id, "delivery_local_pickup_tour")
```

victoralmau:
Thanks @CarlosRoca13, changes done.

CarlosRoca13 (approved):

PR #416, opened 2021-09-17
[12.0][IMP] delivery_carrier_label_paazl: removed description sent to carrier

dkubiak789:


delivery_carrier_label_paazl/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Carrier labels for paazl",
     "summary": "Print carrier labels for paazl",
-    "version": "12.0.1.1.2",
+    "version": "12.0.1.1.3",
```
astirpe: No need to change the version, since it will be automatically updated when merged.

delivery_carrier_label_paazl/models/stock_picking.py
```diff
@@ -214,7 +214,6 @@ def _paazl_commit_order_data(self):
                     product["quantity"] * product["weight"]
                     for product in self._paazl_order_data()["products"]["product"]
                 ),
-                description=self.note or "/",
```
astirpe, line 217: Can you motivate this change for the people who already use this module and will need to upgrade it?

astirpe (changes requested):

delivery_carrier_label_paazl/models/stock_picking.py
dkubiak789: What do you mean?

astirpe: I mean that the people who will watch this PR will not understand why this change was needed.

dkubiak789:
Note field on delivery order may contain some internal customer-related notes. We don't want to send this information to Paazl.

delivery_carrier_label_paazl/models/stock_picking.py
dkubiak789: sure

astirpe (approved):

StefanRijnhart:
Can you squash the two commits into one?

dkubiak789:
> Can you squash the two commits into one?

Sure.

StefanRijnhart (approved):
Agreed with the proposed change. No other carrier implementations in this repo send the picking note to the carrier.

PR #418, opened 2021-09-22
[FIX] delivery_carrier_label_ups: UPS returns labels creation

astirpe:
This PR fixes the error occurring when creating labels for a UPS return: when the `ReturnService` code is `'8'` (UPS Electronic Return Label), some extra data in payload become mandatory. These data include: `LabelDelivery`, `LabelSpecification` and `Notification`.
This PR also implements a split of method, to ease the customization of the `ReturnService` payload.

### DIFF
--- a/delivery_carrier_label_ups/models/stock_picking.py
+++ b/delivery_carrier_label_ups/models/stock_picking.py
@@ -138,11 +138,8 @@ class StockPicking(models.Model):
             shipping_data.update(
                 ShipTo=shipping_data.pop('ShipFrom'),
                 ShipFrom=shipping_data.pop('ShipTo'),
-                ReturnService=dict(
-                    # UPS Electronic Return Label (ERL)
-                    Code='8',
-                ),
             )
+            shipping_data.update(self._ups_return_service())
         return dict(
             ShipmentRequest=dict(
                 Shipment=dict(
@@ -167,7 +164,67 @@ class StockPicking(models.Model):
                     ],
                     **shipping_data,
                 ),
+                LabelSpecification=self._ups_label_specification(account),
             )
+        )
+
+    def _ups_return_service(self):
+        """Return a dict describing the UPS Return Label code"""
+        self.ensure_one()
+        customer_email = self.partner_id.email
+        warehouse_partner = self.picking_type_id.warehouse_id.partner_id
+        warehouse_partner_email = warehouse_partner.email
+        if self.partner_id.country_id.code != 'US':
+            subject_code = '01'  # First shipment reference
+        else:
+            subject_code = '03'  # First package reference
+
+        return dict(
+            ReturnService=dict(
+                # UPS Electronic Return Label (ERL)
+                Code='8',
+            ),
+            ShipmentServiceOptions=dict(
+                LabelDelivery=dict(
+                    EMailMessage=dict(
+                        EMailAddress=customer_email,
+                        FromEMailAddress=warehouse_partner_email,
+                        SubjectCode=subject_code,
+                    )
+                ),
+                Notification=[
+                    dict(
+                        NotificationCode='2',
+                        EMail=dict(
+                            EMailAddress=customer_email,
+                            UndeliverableEMailAddress=warehouse_partner_email,
+                            FromEMailAddress=warehouse_partner_email,
+                        ),
+                    ),
+                    dict(
+                        NotificationCode='5',
+                        EMail=dict(
+                            EMailAddress=customer_email,
+                            FromEMailAddress=warehouse_partner_email,
+                        ),
+                    ),
+                    dict(
+                        NotificationCode='7',
+                        EMail=dict(
+                            EMailAddress=customer_email,
+                            FromEMailAddress=warehouse_partner_email,
+                        ),
+                    ),
+                    dict(
+                        NotificationCode='8',
+                        EMail=dict(
+                            EMailAddress=customer_email,
+                            UndeliverableEMailAddress=warehouse_partner_email,
+                            FromEMailAddress=warehouse_partner_email,
+                        ),
+                    )
+                ]
+            ),
         )
 
     def _ups_partner_to_shipping_data(self, partner, **kwargs):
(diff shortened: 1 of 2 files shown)


delivery_carrier_label_ups/models/stock_picking.py
```diff
@@ -170,6 +170,14 @@ def _ups_shipping_data(self):
             )
         )
 
+    def _ups_return_service(self):
+        """Return a dict describing the UPS Return Label code"""
+        self.ensure_one()
+        return dict(
+            # UPS Electronic Return Label (ERL)
+            Code='8',
```
StefanRijnhart: The return service type was implemented by our team for our customer. If we can't get it to work for us, it will not work for anybody and it has to be reverted as a whole.

astirpe: @StefanRijnhart indeed `Code='8`' seems requiring extra info, like `ShipmentServiceOptions`, which is not implemented in this module. Without that extra info, the UPS request for a return shipment will be always failing. Better to remove `Code='8'` from here for now and see if later on someone will improve this part by adding extra logic. What do you think?

StefanRijnhart: If we need it anyway for our project, do you see a chance to implement it here rather than in our customizations?

astirpe: Yes, I will try to implement it here. I added https://github.com/OCA/delivery-carrier/pull/418/commits/9c5e14e8acce70e9e3ff2c3d37cf4bd3c5b96748 which is still a work in progress as I don't have yet a full understanding of the UPS specifications.

astirpe: I put this PR to ready for review. I will update it in case something will come out from the tests at the customer's side.

hbrunn: weird thing is this worked in the test environment of UPS...

astirpe: @hbrunn I don' know why. Did they maybe change something in UPS api? Or is it maybe a matter of configuration on UPS customer side?

hbrunn: thankfully I've forgotten most of the super unwieldy hard to search non-indexed PDFs that UPS calls the documentation for their API, so I happily leave this to you :-)

pankk (approved):
Code review

delivery_carrier_label_ups/models/stock_picking.py
```diff
@@ -167,9 +164,68 @@ def _ups_shipping_data(self):
...
+        return dict(
+            ReturnService=dict(
+                # UPS Electronic Return Label (ERL)
+                Code='8',
+            ),
+            ShipmentServiceOptions=dict(
+                LabelDelivery=dict(
+                    EMailMessage=dict(
+                        EMailAddress=email,
+                        FromEMailAddress=email,
```
StefanRijnhart: Are you sure the From email address should be set to the same email address as the recipient (i.e. the customer address in our case?)

delivery_carrier_label_ups/models/stock_picking.py
```diff
@@ -167,9 +164,68 @@ def _ups_shipping_data(self):
...
+                        SubjectCode=subject_code,
+                    )
+                ),
+                Notification=[
+                    dict(
+                        NotificationCode='2',
+                        EMail=dict(
+                            EMailAddress=email,
+                            UndeliverableEMailAddress=email,
```
StefanRijnhart: Sorry if I'm being redundant, but the undeliverable email address should also be set to the sender email if you ask me.

astirpe:
@StefanRijnhart code updated, would you give it a look?

StefanRijnhart:
Thanks, looking good now! I'll wait with my approval until this code is accepted by our users.

StefanRijnhart:
Can you squash commits and update the PR title to indicate you actually fixed the implementation of UPS returns?

pankk (approved):

hbrunn (approved):

delivery_carrier_label_ups/models/stock_picking.py
```diff
@@ -167,9 +164,69 @@ def _ups_shipping_data(self):
...
+                LabelSpecification=self._ups_label_specification(account),
             )
         )
 
+    def _ups_return_service(self):
+        """Return a dict describing the UPS Return Label code"""
+        self.ensure_one()
+        recipient = self.picking_type_id.warehouse_id.partner_id
+        recipient_email = recipient.email
+        shipper_email = self.partner_id.email
```
StefanRijnhart: Wait, is the shipper_email set to the customer email? And the recipient email to the company's warehouse email? Isn't that the wrong way around?

hbrunn: now I'm confused. @astirpe could you clarify? We're in danger of merging...

astirpe: This is a return, isn't the recipient the company's warehouse? And the shipper the customer?

StefanRijnhart: There is something to say for that in the case of returns but in any case you have to make sure that FromEmailAddress below is set to the company email address and the EmailAddress to the customer email address, which is not the case now.

StefanRijnhart: And in our 8.0 code, the subject_code was dependent on the customer country code, not the company warehouse country code. That is now swapped around too.
In our 8.0 code, `recipient` always referred to the customer, and `shipper` always referred to the warehouse. If this is causing confusion, maybe we should avoid these terms and always use variable names `customer` and `warehouse_partner`.

astirpe: @StefanRijnhart you are right! I was convinced they were inverted. Thanks!

StefanRijnhart (approved):

delivery_carrier_label_ups/models/stock_picking.py
StefanRijnhart: Alright, thanks for the update!

astirpe:
Is it ok to merge this one?

StefanRijnhart:
Yes!
/ocabot merge minor

PR #421, opened 2021-09-28
[13.0][FIX] delivery_ups_oca: Prevent error when convert price in different currency

victoralmau:
Prevent error when convert price in different currency (We need to convert the price if the currency is different).
It is necessary to do the currency conversion in the `delivery.carrier` model because the `UpsRequest` class does not have access to `self.env`

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT30631

### DIFF
--- a/delivery_ups_oca/tests/test_delivery_ups.py
+++ b/delivery_ups_oca/tests/test_delivery_ups.py
@@ -85,6 +85,16 @@ class DeliveryUps(common.SavepointCase):
         self.assertGreater(res["price"], 0)
         self.assertTrue(res["success"])
 
+    def test_order_ups_rate_shipment_currency_extra(self):
+        usd = self.env.ref("base.USD")
+        eur = self.env.ref("base.EUR")
+        currency = self.env.ref("base.main_company").currency_id
+        currency_extra = eur if currency == usd else usd
+        self.sale.currency_id = currency_extra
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Steps to reproduce the problem? I'm a bit lost now that you have removed the conversion in some places, and not the full response is passed through different methods.

victoralmau:
I have updated the description to explain the use and need for this change.

delivery_ups_oca/tests/test_delivery_ups.py
```diff
@@ -81,6 +81,11 @@ def _create_sale_order(self):
         return sale
 
     def test_order_ups_rate_shipment(self):
+        usd = self.env.ref("base.USD")
+        eur = self.env.ref("base.EUR")
+        currency = self.env.ref("base.main_company").currency_id
+        currency_extra = eur if currency == usd else usd
+        self.sale.currency_id = currency_extra
```
joao-p-marques, line 93: I think it would be better to create another test for this specific case instead of re-using this existing one. That way, it will be easier to avoid regression changes...

joao-p-marques (changes requested):
Apart from what @pedrobaeza already asked, if this change is actually necessary, the code LGTM. Just a comment regarding the tests...

joao-p-marques (approved):

pedrobaeza (approved):

PR #426, opened 2021-10-14
[14.0][ADD] partner_default_delivery_carrier

SilvioC2C:
New module that allows defining a default delivery carrier for new partners in a multi-company environment.


pedrobaeza:
Why not using standard `ir.defaults` feature for doing this?

SilvioC2C:
@pedrobaeza because the standard behavior when setting a `default_*` field in `res.config.settings` ignores the company variable: https://github.com/odoo/odoo/blob/14.0/odoo/addons/base/models/res_config.py#L560-L569
Instead, the module allows defining different default carriers for different companies.

pedrobaeza:
I'm not talking about settings shortcut, but real user-defined values, that allows to be set by company:

Selección_046

Thus, you don't need any extra module, just DB configuration.

SilvioC2C:
@pedrobaeza in most environments, users do not have access to the technical features, while sales manager do have access to sales configuration. This allows sales manager to define multi-company default values without having to access the technical features menu to do it.

pedrobaeza:
OK, no problem if you prefer to go on. I always try to avoid maintenance burden in my installations for such simple configuration tasks that are done once or two.

ivantodorovich (commented):
I agree with @pedrobaeza that this should use `ir.default`s for the default value.

IMO this module could rely on `ir.default` for the implementation, while still providing the same UX through `res.config.settings`. You just have to make the `res.config.setting` field read and write (or create) the proper `ir.default` record for that company. Should be straightforward with `compute` and `inverse`

SilvioC2C:
Hello @pedrobaeza @ivantodorovich 
I've updated the module using `ir.default` records to store the default values as you suggested.

ivantodorovich (approved):

ajaniszewska-dev (approved):

PR #429, opened 2021-10-21
[14.0][FIX] delivery_package_fee: fix copy_data

sebalix:
`values["order_line"]` could be set to `False`

This will makes `shopinvader` compatible with the current module: https://github.com/shopinvader/odoo-shopinvader/blob/38d1768acf925ee82d9242297e1b32de921bc4ca/shopinvader/services/cart.py#L121

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -96,7 +96,7 @@ repos:
       - id: pyupgrade
         args: ["--keep-percent-format"]
   - repo: https://github.com/PyCQA/isort
-    rev: 5.5.1
+    rev: 5.9.3
     hooks:
       - id: isort
         name: isort except __init__.py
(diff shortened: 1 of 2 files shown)


sebalix:
The release of Python 3.10 has broken the isort pre-commit, I updated it in the second commit.

simahawk:
sounds like the fix done on v13 https://github.com/OCA/delivery-carrier/commit/826692bf1a322cd5dff354e0d9c333c292811b60

.pre-commit-config.yaml
```diff
@@ -96,7 +96,7 @@ repos:
       - id: pyupgrade
         args: ["--keep-percent-format"]
   - repo: https://github.com/PyCQA/isort
-    rev: 5.5.1
```
simahawk, line 99: shouldn't go to https://github.com/OCA/oca-addons-repo-template ?

sebalix: Yes for sure

sebalix:
@simahawk thanks, I'll rework the PR to include your fix instead

.pre-commit-config.yaml
sebalix: Already fixed there.

sebalix:
Closing in favor of:
- #431

PR #430, opened 2021-10-22
[13.0][REF] delivery_ups_oca: Code refactor.

victoralmau:
Code refactor (standardize code according to other addons).

Please @chienandalu and @pedrobaeza can you review it?

@Tecnativa

### DIFF
--- a/delivery_ups_oca/models/delivery_carrier.py
+++ b/delivery_ups_oca/models/delivery_carrier.py
@@ -77,24 +77,6 @@ class DeliveryCarrier(models.Model):
         "webservice (you will necessary to activate tracking API)",
     )
 
-    def _get_ups_params(self, service):
-        return {
-            "prod": self.prod_environment,
-            "service": service,
-            "params": {
-                "access_license_number": self.ups_access_license,
-                "username": self.ups_ws_username,
-                "password": self.ups_ws_password,
-                "default_packaging_id": self.ups_default_packaging_id,
-                "shipper_number": self.ups_shipper_number,
-                "service_code": self.ups_service_code,
-                "file_format": self.ups_file_format,
-                "package_dimension_code": self.ups_package_dimension_code,
-                "package_weight_code": self.ups_package_weight_code,
-                "transaction_src": "Odoo (%s)" % self.env.cr.dbname,
-            },
-        }
-
     def ups_send_shipping(self, pickings):
         return [self.ups_create_shipping(p) for p in pickings]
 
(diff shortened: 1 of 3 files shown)


delivery_ups_oca/views/delivery_carrier_view.xml
```diff
@@ -65,6 +65,18 @@ License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
                         </group>
                     </group>
                 </page>
+                <page
+                    string="Technical"
+                    attrs="{'invisible': [('delivery_type', '!=', 'ups')]}"
+                    name="technical_ups_oca"
```
pedrobaeza: If technical, then put `group_no_one` in the page. And also don't call it `Technical`, but `UPS Technical` or similar for identifying that is that. Should this go anyways in a new page?

victoralmau: Changes done.

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -76,24 +76,12 @@ class DeliveryCarrier(models.Model):
...
-                "file_format": self.ups_file_format,
-                "package_dimension_code": self.ups_package_dimension_code,
-                "package_weight_code": self.ups_package_weight_code,
-                "transaction_src": "Odoo (%s)" % self.env.cr.dbname,
-            },
-        }
+    ups_last_request = fields.Text(
+        string="Last UPS API request", help="Used for debugging", readonly=True
+    )
+    ups_last_response = fields.Text(
```
pedrobaeza: I'm thinking about the usefulness of this, as not having bound to any operation, or if several of them are performed in batch, having the last one is not very useful at all, don't you think? @chienandalu how do you see it?

victoralmau: This is an interesting question.
There are currently different options:
- `delivery_seur` The request information is saved in the `stock.picking` record.
- `delivery_dhl_parcel` The request information is saved in the `delivery.carrier` record.

What do you think should be the correct option?

chienandalu: I'd go for the standard core method which handles a parallel cursor in case of error, can be activated/deactivated with the proper UI button (as mentioned in other PR) and avoids aditional fields: https://github.com/OCA/delivery-carrier/blob/bb074e0da12737ad0fcce85357b10d15ef3fe50d/delivery_schenker/models/delivery_carrier.py#L186-L187

victoralmau: Ok thanks, it is changed based on this suggestion.
I think it will be a good idea to also change it in `delivery_seur` and `delivery_dhl_parcel` so that all carrier addons are "similar".

pedrobaeza (approved):

chienandalu (approved):

PR #434, opened 2021-10-28
[14.0][ADD] delivery_correos_express: Correos Express integration

hildickethan:
Took inspiration from https://github.com/OCA/delivery-carrier/pull/196, but since that PR has been WIP for 2 years and it uses the old methodology with `base_delivery_carrier_label` that has been refactored out of recent modules, I made this one based on the other more recent delivery modules in l10n-spain.

I'm creating this PR here because in that linked PR it was stated it's valid for Spain and Portugal so they created theirs here in a generic repo, but I am no opposed to moving it to l10n-spain, I do think it makes a bit more sense

This integration implements the Delivery portion of the documentation, the pick-ups will be developed on when needed.

No public test user so no real tests, just mocks


delivery_correos_express/models/delivery_carrier.py
```diff
@@ -0,0 +1,260 @@
...
+    def correos_express_cancel_shipment(self, pickings):
+        for picking in pickings.filtered("carrier_tracking_ref"):
+            picking.message_post(
+                body=_(
+                    "Correos Express does not provide a method to cancel a shipment "
+                    "that has been registered. If you need to change some information "
+                    "you, create a new shipment with a new label. This doesn't mean "
+                    "that the shipment will be invoiced, this only happens if the "
+                    "package is picked up and it enters the shipping stage"
+                )
```
victoralmau, line 240:
```suggestion
        raise NotImplementedError(_("""Correos Express does not allow you to cancel a shipment.""")
```

delivery_correos_express/models/delivery_carrier.py
```diff
@@ -0,0 +1,260 @@
...
+                "Correos Express API doesn't provide methods to compute "
+                "delivery rates, so you should rely on another price method "
+                "instead or override this one in your custom code."
+            ),
+            "warning_message": _(
+                "Correos Express API doesn't provide methods to compute "
+                "delivery rates, so you should rely on another price method "
+                "instead or override this one in your custom code."
+            ),
+        }
```
victoralmau, line 275:
```suggestion
      raise NotImplementedError(
            _(
                """
                Correos Express API doesn't provide methods to compute delivery rates, so
            you should relay on another price method instead or override this
            one in your custom code.
        """
            )
        )
```

delivery_correos_express/models/delivery_carrier.py
```diff
@@ -0,0 +1,260 @@
...
+        for picking in pickings:
+            vals = self._prepare_correos_express_shipping(picking)
+            response = correos_express_request.create_shipment(vals)
+            if not response:
+                result.append(vals)
+                continue
+            vals.update(
+                {
+                    "tracking_number": response.get("datosResultado", ""),
+                    "exact_price": 0,
```
victoralmau, line 161:
```suggestion
```

delivery_correos_express/models/delivery_carrier.py
```diff
@@ -0,0 +1,260 @@
...
+        streets = self._get_partner_streets(partner)
+        phone = partner.mobile or partner.phone or ""
+        return {
+            "codDest": "",
+            "nomDest": partner.name or "",  # mandatory
+            "nifDest": "",
+            "dirDest": "".join(streets)[:300],  # mandatory
+            "pobDest": partner.city[:40] or "",  # mandatory
+            "codPosNacDest": partner.zip,  # mandatory
+            "paisISODest": "",
```
victoralmau:
```suggestion
            "paisISODest": partner.country_id.code,
```

delivery_correos_express/models/delivery_carrier.py
```diff
@@ -0,0 +1,260 @@
...
+        return streets
+
+    def _get_correos_express_receiver_info(self, picking):
+        partner = picking.partner_id
+        streets = self._get_partner_streets(partner)
+        phone = partner.mobile or partner.phone or ""
+        return {
+            "codDest": "",
+            "nomDest": partner.name or "",  # mandatory
+            "nifDest": "",
```
victoralmau:
```suggestion
            "nifDest": partner.vat or "",
```

delivery_correos_express/models/delivery_carrier.py
```diff
@@ -0,0 +1,260 @@
...
+            "emailDest": partner.email[:75] or "",
+        }
+
+    def _get_correos_express_sender_info(self, picking):
+        partner = picking.picking_type_id.warehouse_id.partner_id
+        streets = self._get_partner_streets(partner)
+        return {
+            "codRte": self.correos_express_customer_code,
+            "nomRte": partner.name or "",
+            "nifRte": "",
```
victoralmau:
```suggestion
            "nifRte": partner.vat or "",
```

delivery_correos_express/models/delivery_carrier.py
```diff
@@ -0,0 +1,260 @@
...
+            "dirRte": "".join(streets)[:300],  # mandatory
+            "pobRte": partner.city,  # mandatory
+            "codPosNacRte": partner.zip,  # mandatory
+            "paisISORte": "",
```
victoralmau:
```suggestion
            "paisISORte": partner.country_id.code,
```

delivery_correos_express/models/delivery_carrier.py
```diff
@@ -0,0 +1,260 @@
...
+                "emailOtrs": "",
+                "observac": "",
+                "numBultos": number_of_packages or 1,  # mandatory
+                "kilos": "%.3f" % (picking.shipping_weight or 1),  # mandatory
+                "volumen": "",
+                "alto": "",
+                "largo": "",
+                "ancho": "",
+                "producto": "93",  # ePaq24  # mandatory
+                "portes": "P",  # mandatory
```
victoralmau:
```suggestion
                "producto": "93",  # ePaq24  # mandatory
                "portes": "P",  # mandatory
```
Perhaps this could be configured in new fields added in the carrier, right?

delivery_correos_express/models/delivery_carrier.py
```diff
@@ -0,0 +1,260 @@
...
+                            base64.b64decode(
+                                base64.b64decode(label.get("etiqueta1", ""))
+                            )
+                            if is_pdf
+                            else label.get("etiqueta2", "")
+                        ),
+                    )
+                    for index, label in enumerate(response.get("etiqueta"))
+                ]
+            picking.message_post(body=_(""), attachments=attachments)
```
victoralmau, line 186:
```suggestion
                attachments = [
                    (
                        "correos_express_{}_{}.{}".format(
                            response.get("datosResultado", ""),
                            index + 1,
                            "pdf" if is_pdf else "txt",
                        ),
                        (
                            base64.b64decode(
                                base64.b64decode(label.get("etiqueta1", ""))
                            )
                            if is_pdf
                            else label.get("etiqueta2", "")
                        ),
                    )
                    for index, label in enumerate(response.get("etiqueta"))
                ]
            picking.message_post(body=_(""), attachments=attachments)
```
Another option would be to create the attachment directly (example: https://github.com/OCA/delivery-carrier/blob/13.0/delivery_ups_oca/models/delivery_carrier.py#L124-L132)

delivery_correos_express/models/stock_picking.py
```diff
@@ -0,0 +1,39 @@
...
+                    "correos_express_{}_{}.{}".format(
+                        tracking_ref, index + 1, label_format
+                    ),
+                    label,
+                )
+                for index, label in enumerate(decoded_labels)
+            ],
+        )
+        # We return label in case it wants to be used in an inheritance
+        return decoded_labels
```
victoralmau, line 39: Is this function really necessary if the label has already been generated when confirming the picking?

delivery_correos_express/readme/INSTALL.rst
```diff
@@ -0,0 +1,5 @@
+Depende de los módulos de OCA/delivery-carrier `delivery_package_number` y `delivery_state`.
+
+La API de DHL Parcel no provee métodos de cálculo de precio, de modo que para poder
```
victoralmau:
```suggestion
La API de Correos Express no provee métodos de cálculo de precio, de modo que para poder
```

delivery_correos_express/models/correos_express_request.py
```diff
@@ -0,0 +1,98 @@
...
+            if request_type == "GET":
+                res = requests.get(url=url, auth=auth)
+            elif request_type == "POST":
+                _logger.debug(data)
+                res = requests.post(url=url, auth=auth, json=data)
+            else:
+                raise UserError(
+                    _("Unsupported request type, please only use 'GET' or 'POST'")
+                )
+            result = res.json()
```
victoralmau, line 72:
```suggestion
            result = res.json()
            correos_express_last_request = ("URL: {}\nData: {}").format(url, data)
            self.carrier.log_xml(correos_express_last_request, "correos_express_last_request")
            self.carrier.log_xml(result, "correos_express_last_response")
```

delivery_correos_express/models/correos_express_request.py
```diff
@@ -0,0 +1,98 @@
...
+                    [
+                        self.carrier_id.correos_express_username,
+                        self.carrier_id.correos_express_password,
+                    ]
+                )
+            if request_type == "GET":
+                res = requests.get(url=url, auth=auth)
+            elif request_type == "POST":
+                _logger.debug(data)
+                res = requests.post(url=url, auth=auth, json=data)
```
victoralmau:
```suggestion
                res = requests.post(url=url, auth=auth, json=data, timeout=60)
```

delivery_correos_express/models/correos_express_request.py
```diff
@@ -0,0 +1,98 @@
...
+            auth = tuple()
+            if not skip_auth:
+                auth = tuple(
+                    [
+                        self.carrier_id.correos_express_username,
+                        self.carrier_id.correos_express_password,
+                    ]
+                )
+            if request_type == "GET":
+                res = requests.get(url=url, auth=auth)
```
victoralmau:
```suggestion
                res = requests.get(url=url, auth=auth, timeout=60)
```

victoralmau (changes requested):
Good job!
Perhaps, all the private functions specific to this addon and that are currently in the carrier model can be better defined in emails_express_request (having only what is necessary in delivery_carrier can make the revision easier and also add / change to things in the future).

@chienandalu

delivery_correos_express/models/stock_picking.py
hildickethan: This for manually generating the label again, in case it was lost somehow

delivery_correos_express/models/delivery_carrier.py
hildickethan: This returns a dict in order to not fail when using this from a website interface, with the dict it doesnt fail but warns you it might not be accurate. I don't like the idea of it simply not being usable from websites without custom code, I'd rather it work and the user can change the price if need be afterwards

victoralmau: I think it is necessary if the carrier is configured this way
odoo-carrier


But it wouldn't be necessary if another pricing method is used, what do you think?

delivery_correos_express/models/delivery_carrier.py
hildickethan: Does this post it in the chatter at the side? I prefer having it show up there

victoralmau: This comment is only a suggestion although in both cases the attachment will be created and will remain linked to the record (the position of the chatter depends on the user's configuration).

delivery_correos_express/models/delivery_carrier.py
hildickethan: I left this without a raise so it would let you cancel it and re-use the same picking to re-send it with different data if needed

delivery_correos_express/models/delivery_carrier.py
hildickethan: I think I prefer letting Odoo handle the attachment creation with the `message_post`

delivery_correos_express/models/delivery_carrier.py
hildickethan: Where is that config? I don't see it; Either way I still prefer having it not error and instead just warn

hildickethan:
Updated with the changes marked as resolved

delivery_correos_express/models/delivery_carrier.py
victoralmau: Sorry, this is related to the `delivery_price_method` addon (which is not migrated to v14 yet).
It is not necessary then to make any changes.

victoralmau (approved):

pedrobaeza (approved):

PR #436, opened 2021-11-02
[13.0][ADD] delivery_tnt_oca: Integrate TNT webservice

victoralmau:
Integrate TNT webservice

@Tecnativa TT30667


delivery_tnt_oca/models/delivery_carrier.py
```diff
@@ -0,0 +1,170 @@
...
+from .tnt_request import TntRequest
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = "delivery.carrier"
+
+    delivery_type = fields.Selection(selection_add=[("tnt_oca", "TNT")])
+    tnt_oca_ws_username = fields.Char(string="ExpressConnect Username WS")
+    tnt_oca_ws_password = fields.Char(string="ExpressConnect Password WS")
+    tnt_oca_ws_account = fields.Char(string="ExpressConnect Account WS")
```
yajo: What is this "WS"?

delivery_tnt_oca/models/delivery_carrier.py
```diff
@@ -0,0 +1,170 @@
...
+        string="Product service (Non docs)",
+    )
+    tnt_payment_indicator = fields.Selection(
+        selection=[("S", "Sender pays"), ("R", "Receiver pays")],
+        default="S",
+        string="Payment indicator",
+    )
+    tnt_line_of_business = fields.Selection(
+        selection=[
+            ("1", "Domestic dk to dk transfers"),
```
yajo: What's "dk"?

delivery_tnt_oca/models/delivery_carrier.py
```diff
@@ -0,0 +1,170 @@
...
+            ("2", "international non-domestic transfers"),
```
yajo:
```suggestion
            ("2", "International non-domestic transfers"),
```

delivery_tnt_oca/models/delivery_carrier.py
```diff
@@ -0,0 +1,170 @@
...
+        ],
+        default="1",
+        string="Line of business",
+    )
+    tnt_collect_time_from = fields.Char(default="10:30", string="Collect time from",)
+    tnt_collect_time_to = fields.Char(default="16:00", string="Collect time to",)
```
yajo: Usually in Odoo when you want to represent time, you use a `Float` field in the backend and `widget="float_time"` in the frontend. Did you try that?

```suggestion
    tnt_collect_time_from = fields.Float(default=10.5, string="Collect time from")
    tnt_collect_time_to = fields.Float(default=16, string="Collect time to")
```

Of course this will mean some other code changes elsewhere.

delivery_tnt_oca/models/delivery_carrier.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def _compute_tnt_product_code(self):
```
yajo, line 85: - missing `@api.depends(...)`
- missing set to `False` when `delivery_type != tnt_oca`

delivery_tnt_oca/models/delivery_carrier.py
```diff
@@ -0,0 +1,170 @@
...
+        for item in self:
+            if item.tnt_product_type == "D":
+                item.tnt_product_code = item.tnt_product_code_d
+            else:
+                item.tnt_product_code = item.tnt_product_code_n
+
+    def _compute_tnt_product_service(self):
```
yajo, line 96: - missing `@api.depends(...)`
- missing set to `False` when `delivery_type != tnt_oca`

delivery_tnt_oca/models/delivery_carrier.py
```diff
@@ -0,0 +1,170 @@
...
+        for item in self:
+            if item.tnt_product_type == "D":
+                item.tnt_product_service = item.tnt_product_service_d
+            else:
+                item.tnt_product_service = item.tnt_product_service_n
+
+    def tnt_oca_rate_shipment(self, order):
```
yajo, line 106: This doesn't seem like the kind of method that should be called from RPC, right? So better make it private:

```suggestion
    def _tnt_oca_rate_shipment(self, order):
```

delivery_tnt_oca/models/delivery_carrier.py
```diff
@@ -0,0 +1,170 @@
...
+        if response["currency"] != currency.name:
+            price = currency._convert(
+                price,
+                self.env["res.currency"].search([("name", "=", response["currency"])]),
+                company,
+                fields.Date.today(),
+            )
+        return price
+
+    def tnt_oca_action_label(self, picking):
```
yajo: Again should be private?

```suggestion
    def _tnt_oca_action_label(self, picking):
```

delivery_tnt_oca/models/delivery_carrier.py
```diff
@@ -0,0 +1,170 @@
...
+        raise NotImplementedError(
+            _("""TNT API does not allow you to cancel a shipment.""")
+        )
+
+    def tnt_oca_get_tracking_link(self, picking):
+        return "%s/%s?searchType=con&cons=%s" % (
+            "https://www.tnt.com",
+            "express/es_es/site/herramientas-envio/seguimiento.html",
+            picking.carrier_tracking_ref,
+        )
```
yajo: About all these methods, please check if they have to be called from UI and, otherwise, prefix them with `_` to make them private.

delivery_tnt_oca/models/delivery_carrier.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def tnt_oca_tracking_state_update(self, picking):
+        self.ensure_one()
+        if picking.carrier_tracking_ref:
+            tnt_request = TntRequest(self, picking)
+            response = tnt_request.tracking_state_update()
+            picking.delivery_state = response["delivery_state"]
+            picking.tracking_state_history = response["tracking_state_history"]
+
+    def tnt_oc_cancel_shipment(self, pickings):
```
yajo: By "oc" you meant "oca"?

delivery_tnt_oca/models/stock_picking.py
```diff
@@ -0,0 +1,31 @@
...
+    consignment_destination_depot = fields.Char(sparse="tnt_consignment_data")
+    consignment_destination_depot_day = fields.Integer(sparse="tnt_consignment_data")
+    consignment_cluster_code = fields.Char(sparse="tnt_consignment_data")
+    consignment_origin_depot = fields.Char(sparse="tnt_consignment_data")
+    consignment_product = fields.Char(sparse="tnt_consignment_data")
+    consignment_option = fields.Char(sparse="tnt_consignment_data")
+    consignment_market = fields.Char(sparse="tnt_consignment_data")
+    consignment_transport = fields.Char(sparse="tnt_consignment_data")
+    consignment_transit_depot = fields.Char(sparse="tnt_consignment_data")
+    consignment_xray = fields.Char(sparse="tnt_consignment_data")
```
yajo: Prefix these with `tnt_` please.

delivery_tnt_oca/models/sale_order.py
```diff
@@ -0,0 +1,14 @@
...
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    tnt_last_request = fields.Text(
+        string="Last TNT API request", help="Used for debugging", readonly=True
+    )
+    tnt_last_response = fields.Text(
+        string="Last TNT API response", help="Used for debugging", readonly=True
+    )
```
yajo: If these fields are only used for debugging, then I think it's more useful to use logs for that. Just remove these and log the data you think might be needed in the future and you're done.

delivery_tnt_oca/models/stock_picking.py
```diff
@@ -0,0 +1,31 @@
...
+    tnt_piece_data = fields.Serialized()
+    piece_barcode = fields.Char(sparse="tnt_piece_data")
+    tnt_last_request = fields.Text(
+        string="Last TNT API request", help="Used for debugging", readonly=True
+    )
+    tnt_last_response = fields.Text(
+        string="Last TNT API response", help="Used for debugging", readonly=True
+    )
```
yajo: Again, use logs please

delivery_tnt_oca/models/stock_picking.py
```diff
@@ -0,0 +1,31 @@
...
+    consignment_cluster_code = fields.Char(sparse="tnt_consignment_data")
+    consignment_origin_depot = fields.Char(sparse="tnt_consignment_data")
+    consignment_product = fields.Char(sparse="tnt_consignment_data")
+    consignment_option = fields.Char(sparse="tnt_consignment_data")
+    consignment_market = fields.Char(sparse="tnt_consignment_data")
+    consignment_transport = fields.Char(sparse="tnt_consignment_data")
+    consignment_transit_depot = fields.Char(sparse="tnt_consignment_data")
+    consignment_xray = fields.Char(sparse="tnt_consignment_data")
+    tnt_piece_data = fields.Serialized()
+    piece_barcode = fields.Char(sparse="tnt_piece_data")
```
yajo:
```suggestion
    tnt_piece_barcode = fields.Char(sparse="tnt_piece_data")
```

delivery_tnt_oca/models/tnt_request.py
```diff
@@ -0,0 +1,433 @@
...
+            res.raise_for_status()
+            self.record.write(
+                {
+                    "tnt_last_request": ("URL: {}\nData: {}").format(self.url, data),
+                    "tnt_last_response": res.text or "",
+                }
+            )
+            res = res.text
+        except requests.exceptions.Timeout:
+            raise UserError(_("Timeout: the server did not reply within 60s"))
```
yajo, line 52: Where are you setting that timeout?

delivery_tnt_oca/__manifest__.py
```diff
@@ -0,0 +1,28 @@
...
+    "category": "Delivery",
+    "website": "https://github.com/OCA/delivery-carrier",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": [
+        "delivery",
+        "delivery_package_number",
+        "delivery_state",
+        "product_dimension",
+        "base_sparse_field",
```
yajo, line 16: I see you're defining ZPL templates, but I see no dependency on any zpl module. Shouldn't you add it?

delivery_tnt_oca/views/delivery_carrier_view.xml
```diff
@@ -0,0 +1,78 @@
...
+<odoo>
+    <record id="view_delivery_carrier_tnt_oca_form" model="ir.ui.view">
+        <field name="model">delivery.carrier</field>
+        <field name="inherit_id" ref="delivery.view_delivery_carrier_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//notebook" position="inside">
+                <page
+                    string="TNT"
+                    name="configuration_tnt_oca"
+                    attrs="{'invisible': ['|', ('delivery_type', '!=', 'tnt_oca')]}"
```
yajo: This seems like a bug:

```suggestion
                    attrs="{'invisible': [('delivery_type', '!=', 'tnt_oca')]}"
```

yajo (changes requested):
Looks quite good.

delivery_tnt_oca/models/delivery_carrier.py
victoralmau: WS = Webservice.
The names of the fields are similar to those used in other addons of similar carriers: `delivery_ups_oca` or `delivery_seur` for example.

yajo: OK, then you need to reorder the name for it to make sense in English:

```suggestion
    tnt_oca_ws_username = fields.Char(string="ExpressConnect WS Username")
    tnt_oca_ws_password = fields.Char(string="ExpressConnect WS Password")
    tnt_oca_ws_account = fields.Char(string="ExpressConnect WS Account")
```

delivery_tnt_oca/models/stock_picking.py
victoralmau: The debate on these fields is here: https://github.com/OCA/delivery-carrier/pull/430#discussion_r741658032
In my opinion these fields are quite useful if we want to know (without entering the logs) the request and response obtained for each record.

delivery_tnt_oca/__manifest__.py
victoralmau: In my opinion there is no need to add any dependencies relative to .zpl.
Odoo allow us to create a report that has the content of a label in zpl format; this addon is generating that report and creating the attachment with the name "TNT-XX.zpl" to have the file that will later be printed.

delivery_tnt_oca/models/delivery_carrier.py
victoralmau: The `tnt_oca_tracking_state_update` method is correct according to https://github.com/OCA/delivery-carrier/blob/13.0/delivery_state/models/stock_picking.py#L64 (this is done the same in OCA carrier addons that use the dependency on `delivery_state`).

delivery_tnt_oca/models/delivery_carrier.py
victoralmau: This method cannot be changed (the function name is defined at https://github.com/odoo/odoo/blob/13.0/addons/delivery/models/delivery_carrier.py#L149).

delivery_tnt_oca/models/delivery_carrier.py
victoralmau: The names of these methods are correct as defined in https://github.com/odoo/odoo/blob/13.0/addons/delivery/models/delivery_carrier.py#L24-L30

delivery_tnt_oca/models/stock_picking.py
yajo: I don't think an end user can interpret api responses properly. Also I don't think having only the *last* request and response data is helpful (because what happens if the one that failed wasn't the last?).

IMHO it's better to use tools for what they are, and for debugging we have the logs. 🤷🏼‍♂️

delivery_tnt_oca/__manifest__.py
yajo: OK I understand.

delivery_tnt_oca/models/stock_picking.py
chienandalu: In `delivery_schenker` I use them (https://github.com/OCA/delivery-carrier/blob/bb074e0da12737ad0fcce85357b10d15ef3fe50d/delivery_schenker/models/delivery_carrier.py#L186-L187)

And I must say I prefer them as well. They even remove the overhead of having to trick separate cursors and stuff like that as it comes with the core internals already.

victoralmau:
Thanks, changes done.

chienandalu (approved):
This was a hard one :smile: Good job!

delivery_tnt_oca/models/delivery_carrier.py
pedrobaeza: Remove WS at all, as ExpressConnect is the webservice, and there can't be any other thing with that name.

victoralmau:
Can you check it again @Yajo ? Thanks.

victoralmau:
For my part, there are no new changes to make, if everything is correct, it could be merge, right? Ping @pedrobaeza

pedrobaeza (approved):

PR #438, opened 2021-11-12
[13.0][ADD] delivery_gefco: New module

victoralmau:
This module adds Gefco to the available carriers.

@Tecnativa TT30662


delivery_gefco/README.rst
```diff
@@ -0,0 +1,115 @@
...
+
+To configure a Gefco delivery method:
+
+#. Go to *Inventory > Configuration > Delivery > Shipping methods* and create a new one.
+#. Choose *Gefco* as provider.
+#. Configure *Gefco Agency Code* and "Gefco Shipper Identification" fields.
+#. Go to *Inventory > Configuration > Delivery > Gefco Destinations* and import records.
+
+IMPORTANT: It is necessary to import the Gefco destinations that are needed according 
+to the shipping countries of the partners. /data/gefco.destination.csv
```
chienandalu, line 54: Why not to load them directly on the manifest?

victoralmau: This is a performance issue because the file has almost 150,000 records.

chienandalu: Mm, doesn't the orm create them in one go? Did you test it?

pedrobaeza: Anyways that file seems to just reproduce the Spanish ZIP codes and must disappear for using standard data instead.

victoralmau: Gefco has specific codes for each postal code in each country that must be specified on the label.
Do you think it would be better to use the table from `res.better.zip` instead? (the problem of updating that number of records may be the same).

victoralmau:
Finally not needed by the customer and having questionable model data, better to not continue. Anybody wanting to implement it, feel free to take it over from here.

PR #440, opened 2021-11-25
[13.0][FIX] delivery_auto_refresh: Fix error when adding products to cart on website

CarlosRoca13:
cc @Tecnativa 

Ping @chienandalu @pedrobaeza

### DIFF
--- a/delivery_auto_refresh/models/sale_order.py
+++ b/delivery_auto_refresh/models/sale_order.py
@@ -25,6 +25,10 @@ class SaleOrder(models.Model):
     def _get_param_auto_add_delivery_line(self):
         get_param = self.env["ir.config_parameter"].sudo().get_param
         param = "delivery_auto_refresh.auto_add_delivery_line"
+        # When we have the context 'website_id' it means that we are doing the order from
+        # e-commerce. So we don't want to add the delivery line automatically.
+        if self.env.context.get("website_id"):
+            return False
         return safe_eval(get_param(param, "0"))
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Why not adding automatically the delivery line? Steps to reproduce the problem? We should handle it other way if not, but the feature should be kept.

chienandalu:
STR are:

- Set `delivery_auto_refresh.auto_add_delivery_line`
- Set a default delivery method on the user's partner.
- Place a new ecommerce order with that user adding to the cart:

``` 
Error message:

Record does not exist or has been deleted.
(Record: sale.order.line(63,), User: 1)
```

We didn't dive deep into the cause (I guess there's an unlink method over there causing a race condition). Anyway, in an ecommerce context makes no sense to add the line automatically, as the user will be choosing the proper delivery method in the end...

pedrobaeza:
OK, got it, and you are right

pedrobaeza:
@CarlosRoca13 please put these steps to reproduce in the commit message, and a comment in code in the if for saying that having such context means that the order is being done in the e-commerce.

chienandalu (commented):
@CarlosRoca13 add some comments on the code, please

CarlosRoca13:
Done @chienandalu @pedrobaeza

pedrobaeza (approved):

pedrobaeza:
Please fw-port it to v14:

/ocabot merge patch

chienandalu (approved):

PR #443, opened 2021-12-03
[14] Fix encoding on laposte webservice error handling

florian-dacosta:
request is a utf-8 encoded string, it needs to be decoded to be able to replace/hide the webservice password in the call


florian-dacosta:
@hparfr Small review please

hparfr (approved):

PR #445, opened 2021-12-09
[12.0][FIX] delivery_carrier_label_ups: total UPS shipment price

astirpe:
The total price by public rates is always shown on the chatter of the shipment, even when UPS is charging the contracted rate.
This PR is a proposal to get the total price by contracted rate if present, otherwise get the standard price by public rates.

@hbrunn

### DIFF
--- a/delivery_carrier_label_ups/models/stock_picking.py
+++ b/delivery_carrier_label_ups/models/stock_picking.py
@@ -65,15 +65,12 @@ class StockPicking(models.Model):
                 continue
             this._set_a_default_package()
             status = this._ups_request_shipping()
-            total_charges = status["ShipmentResponse"]["ShipmentResults"][
-                "ShipmentCharges"
-            ]["TotalCharges"]
-            price = float(total_charges["MonetaryValue"])
-            if total_charges["CurrencyCode"] != this.company_id.currency_id.name:
+            currency_code, price = self._ups_get_total_price_currency(status)
+            if currency_code != this.company_id.currency_id.name:
                 price = this.company_id.currency_id._convert(
                     price,
                     self.env["res.currency"].search(
-                        [("name", "=", total_charges["CurrencyCode"])]
+                        [("name", "=", currency_code)]
                     ),
                     this.company_id,
                     fields.Date.today(),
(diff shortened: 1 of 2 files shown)


delivery_carrier_label_ups/tests/test_delivery_carrier_label_ups.py
```diff
@@ -136,6 +141,15 @@ def _picking_data(self):
...
         return self.env.ref("delivery_carrier_label_ups.carrier_ups")
 
+    def test_picking_total_price(self):
+        """ Check total contracted price. Negotiated rates
+        have the priority over the published rates.
+         """
+        self._create_order_picking()
+        self.assertTrue(self.picking)
+        self.assertNotEqual(self.picking.carrier_price, 42.00)
+        self.assertEqual(self.picking.carrier_price, 39.89)
```
hbrunn, line 150: I think this is sufficient, it's implied then that it's also not equal so some other number ;-)

hbrunn (approved):

delivery_carrier_label_ups/tests/test_delivery_carrier_label_ups.py
astirpe: Yes, I just thought this way it would be more self-explanatory. But ok, I'll remove it.

StefanRijnhart (approved):

astirpe:
@hbrunn @StefanRijnhart with https://github.com/OCA/delivery-carrier/pull/445/commits/63dc4179dbcb32f50936e5e6a3cc42268c46acb0 I added the logging of request and response

delivery_carrier_label_ups/models/stock_picking.py
```diff
@@ -398,9 +412,16 @@ def _ups_request(
...
-        return getattr(requests, method)(
+        res = getattr(requests, method)(
             url, json=payload, headers=headers, params=query_parameters
         ).json()
+        # log request and response
+        ups_last_request = ("URL: {}").format(url)
+        if payload:
+            ups_last_request += ("\nData: {}").format(str(payload))
+        self.carrier_id.log_xml(ups_last_request, "ups_last_request")
+        self.carrier_id.log_xml(str(res.get("response", {})) or "", "ups_last_response")
```
StefanRijnhart: This creates empty logs in our setup. It seems that successful requests don't have a `response` entry, but a `ShippingResponse`, or a `LabelRecoveryResponse` etc. I think you can log `res` itself.

StefanRijnhart: Thanks, works now!

PR #448, opened 2021-12-14
[MIG] delivery_schenker: Migration to 14.0

hildickethan:
Migration to 14.0, I separated the migration changes from the additions I made along the way so they're easier to backport to 13.0 if desired, those being:

- Add 0.01 minimum to individual package volume, just like the other volume fields
- Add tracking, making use of the `delivery_state` fields. It is considerable to add an extra field to `stock.picking` for the specific Schenker status codes so a customer can organise it entirely with Odoo without having to go to the Schenker website, but I haven't done this yet.
Example of tracking a real delivery, excuse the translation
Captura de pantalla de 2021-12-14 17-25-22
- Add the "Schenker Label" button back and make it work correctly, printing another copy of the label and posting it in the chatter. This is mainly if somehow you lose the PDF saved in the chatter or want to re-print in A4 or A6, different A4 position etc
- Added spanish translation

Ping @chienandalu @victoralmau @pedrobaeza


chienandalu (approved):
Thanks! Looks awesome :slightly_smiling_face: :+1:

PR #453, opened 2022-01-07
[14] Improve delivery_roulier : no failure when asking price / addresses improvements

florian-dacosta:
- Getting the price from a carrier webservice using roulier is not implemented yet. But we avoid failure in case of it is used in Odoo
- Make a seperated method for the address data taken from the parent partner, to ease overrides
- Take email and mobile phone from parent if not found on the picking address
replace https://github.com/OCA/delivery-carrier/pull/410

### DIFF
--- a/delivery_roulier/models/stock_picking.py
+++ b/delivery_roulier/models/stock_picking.py
@@ -157,6 +157,18 @@ class StockPicking(models.Model):
         tomorrow = date.today() + timedelta(1)
         return tomorrow
 
+    def _get_address_info_from_parent(self, partner, address):
+        res = {}
+        if not address.get("company") and partner.parent_id.is_company:
+            res["company"] = partner.parent_id.name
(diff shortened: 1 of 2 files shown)


florian-dacosta:
@hparfr @bealdav

delivery_roulier/models/stock_picking.py
```diff
@@ -157,6 +157,18 @@ def _roulier_get_shipping_date(self, package=None):
         tomorrow = date.today() + timedelta(1)
         return tomorrow
 
+    def _get_address_info_from_parent(self, partner, address):
+        res = {}
+        if not address.get("company") and partner.parent_id.is_company:
```
hparfr, line 162: does partner.commercial_partner_id helps in this case ?

hparfr (approved):
LGTM

PR #455, opened 2022-01-13
[14.0] base_delivery_carrier_label: add check when no carrier_id assigned

sonhd91:
Add check when no carrier assigned on picking
Fix `import mock `
Merge with [PR#458](https://github.com/OCA/delivery-carrier/pull/458) to improve and fix conflict
* Fix `_set_default_package` from `base_delivery_carrier_label`
* Fix `delivery_carrier_pricelist` module test
* Fix  `delivery_package_fee` module test
* Add `delivery_type` check in `delivery_postlogistics`
* Add `delivery_type` check in `delivery_multi_destination`

PR #396, opened 2021-07-17
14.0 mig delivery carrier label default

KKamaa:
migration to v14.0, see sample below; it contains default label and generating .csv docs all in one sample.

delivery-carrier

PR #456, opened 2022-01-13
[14.0] delivery_carrier_label_default: Migrate to version 14.0

sonhd91:
Depend on [PR#455](https://github.com/OCA/delivery-carrier/pull/455)
reuse `_get_packages_from_picking` for print default label for selected packs


leemannd (approved):

sonhd91:
Working on red travis

hailangvn:
Travis is now green.

sebalix (commented):
There was already a PR https://github.com/OCA/delivery-carrier/pull/396 open before this one (we missed it), and the migration commit is quite different, I can't say which one is the best.

I know we are using the current one in production for some months now, but I guess we have to check the other one (ping @sonhd91 ).
One thing we are missing here are the translations and README update commits, but it's minor, not sure we should worry about this.

thomaspaulb:
Seems the main difference is that in #396 we are inheriting `send_shipping` instead of here `alternative_send_shipping` (need to dive deeper into it to judge this, @KKamaa could you comment on this?), as well as deleting some unit tests (which is bad, of course)

@sonhd91 red tests look easy to solve when the Development status of this one is set to Beta

sonhd91:
> Seems the main difference is that in #396 we are inheriting `send_shipping` instead of here `alternative_send_shipping` (need to dive deeper into it to judge this, @KKamaa could you comment on this?), as well as deleting some unit tests (which is bad, of course)
> 
> @sonhd91 red tests look easy to solve when the Development status of this one is set to Beta

Thank, I just update the branch

@sebalix 
The `alternative_send_shipping` was used when we have picking(by default label would have no carrier) with [pricelist module](https://github.com/OCA/delivery-carrier/blob/14.0/delivery_carrier_pricelist/models/delivery_carrier.py#L96), since no carrier was choosen, so the `exact_price` and `tracking_number` should be `False`

delivery_carrier_label_default/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "when no specific carrier is selected.",
+    "version": "14.0.1.0.0",
+    "development_status": "Beta",
+    "category": "Delivery",
+    "website": "https://github.com/OCA/delivery-carrier",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "installable": True,
+    "depends": [
+        "base_delivery_carrier_label",
```
chienandalu, line 15: Is this dependency really needed? Such a large one for a module that could stand alone...

PR #457, opened 2022-01-13
[14.0][IMP] delivery_schenker: Return expected dict instead of error

hildickethan:
Avoid errors when trying to rate the shipment, not returning the `dict` properly makes it unusable for website created orders, when practically the price can just be edited on the sale order if it needs to be customised rather than block the whole creation

### DIFF
--- a/delivery_schenker/models/delivery_carrier.py
+++ b/delivery_schenker/models/delivery_carrier.py
@@ -561,13 +561,20 @@ class DeliveryCarrier(models.Model):
 
     def schenker_rate_shipment(self, order):
         """There's no public API so another price method should be used."""
-        raise NotImplementedError(
-            _(
+        return {
+            "success": True,
+            "price": self.product_id.lst_price,
+            "error_message": _(
                 "Schenker API doesn't provide methods to compute delivery "
                 "rates, so you should relay on another price method instead or "
                 "override this one in your custom code."
-            )
-        )
+            ),
+            "warning_message": _(
+                "Schenker API doesn't provide methods to compute delivery "
+                "rates, so you should relay on another price method instead or "
+                "override this one in your custom code."
+            ),
+        }
 
     # UX Control over not implemented features.
 


pedrobaeza:
Not sure about this, as you have to select to not obtain the carrier price for avoiding the message, not to return a fake one.

hildickethan:
> Not sure about this, as you have to select to not obtain the carrier price for avoiding the message, not to return a fake one.

I haven't seen how to disable returning the carrier price, as far as I saw the `rate_shipment` always gets called and just blocks the order

chienandalu:
Use delivery_price_method. Otherwise you won't be able to charge your customer

chienandalu (commented):
Use delivery_price_method. Otherwise you won't be able to charge your customer

pedrobaeza:
Anyway, if such dependency is not forced, does this change makes sense? How it's done in other carriers?

chienandalu:
This the way we use to implement it. The dependency is a soft one as in some cases the customer develope their own computations in custom and don't need the other module.

This wont make any harm anyway. In any case, why to throw a warning if the price can be considered right?

It should be stated in the readme as well, I think

hildickethan:
You can just edit the sale line with the delivery in order to charge them no? Regardless I don't think this should raise an error and it should flow properly without the other soft dependency module.

But I'm looking into `delivery_price_method` to see if it helps

pedrobaeza:
Can you check on enterprise modules how is done? Or all of them return prices for the carrier?

hildickethan:
They all compute their prices properly so it's hard to judge what should be done for an unimplemented case, but they all return this `dict`

From what I understand if using `delivery_price_method` then it should take priority anyway? This change is more just a fallback as a default value

pedrobaeza:
Isn't better to return 0 instead of list_price then?

pedrobaeza:
Well, thinking twice, better to put a higher price on list price for not giving free deliveries.

hildickethan:
bump to not close, I believe this is still good for a default fallback so it doesn't fail if the code reaches the NotImplemented error, though I will be recommending `delivery_price_method` to clients which also avoids the problem

pedrobaeza:
For me is OK. @chienandalu what do you think?

chienandalu (approved):
Ok for me :+1:

PR #459, opened 2022-01-20
[13.0][FIX] delivery_auto_refresh: behave as core

chienandalu:
We should get as the same default carrier as Odoo does.

cc @Tecnativa TT34020

ping @CarlosRoca13 @pedrobaeza

### DIFF
--- a/delivery_auto_refresh/models/sale_order.py
+++ b/delivery_auto_refresh/models/sale_order.py
@@ -13,14 +13,13 @@ class SaleOrder(models.Model):
         compute="_compute_carrier_id", store=True, readonly=False
     )
 
-    @api.depends("partner_id")
+    @api.depends("partner_id", "partner_shipping_id")
     def _compute_carrier_id(self):
         if hasattr(super(), "_compute_carrier_id"):
             super()._compute_carrier_id()
-        for rec in self:
-            if rec.partner_id:
-                if rec.partner_id.property_delivery_carrier_id:
-                    rec.carrier_id = rec.partner_id.property_delivery_carrier_id
+        for order in self:
+            action = order.action_open_delivery_wizard()
+            order.carrier_id = action["context"]["default_carrier_id"]
 
     def _get_param_auto_add_delivery_line(self):
         get_param = self.env["ir.config_parameter"].sudo().get_param
--- a/delivery_multi_destination/tests/test_delivery_multi_destination.py
+++ b/delivery_multi_destination/tests/test_delivery_multi_destination.py
@@ -145,8 +145,8 @@ class TestDeliveryMultiDestination(common.SavepointCase):
         sale_order_line = order.order_line.filtered("is_delivery")
         self.assertAlmostEqual(sale_order_line.price_unit, 100, 2)
         self.assertTrue(sale_order_line.is_delivery)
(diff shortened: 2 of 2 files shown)


pedrobaeza:
Can you point the code where Odoo does this?

Travis is red, and it seems this change is affecting tests of `delivery_multi_destination`. Maybe they become incompatible or there's an error on what to expect in the tests or lack of precision on the assigned data that leads to the problem. Please check.

chienandalu:
Comment added and tests in `delivery_multi_destination` fixed to ensure test resilience

CarlosRoca13 (approved):

delivery_auto_refresh/models/sale_order.py
```diff
@@ -13,14 +13,18 @@ class SaleOrder(models.Model):
...
-                if rec.partner_id.property_delivery_carrier_id:
-                    rec.carrier_id = rec.partner_id.property_delivery_carrier_id
+        # Compute the carrier as Odoo does in its core
+        # https://github.com/OCA/OCB/blob/27edc28a2c19fad80053e668a8c72e7a629e4c1a
+        # /addons/delivery/models/sale_order.py#L68-L71
+        for rec in self.filtered("partner_shipping_id"):
+            rec.carrier_id = (
+                rec.partner_shipping_id.property_delivery_carrier_id
+                or rec.partner_shipping_id.commercial_partner_id.property_delivery_carrier_id
+            ).filtered("active")
```
pedrobaeza: What do you think about calling the same method and extract the carrier and being compatible with possible extensions?

```suggestion
        for order in self:
            action = order.action_open_delivery_wizard()
            order.carrier_id = action["context"]["default_carrier_id"]
```

delivery_multi_destination/tests/test_delivery_multi_destination.py
```diff
@@ -160,7 +160,8 @@ def test_delivery_multi_destination(self):
         sale_order_line = order.order_line.filtered("is_delivery")
         self.assertAlmostEqual(sale_order_line.price_unit, 50, 2)
         self.assertTrue(sale_order_line.is_delivery)
-        order.partner_shipping_id = self.partner_3.id
+        order.partner_shipping_id = self.partner_3
```
pedrobaeza: Is modifying this line without the `.id` relevant for the test?

chienandalu: No, it was a picky edit

pedrobaeza: There are several of this FYI

delivery_auto_refresh/models/sale_order.py
chienandalu: I like it. Let's go for it then

delivery_multi_destination/tests/test_delivery_multi_destination.py
pedrobaeza: I still see them. Please reduce the diff undoing these changes.

chienandalu: Yeah, fixed

pedrobaeza (approved):

PR #461, opened 2022-01-31
[14.0][FIX] delivery_postlogistics: fix split on bool

mmequignon:
When a packaging doesn't have a `shipper_package_code`, calling `_get_packaging_codes` on it raises an exception.

### DIFF
--- a/delivery_postlogistics/models/product_packaging.py
+++ b/delivery_postlogistics/models/product_packaging.py
@@ -16,4 +16,6 @@ class ProductPackaging(models.Model):
         Return the list of packaging codes
         """
         self.ensure_one()
-        return [code.strip() for code in self.shipper_package_code.split(",")]
+        if self.shipper_package_code:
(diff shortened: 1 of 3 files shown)


delivery_postlogistics/models/product_packaging.py
```diff
@@ -16,4 +16,6 @@ def _get_packaging_codes(self):
         Return the list of packaging codes
         """
         self.ensure_one()
-        return [code.strip() for code in self.shipper_package_code.split(",")]
+        if self.shipper_package_code:
+            return [code.strip() for code in self.shipper_package_code.split(",")]
```
simahawk, line 20:
```suggestion
            return [code.strip() for code in self.shipper_package_code.split(",") if code.strip()]
```

simahawk (approved):

santostelmo (approved):

PR #462, opened 2022-02-03
[FIX] bug with phone/mobile

mathieudelva:
When we have 2 differents value in phone and mobile, the value of the phone was setted with the value of the mobile 
So i changed to complete with the phone first if there are a value and then with de mobile

### DIFF
--- a/delivery_roulier/models/stock_picking.py
+++ b/delivery_roulier/models/stock_picking.py
@@ -198,7 +198,7 @@ class StockPicking(models.Model):
             if address.get(tel):
                 address[tel] = address[tel].replace(u"\u00A0", "").replace(" ", "")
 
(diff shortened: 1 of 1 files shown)


florian-dacosta:
@mathieudelva Actually, it is because we mostly manage one unique number in roulier.
This may need to change, but usually we prefer having a mobile number for carrier stuff than the home number one.

We can't  make this change as it is because it change the behavior for existing projects and this may require a change in roulier lib depending on the carriers.

If we want to start change on this subject, I think in delivery_roulier, we should keep phone as it is, for backward compatibility and mark it as deprecated => to remove in next main version
 and add 2 new fields phoneNumber and mobileNumber.
 Then, in roulier, we could also manage these 3 fields, also marking the phone as deprecated.
 
 But this change as it is is not possible IMO

PR #465, opened 2022-02-08
[FIX] Always add postlogistics_office to customer data

ajaniszewska-dev:
due to https://github.com/OCA/delivery-carrier/pull/183/files postlogistics_office was added to customer data only in case of outgoing pickings. With this setup API returns code "E2002: A domicile post office must be provided" which is preventing correct label generation.

### DIFF
--- a/delivery_carrier_label_postlogistics/postlogistics/web_service.py
+++ b/delivery_carrier_label_postlogistics/postlogistics/web_service.py
@@ -227,12 +227,8 @@ class PostlogisticsWebService(object):
             'ZIP': partner.zip,
             'City': partner.city,
             'Country': partner.country_id.code,
+            'DomicilePostOffice': company.postlogistics_office or "",
         }
-        if (
-            picking.picking_type_id.code == 'outgoing'
-            and company.postlogistics_office
-        ):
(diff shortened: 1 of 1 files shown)


i-vyshnevska (approved):

yankinmax (approved):

delivery_carrier_label_postlogistics/postlogistics/web_service.py
```diff
@@ -227,12 +227,8 @@ def _prepare_customer(self, picking):
             'ZIP': partner.zip,
             'City': partner.city,
             'Country': partner.country_id.code,
+            'DomicilePostOffice': company.postlogistics_office,
```
mmequignon: What if `company.postlogistics_office` is False ?

TDu: Yep maybe
```suggestion
            'DomicilePostOffice': company.postlogistics_office or "",
```
Would be safer.

ajaniszewska-dev: @mmequignon  @TDu  thanks a lot for catching this, fixed code.

TDu (approved):

PR #469, opened 2022-03-18
[FIX] delivery_auto_refresh: don't allow to pick a not allowed carrier

chienandalu:
When choosing the sale carrier from the header of the order form we
should filter those carriers not available for the given shipping
address the in the same way the wizard does it.

cc @Tecnativa TT35200

### DIFF
--- a/delivery_auto_refresh/models/sale_order.py
+++ b/delivery_auto_refresh/models/sale_order.py
@@ -12,6 +12,9 @@ class SaleOrder(models.Model):
     carrier_id = fields.Many2one(
         compute="_compute_carrier_id", store=True, readonly=False
     )
+    available_carrier_ids = fields.Many2many(
+        comodel_name="delivery.carrier", compute="_compute_available_carrier_ids",
+    )
 
     @api.depends("partner_id", "partner_shipping_id")
     def _compute_carrier_id(self):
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
Shouldn't this be made the same in partner's carrier selection?

chienandalu:
> Shouldn't this be made the same in partner's carrier selection?

Maybe that's a little out of the scope for this module

delivery_auto_refresh/models/sale_order.py
```diff
@@ -10,7 +10,10 @@ class SaleOrder(models.Model):
     _inherit = "sale.order"
 
     carrier_id = fields.Many2one(
-        compute="_compute_carrier_id", store=True, readonly=False
+        compute="_compute_carrier_id", store=True, readonly=False,
```
pedrobaeza: Reduce the diff not putting this final comma

delivery_auto_refresh/models/sale_order.py
```diff
@@ -12,14 +12,31 @@ class SaleOrder(models.Model):
...
             super()._compute_carrier_id()
         for order in self:
             action = order.action_open_delivery_wizard()
-            order.carrier_id = action["context"]["default_carrier_id"]
+            carrier_id = self.env["delivery.carrier"].browse(
+                action["context"]["default_carrier_id"]
+            )
+            # If the carrier isn't allowed for the current shipping address, we wont
+            # default to it.
+            order.carrier_id = carrier_id.filtered(
```
pedrobaeza: But if non allowed, it will empty the field. Isn't better to leave the previous one?

chienandalu: Ok, but I'm going to filter both so if any of them is allowed none will be set. Take a look on this new implementation

delivery_auto_refresh/models/sale_order.py
```diff
@@ -12,14 +12,33 @@ class SaleOrder(models.Model):
...
+            carrier_id = self.env["delivery.carrier"].browse(
+                action["context"]["default_carrier_id"]
+            )
+            # If the carrier isn't allowed for the current shipping address, we wont
+            # default to it. In that case we'd try to fallback to the former carrier.
+            order.carrier_id = fields.first(
+                (carrier_id | order.carrier_id).filtered(
+                    lambda x: x in order.available_carrier_ids._origin
+                )
+            )
```
pedrobaeza, line 34: This seems a bit convoluted, and it can trigger other onchanges with no real change, so please change it to this:

```suggestion
            # If the carrier isn't allowed for the current shipping address, we wont
            # default to it. In that case we'd try to fallback to the former carrier.
            carrier = self.env["delivery.carrier"].browse(
                action["context"]["default_carrier_id"]
            ).filtered(lambda x: x in order.available_carrier_ids._origin)
            if carrier:
                order.carrier_id = carrier
```

chienandalu: The problem is that you're not testing if the carrier is allowed for the current shipping address. Case:

- Delivery A - restricted to Spain


Customer commercial partner: delivery method -> Delivery A
Child delivery address in Spain (no delivery method configured)
Child delivery address in France (no delivery method configured)

- We set the order with delivery address in Spain -> Delivery method will be A
- We change the address to France -> Delivery method stays on A which is not correct.

pedrobaeza: OK, got it.

pedrobaeza (approved):

pedrobaeza:
Let's merge it for fixing the bug:

/ocabot merge patch

Please fw-port it to v14

chienandalu:
merge failed

PR #478, opened 2022-04-21
[13.0][IMP] delivery_*: Promote modules to Stable

pedrobaeza:
@Tecnativa

### DIFF
--- a/delivery_package_number/__manifest__.py
+++ b/delivery_package_number/__manifest__.py
@@ -10,6 +10,7 @@
     "license": "AGPL-3",
     "installable": True,
     "application": False,
+    "development_status": "Production/Stable",
     "depends": ["delivery"],
     "data": [
         "views/stock_picking_views.xml",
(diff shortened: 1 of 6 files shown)


pedrobaeza:
@victoralmau do you know the error in UPS?:

https://github.com/OCA/delivery-carrier/runs/6116997723?check_suite_focus=true#step:8:407

rousseldenis:
@pedrobaeza I think the problem is the logged user:

image

pedrobaeza:
Oh, I see, thanks. I have retried now to see if it's unlocked.

pedrobaeza:
The same problem. It seems somebody has blocked the test user using a wrong password, and it locks out the same as the rest of the users. @victoralmau can you try to contact UPS technical support for unlocking it and eventually ask for avoiding such lock?

victoralmau:
> The same problem. It seems somebody has blocked the test user using a wrong password, and it locks out the same as the rest of the users. @victoralmau can you try to contact UPS technical support for unlocking it and eventually ask for avoiding such lock?

Exactly, the user seems to be blocked. I have sent an email to UPS technical support to try to unlock it.

pedrobaeza:
Víctor, any news from UPS?

victoralmau:
Nothing important, maybe we can create an issue on this subject so that odoo can update it.

PR #480, opened 2022-04-25
[13.0][FIX] delivery_ups_oca: Pevent errors in tests.

victoralmau:
Pevent errors in tests.

Related to: https://github.com/OCA/delivery-carrier/pull/478

@Tecnativa

### DIFF
--- a/delivery_ups_oca/models/delivery_carrier.py
+++ b/delivery_ups_oca/models/delivery_carrier.py
@@ -77,6 +77,11 @@ class DeliveryCarrier(models.Model):
         "webservice (you will necessary to activate tracking API)",
     )
 
+    def ups_test_call(self, order):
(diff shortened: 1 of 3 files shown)


chienandalu (commented):
There are developement status dependency errors :/

rousseldenis:
> There are developement status dependency errors :/

Yes, as base modules are in default dev status (Beta), you maybe should remove the dev status entry in manifest.

pedrobaeza:
Please include these changes in #478 instead. I'm not sure though to skip the tests when having error in UPS service. This is the same dilemma as with Geonames. How do you know that error is due to the incorrect user or UPS is having problems or API changes and then your customers are going to suffer such issues? And in this case, it's even more problematic, as you may use the feature everyday (not only importing cities).

PR #481, opened 2022-04-26
[14.0] delivery_carrier_label_batch: Allow testing labels generation

grindtildeath:
The threading option doesn't work in test mode but was not tested
because labels were already generated in the test.
Using a condition to use different threads only when test mode
is not activated allows to test label generation.

Requires: #479

### DIFF
--- a/delivery_carrier_label_batch/wizard/generate_labels.py
+++ b/delivery_carrier_label_batch/wizard/generate_labels.py
@@ -139,18 +139,24 @@ class DeliveryCarrierLabelGenerate(models.TransientModel):
                 groups.setdefault(picking.id, []).append((pack, picking, label))
 
         for group in groups.values():
-            data_queue.put(group)
+            if not getattr(threading.currentThread(), "testing", False):
+                data_queue.put(group)
+            else:
+                self._do_generate_labels(group)
 
-        # create few workers to parallelize label generation
-        num_workers = self._get_num_workers()
-        _logger.info("Starting %s workers to generate labels", num_workers)
-        for _i in range(num_workers):
-            t = threading.Thread(target=self._worker, args=(data_queue, error_queue))
-            t.daemon = True
-            t.start()
+        if not getattr(threading.currentThread(), "testing", False):
+            # create few workers to parallelize label generation
+            num_workers = self._get_num_workers()
+            _logger.info("Starting %s workers to generate labels", num_workers)
+            for _i in range(num_workers):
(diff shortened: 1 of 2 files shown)


santostelmo:
@grindtildeath test_action_generate_labels is not passing in travis yet

delivery_carrier_label_batch/wizard/generate_labels.py
```diff
@@ -139,18 +139,22 @@ def _get_all_files(self, batch):
                 groups.setdefault(picking.id, []).append((pack, picking, label))
 
         for group in groups.values():
```
hailangvn, line 142: This is just a minor non-blocking suggestion.
```suggestion
        is_in_testing = getattr(threading.current_thread(), 'testing', False)
        for group in groups.values():
```

delivery_carrier_label_batch/wizard/generate_labels.py
```diff
@@ -139,18 +139,22 @@ def _get_all_files(self, batch):
...
-        num_workers = self._get_num_workers()
-        _logger.info("Starting %s workers to generate labels", num_workers)
-        for _i in range(num_workers):
-            t = threading.Thread(target=self._worker, args=(data_queue, error_queue))
-            t.daemon = True
-            t.start()
-
-        # wait for all tasks to be done
-        data_queue.join()
+            if not getattr(threading.currentThread(), 'testing', False):
```
hailangvn:
```suggestion
            if not is_in_testing:
```

delivery_carrier_label_batch/wizard/generate_labels.py
```diff
@@ -139,18 +139,22 @@ def _get_all_files(self, batch):
...
+                data_queue.put(group)
+            else:
+                self._do_generate_labels(group)
+
+        if not getattr(threading.currentThread(), 'testing', False):
```
hailangvn:
```suggestion
        if not is_in_testing:
```

hailangvn (approved):

sebalix:
@grindtildeath I pushed a fix in your PR to avoid conflicts, the goal of this fix is to empty the cache once the work done by the threads is finished so the main thread will get up-to-date data (=> mainly the `parcel_tracking` values so the picking labels can be concatenated in the right order by the batch).

StephaneMangin (approved):

sebalix (approved):

PR #483, opened 2022-05-04
[14.0][IMP] delivery_postlogistics: improve web_service and fix bugs

hailangvn:
* This is to cherry-pick web_service changes from #320 (Inverse sender and recipient addresses on incoming shipments);
* Also fix setting package type when this module is installed together with base_delivery_carrier_label;
* Fix partner_id of stock.location as well.

### DIFF
--- a/delivery_postlogistics/models/stock_picking.py
+++ b/delivery_postlogistics/models/stock_picking.py
@@ -8,7 +8,6 @@ from odoo import _, exceptions, fields, models
 
 from ..postlogistics.web_service import PostlogisticsWebService
 
-
 _logger = logging.getLogger(__name__)
 
 
@@ -94,7 +93,9 @@ class StockPicking(models.Model):
             )
             if move_lines:
                 picking._put_in_pack(move_lines)
-        postlogistics_picks = self.filtered(lambda p: p.delivery_type == "postlogistics")
+        postlogistics_picks = self.filtered(
+            lambda p: p.delivery_type == "postlogistics"
+        )
         if postlogistics_picks:
             postlogistics_picks._assign_postlogistics_packaging_on_packages()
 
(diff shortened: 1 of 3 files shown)


delivery_postlogistics/models/stock_picking.py
```diff
@@ -75,26 +75,34 @@ def attach_shipping_label(self, label):
...
+        if self.delivery_type != "postlogistics":
+            return result
+        for move_line in move_line_ids:
+            package = move_line.result_package_id
+            carrier = move_line.picking_id.carrier_id
+            default_packaging = carrier.postlogistics_default_packaging_id
+            if not default_packaging:
+                continue
+            vals = {"packaging_id": default_packaging}
+            package.write(vals)
```
grindtildeath: :warning: Here, you're overwriting `packaging_id` without checking if it's set beforehand.

Actually, if a carrier is set on the picking, before the put in pack operation, Odoo will ask for the packaging to be used through `choose.delivery.package` wizard. And if any other packaging than the default one is set this piece of code will overwrite it.

delivery_postlogistics/models/stock_picking.py
```diff
@@ -75,26 +75,34 @@ def attach_shipping_label(self, label):
             .create(data)
         )
 
+    # When base_delivery_carrier_label is installed, its _set_a_default_package
+    # is called instead of this module's one. So the logic should be moved to
+    # _put_in_pack.
+    # TODO 15.0: refactor to depends on base_delivery_carrier_label
```
grindtildeath: BTW: We definitely do not want to depend anymore on `base_delivery_carrier_label` for this module as it used to be the case before.

grindtildeath:
@hailangvn I did some changes in https://github.com/hailangvn/delivery-carrier/pull/1

sebalix:
Hello, I did a fix here too: https://github.com/hailangvn/delivery-carrier/pull/3

PR #486, opened 2022-05-18
[MIG] delivery_price_method: Migration to 14.0

Vicenteagf:
Migration to 14.0


hildickethan (approved):
Works properly

fuentes73 (approved):

delivery_price_method/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+{
+    "name": "Delivery Price Method",
+    "summary": "Provides fields to be able to contemplate the tracking states"
+    "and also adds a global fields",
+    "author": "Trey (www.trey.es), " "Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/delivery-carrier",
+    "license": "AGPL-3",
+    "development_status": "Production/Stable",
+    "category": "Delivery",
+    "version": "14.0.1.0.2",
```
pedrobaeza:
```suggestion
    "version": "14.0.1.0.0",
```

delivery_price_method/tests/test_delivery_price_method.py
```diff
@@ -0,0 +1,109 @@
...
+                "name": "Test carrier",
+                "delivery_type": "fixed",
+                "product_id": product_shipping_cost.id,
+                "fixed_price": 99.99,
+            }
+        )
+        self.pricelist = self.env["product.pricelist"].create(
+            {
+                "name": "Test pricelist",
+                "item_ids": [],
```
pedrobaeza: This was populated on purpose. Why are you removing it?

pedrobaeza (changes requested):
And please squash both migration commits together.

delivery_price_method/tests/test_delivery_price_method.py
Vicenteagf: On [13.0](https://github.com/odoo/odoo/blob/13.0/addons/delivery/models/delivery_carrier.py#L261) the function fixed_rate_shipment of delivery.carrier returned the fixed_price of the carrier.
On [14.0](https://github.com/odoo/odoo/blob/14.0/addons/delivery/models/delivery_carrier.py#L261) it just returns the product price of the pricelist which in this case is 0.0 and the test fails.

Should I try another approach?

pedrobaeza: This pricelist is put for avoiding weird behaviors when data pricelist has its currency changed (which happened in the past according to the installed modules). Thus, you should adjust the rest of the test code instead of removing this line.

Vicenteagf: I think it works properly now, thank you.

delivery_price_method/tests/test_delivery_price_method.py
```diff
@@ -58,34 +58,35 @@ def _add_delivery(self):
 
     def test_delivery_price_fixed(self):
         sale = self.sale
+        self.pricelist.item_ids[0].write({"fixed_price": 99.99})
```
pedrobaeza, line 61: I still don't understand why is this needed, being the rule to use public price. Shouldn't you put instead this price in the product or see if the pricelist item structure is different?

victoralmau:
Superseed by: https://github.com/OCA/delivery-carrier/pull/518

PR #488, opened 2022-05-21
[14.0][FIX] base_delivery_carrier_label: avoid duplicating carrier_id

hailangvn:
This is to remove duplicated carrier_id.

### DIFF
--- a/base_delivery_carrier_label/views/stock.xml
+++ b/base_delivery_carrier_label/views/stock.xml
@@ -19,10 +19,10 @@
     </record>
     <record id="vpicktree" model="ir.ui.view">
         <field name="model">stock.picking</field>
-        <field name="inherit_id" ref="stock.vpicktree" />
+        <field name="inherit_id" ref="delivery.vpicktree_view_tree" />
         <field name="arch" type="xml">
-            <field name="state" position="before">
-                <field name="carrier_id" />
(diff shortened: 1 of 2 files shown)


base_delivery_carrier_label/views/stock.xml
```diff
@@ -21,8 +21,8 @@
         <field name="model">stock.picking</field>
         <field name="inherit_id" ref="stock.vpicktree" />
```
sebalix: As `carrier_id` doesn't exist in `stock.vpicktree`, you should inherit `delivery.vpicktree_view_tree` instead.
EDIT: I don't know why Travis/Runboat doesn't fail on that, we got the error locally when testing :thinking: 
EDIT 2: it is triggered when we install/update another addon updating the `stock.picking` tree view, and when Odoo tries to apply all the inherited views it fails. The installation/update of `base_delivery_carrier_label` works well on itself even if there is an issue.

sebalix (changes requested):

base_delivery_carrier_label/views/stock.xml
hailangvn: Thanks @sebalix for your review. The view does correctly inherit `delivery.vpicktree_view_tree` now.

sebalix (approved):

PR #493, opened 2022-06-03
[13.0][FIX] delivery_tnt_oca: Encode data to utf-8 to prevent error in some cases (Šempas for example)

victoralmau:
Encode data to utf-8 to prevent error in some cases (Šempas for example)

@Tecnativa TT37181

### DIFF
--- a/delivery_tnt_oca/models/tnt_request.py
+++ b/delivery_tnt_oca/models/tnt_request.py
@@ -39,10 +39,12 @@ class TntRequest(object):
         if data is None:
             data = {}
         try:
-            headers = {"Content-Type": content_type}
+            headers = {"Content-Type": content_type, "charset": "UTF-8"}
             if auth:
                 headers["Authorization"] = "Basic {}".format(self.authorization)
-            res = requests.post(url=url, data=data, headers=headers, timeout=60)
+            res = requests.post(
+                url=url, data=data.encode("utf-8"), headers=headers, timeout=60
+            )
             res.raise_for_status()
             tnt_last_request = ("URL: {}\nData: {}").format(self.url, data)
             self.carrier.log_xml(tnt_last_request, "tnt_last_request")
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Documentation says that requests are handled in that encoding? Isn't this something that can be configure at request level instead of directly encoding? And shouldn't an intermediate method should handle this for doing the conversion for all requests?

victoralmau:
In one use case, the API returned the following error: "_UnicodeEncodeError: 'latin-1' codec can't encode character '\u0160' in position 1873: Body ('Š') is not valid Latin-1. Use body.encode('utf-8') if you want to send it encoded in UTF-8._"

Ok, if you prefer we do it at the request level directly.

pedrobaeza:
Waiting for the other change.

victoralmau:
> Documentation says that requests are handled in that encoding? Isn't this something that can be configure at request level instead of directly encoding? And shouldn't an intermediate method should handle this for doing the conversion for all requests?

The applied change utf8-encodes the data for all requests just before sending it. Do you think it should be done differently?

delivery_tnt_oca/models/tnt_request.py
```diff
@@ -42,6 +42,7 @@ def _send_api_request(
             headers = {"Content-Type": content_type}
```
pedrobaeza: Looking over Internet, it seems you have to put it this way:

```suggestion
            headers = {"Content-Type": content_type; "charset": "UTF-8"}
```

delivery_tnt_oca/models/tnt_request.py
```diff
@@ -39,7 +39,7 @@ def _send_api_request(
         if data is None:
             data = {}
         try:
-            headers = {"Content-Type": content_type}
+            headers = {"Content-Type": content_type, "charset": "UTF-8"}
             if auth:
                 headers["Authorization"] = "Basic {}".format(self.authorization)
             res = requests.post(url=url, data=data, headers=headers, timeout=60)
```
pedrobaeza: But you still require to encode `data`:

```suggestion
            res = requests.post(url=url, data=data.encode("utf-8"), headers=headers, timeout=60)
```

pedrobaeza (approved):

victoralmau:
Ping @sergio-teruel

cesar-tecnativa (approved):

sergio-teruel (approved):

PR #494, opened 2022-06-06
[14.0][ADD] delivery_price_rule_invoice_policy: Rule based pricing

hildickethan:
- Allows the use of the base Odoo "based on rules" pricing with other providers. This is primarily useful when the external provider does not provide a rating method
- Serves as an alternative to `delivery_carrier_pricelist` as it functions in a similar way


pedrobaeza:
This already exists, and the module is https://github.com/OCA/delivery-carrier/tree/13.0/delivery_price_method

hildickethan:
> This already exists, and the module is https://github.com/OCA/delivery-carrier/tree/13.0/delivery_price_method

Nevermind then, I will use this https://github.com/OCA/delivery-carrier/pull/486 thanks

PR #495, opened 2022-06-06
[13.0][FIX] delivery_ups_oca: Show the errors to avoid the error of access to a non-existing key in the code.

victoralmau:
Show the errors to avoid the error of access to a non-existing key in the code.

@Tecnativa TT37248

### DIFF
--- a/delivery_ups_oca/models/ups_request.py
+++ b/delivery_ups_oca/models/ups_request.py
@@ -6,6 +6,7 @@ import logging
 import requests
 
 from odoo import _
+from odoo.exceptions import UserError
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


delivery_ups_oca/tests/test_delivery_ups.py
```diff
@@ -65,7 +65,7 @@ def setUpClass(cls):
         cls.picking.move_lines.quantity_done = 10
         # We make a test call to avoid errors in tests
         response = cls.carrier.ups_test_call(cls.sale)
-        cls.ups_ws_status = True if not response["errors"] else False
+        cls.ups_ws_status = False if "errors" in response else True
```
pedrobaeza:
```suggestion
        cls.ups_ws_status = not ("errors" in response)
```

delivery_ups_oca/models/ups_request.py
```diff
@@ -46,10 +48,13 @@ def _process_reply(
         self.carrier.log_xml(status or "", "ups_last_response")
         errors = status.get("response", {}).get("errors")
         if errors:
-            _logger.info(
-                _("Sending to UPS: %s")
-                % ("\n".join("%(code)s %(message)s" % error for error in errors),)
+            msg = _("Sending to UPS: %s") % (
+                "\n".join("%(code)s %(message)s" % error for error in errors),
             )
+            if self.skip_errors:
```
pedrobaeza: Why not passing this by argument?

victoralmau: I wanted to avoid that all the calls have to add the parameter that I think should be at a general level.

pedrobaeza: Well, this kind of transactional calls should be stateless.

victoralmau: I have added the `_raise_for_status()` function for the errors issue. Do you think it's a good idea to do it like this?

pedrobaeza: Well, sounds good. How is it done in other Odoo integrations?

victoralmau: Each integration does it in a different way, it would be convenient to establish a standard to manage errors in the same way in all integrations.

pedrobaeza: Look if Odoo SA has set any standard, and let's follow it.

pedrobaeza: It's OK this way then.

pedrobaeza (approved):

victoralmau:
Ping @chienandalu

chienandalu (approved):

PR #497, opened 2022-06-09
[13.0][FIX] delivery_ups_oca: Create label from create_shipping response to avoid error.

victoralmau:
Changes done:
- [x] Create label from create_shipping response to avoid error when trying to get the label right after.
- [x] Set the correct URLs.
- [x] Apply domain to `ups_default_packaging_id` field to show only UPS records.
- [x] Set the correct weight if we use packages.
- [x] Fix packages when `package_ids` is not set
- [x] Add `password="True"` to `ups_ws_password` field.
- [x] Add `ups_use_packages_from_picking` field in carrier to use (or not) package from pickings. 
- [x] Set correct `ShipFrom` (warehouse or company address).
- [x] Improve tests to allow inheritance.

`{'code': '9801031', 'message': 'The shipment for the requested tracking number or the combination of reference number plus shipper number could not be found. Please check the submitted data or wait until the shipment is processed.'}`

@Tecnativa TT37248

### DIFF
--- a/delivery_ups_oca/models/delivery_carrier.py
+++ b/delivery_ups_oca/models/delivery_carrier.py
@@ -55,7 +55,9 @@ class DeliveryCarrier(models.Model):
         string="Service code",
     )
     ups_default_packaging_id = fields.Many2one(
-        comodel_name="product.packaging", string="Default Packaging Type"
+        comodel_name="product.packaging",
+        string="Default Packaging Type",
+        domain=[("package_carrier_type", "=", "ups")],
     )
     ups_package_dimension_code = fields.Selection(
         selection=[("IN", "IN"), ("CM", "CM")],
--- a/delivery_ups_oca/models/delivery_carrier.py
+++ b/delivery_ups_oca/models/delivery_carrier.py
@@ -1,5 +1,5 @@
 # Copyright 2020 Hunki Enterprises BV
-# Copyright 2021 Tecnativa - Víctor Martínez
+# Copyright 2021-2022 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import fields, models
 
@@ -85,6 +85,23 @@ class DeliveryCarrier(models.Model):
     def ups_send_shipping(self, pickings):
         return [self.ups_create_shipping(p) for p in pickings]
 
+    def _create_ups_label(self, picking, data):
+        format_code = data["LabelImageFormat"]["Code"].upper()
+        attachment_name = "%s-%s.%s" % (
+            picking.carrier_tracking_ref,
+            format_code,
+            ("txt" if format_code != "PDF" else "pdf"),
+        )
+        return self.env["ir.attachment"].create(
+            {
+                "name": attachment_name,
+                "type": "binary",
+                "datas": data["GraphicImage"],
+                "res_model": picking._name,
+                "res_id": picking.id,
+            }
+        )
+
     def ups_get_label(self, carrier_tracking_ref):
         """Generate label for picking
         :param picking - stock.picking record
@@ -99,24 +116,7 @@ class DeliveryCarrier(models.Model):
         picking = self.env["stock.picking"].search(
             [("carrier_tracking_ref", "=", carrier_tracking_ref)]
         )
-        LabelImageFormatCode = response["LabelImageFormat"]["Code"].upper()
-        if LabelImageFormatCode != "GIF":
-            attachment_name = "%s.pdf" % carrier_tracking_ref
-        else:
-            attachment_name = "%s-%s.%s" % (
-                carrier_tracking_ref,
-                LabelImageFormatCode,
-                LabelImageFormatCode.lower(),
-            )
-        return self.env["ir.attachment"].create(
-            {
-                "name": attachment_name,
-                "type": "binary",
-                "datas": response["GraphicImage"],
-                "res_model": picking._name,
-                "res_id": picking.id,
-            }
-        )
+        return self._create_ups_label(picking, response)
 
     def ups_create_shipping(self, picking):
         """Send packages of the picking to UPS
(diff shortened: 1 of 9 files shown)


pedrobaeza (approved):

delivery_ups_oca/models/ups_request.py
```diff
@@ -18,15 +18,16 @@ def __init__(self, carrier):
...
         self.default_packaging_id = self.carrier.ups_default_packaging_id
+        self.use_packages_from_picking = self.carrier.ups_use_packages_from_picking
         self.shipper_number = self.carrier.ups_shipper_number
         self.service_code = self.carrier.ups_service_code
         self.file_format = self.carrier.ups_file_format
         self.package_dimension_code = self.carrier.ups_package_dimension_code
         self.package_weight_code = self.carrier.ups_package_weight_code
         self.transaction_src = "Odoo (%s)" % self.carrier.name
-        self.url = "https://onlinetools.ups.com"
+        self.url = "https://wwwcie.ups.com"
```
cesar-tecnativa, line 28: could you define these urls as ir.config_parameter ?
May be you may change them to sandbox

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -85,6 +88,35 @@ def ups_test_call(self, order):
...
+            "type": "binary",
+            "datas": values["datas"],
+            "res_model": picking._name,
+            "res_id": picking.id,
+        }
+
+    def _create_ups_label(self, picking, data):
+        val_list = []
+        for PackageResult in data["PackageResults"]:
+            format_code = PackageResult["ShippingLabel"]["ImageFormat"]["Code"].upper()
```
cesar-tecnativa, line 103: format_code = PackageResult["ShippingLabel"]["ImageFormat"]["Code"].upper() or 'txt'

delivery_ups_oca/models/stock_picking.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def ups_get_label(self):
+        self.ensure_one()
+        tracking_ref = self.carrier_tracking_ref
+        if self.delivery_type != "ups" or not tracking_ref:
```
cesar-tecnativa, line 12: if self.delivery_type != "ups" or not tracking_ref or not self.carrier_id:

delivery_ups_oca/models/ups_request.py
```diff
@@ -111,17 +112,27 @@ def _label_data(self):
...
+                self.default_packaging_id, picking, False
+            )
+            package_weight = round(
+                (picking.shipping_weight / picking.number_of_packages), 2
+            )
+            for i in range(0, picking.number_of_packages):
+                package_item = package_info
+                package_name = "%s (%s)" % (picking.name, i + 1)
+                package_item["Description"] = package_name
+                package_item["NumOfPieces"] = "1"
```
cesar-tecnativa, line 132: NumOfPieces? 1

cesar-tecnativa (changes requested):

delivery_ups_oca/models/stock_picking.py
victoralmau: Carrier is always set, therefore this validation is not necessary.

delivery_ups_oca/models/ups_request.py
victoralmau: In other integration modules from other carriers, it has always been defined in the code because the URLs do not change (nor do we want them to change), this change was necessary because they were previously incorrect.

delivery_ups_oca/models/ups_request.py
victoralmau: This value is not too important (it is only needed in some countries where it is mandatory) and it is intended to know how many elements this package contains (if odoo packages are used, this value is defined correctly).

delivery_ups_oca/models/delivery_carrier.py
victoralmau: Personally I prefer to leave this as it is to know exactly the value that the api returns.

pedrobaeza:
@cesar-tecnativa please finish this review

victoralmau:
Ping @cesar-tecnativa

cesar-tecnativa (approved):

delivery_ups_oca/models/ups_request.py
```diff
@@ -158,18 +170,27 @@ def _send_shipping(self, picking):
...
 
     def _quant_package_data_from_order(self, order):
         PackageWeight = 0
         for line in order.order_line.filtered(
             lambda x: x.product_id and x.product_id.weight > 0
         ):
-            PackageWeight += line.product_id.weight * line.product_uom_qty
+            weight_line = line.product_id.weight
+            # Fix if product_dimension_volumetric_weight addon installed
+            if "volumetric_weight" in line.product_id._fields:
```
pedrobaeza: If volumetric weight is admitted by UPS, there should be a hard dependency

victoralmau: Volumetric weight is not required by UPS or any carrier, but carriers typically charge based on volumetric weight (because it is often greater than the weight of the package).

With this little code and only for those who have `product_dimension_volumetric_weight` addon installed, that weight will be taken into account instead of the weight of the product. This is very useful, for example, when you want to obtain the price of a quote per API.

Similar to https://github.com/OCA/delivery-carrier/pull/501#discussion_r941648357

Do you think this should be done differently?

pedrobaeza: Hiding a different way to compute data to rate the delivery is indeed an ugly patch and something to avoid. We need to see if UPS/TNT considers this as correct and official, and if so, put the hard dependency and add the explanation in the README.

victoralmau:
Finally, related changes to `product_dimension_volumetric_weight` has been removed (because it is a deformation of general use case).
Can you review it again?

pedrobaeza (approved):

PR #501, opened 2022-06-13
[13.0][IMP] delivery_tnt_oca: Add pieceLine info from rate_shipment request + Round some data to 2 decimal places

victoralmau:
Changes done:
- [x] Add pieceLine info from rate_shipment request.
- [x] Round some data to 2 decimal places: weight, volume, width, height, lenght.
- [x] Set 0.1 as the minimum value of the volume.
- [x] Apply domain to `tnt_default_packaging_id` field to show only TNT records.
- [x] Set the correct weight if we use packages.
- [x] Limit `addressLine1` to 30 characters.
- [x] Add `addressLine2` to getlabel request.
- [x] Add `tnt_use_packages_from_picking` field in carrier to use (or not) package from pickings. 
- [x] Set correct receiver (warehouse or company address).
- [x] Separate into functions to allow inheritance.
- [x] Improve tests to allow inheritance.

@Tecnativa TT37180

### DIFF
--- a/delivery_tnt_oca/models/tnt_request.py
+++ b/delivery_tnt_oca/models/tnt_request.py
@@ -103,7 +103,8 @@ class TntRequest(object):
             "priceCheck": {
                 "rateId": self.record.name,
                 "sender": self._partner_to_shipping_data(
-                    self.record.company_id.partner_id
+                    self.record.warehouse_id.partner_id
+                    or self.record.company_id.partner_id
                 ),
                 "delivery": self._partner_to_shipping_data(
                     self.record.partner_shipping_id
@@ -226,7 +227,10 @@ class TntRequest(object):
         }
 
     def _prepare_create_shipping(self):
-        receiver = self._prepare_address(self.record.company_id.partner_id)
+        receiver = self._prepare_address(
+            self.record.picking_type_id.warehouse_id.partner_id
+            or self.record.company_id.partner_id
+        )
         del receiver["ACCOUNT"]
         delivery = self._prepare_address(self.record.partner_id)
         del delivery["ACCOUNT"]
--- a/delivery_tnt_oca/models/tnt_request.py
+++ b/delivery_tnt_oca/models/tnt_request.py
@@ -1,4 +1,4 @@
-# Copyright 2021 Tecnativa - Víctor Martínez
+# Copyright 2021-2022 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 import base64
 import json
@@ -27,6 +27,7 @@ class TntRequest(object):
         self.username = self.carrier.tnt_oca_ws_username
         self.password = self.carrier.tnt_oca_ws_password
         self.account = self.carrier.tnt_oca_ws_account
+        self.default_packaging_id = self.carrier.tnt_default_packaging_id
         self.url = "https://express.tnt.com"
         auth_encoding = "%s:%s" % (self.username, self.password)
         self.authorization = base64.b64encode(auth_encoding.encode("utf-8")).decode(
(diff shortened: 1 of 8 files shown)


delivery_tnt_oca/models/tnt_request.py
```diff
@@ -106,10 +112,20 @@ def _prepare_rate_shipment(self):
                 "currency": self.record.currency_id.name,
                 "priceBreakDown": True,
                 "consignmentDetails": {
-                    "totalWeight": totalWeight,
-                    "totalVolume": totalVolume,
+                    "totalWeight": round(totalWeight, 2),
+                    "totalVolume": round(totalVolume, 2),
                     "totalNumberOfPieces": 1,
                 },
+                "pieceLine": {
```
pedrobaeza: I think this should only be informed if present.

delivery_tnt_oca/models/tnt_request.py
```diff
@@ -203,17 +219,20 @@ def _get_data_total_shipping(self):
                 volume += quant.product_id.volume * quant.quantity
         else:
             lines = self.record.move_line_ids_without_package
-            for line in lines.filtered(lambda x: x.qty_done > 0):
+            lines_done = lines.filtered(lambda x: x.qty_done > 0)
+            p_length = max(lines_done.mapped("product_id.product_length"))
+            height = max(lines_done.mapped("product_id.product_height"))
+            width = max(lines_done.mapped("product_id.product_width"))
+            for line in lines_done:
                 volume += line.product_id.volume * line.qty_done
```
pedrobaeza, line 227: Do this through `mapped` as well with `lambda`

delivery_tnt_oca/models/tnt_request.py
```diff
@@ -83,12 +83,18 @@ def _prepare_account(self, partner):
     def _prepare_rate_shipment(self):
         totalWeight = 0
         totalVolume = 0
-        for line in self.record.order_line.filtered(
+        lines = self.record.order_line.filtered(
             lambda x: x.product_id
             and (x.product_id.weight > 0 or x.product_id.volume > 0)
-        ):
+        )
+        totalLength = max(lines.mapped("product_id.product_length")) if lines else 0
```
pedrobaeza: I don't get why using `max`. Maybe you have to package in a different way.

victoralmau:
Can you do a review to merge it if everything is correct?

delivery_tnt_oca/models/tnt_request.py
```diff
@@ -82,20 +84,32 @@ def _prepare_account(self, partner):
...
-        ):
-            totalWeight += line.product_id.weight * line.product_uom_qty
-            totalVolume += line.product_id.volume * line.product_uom_qty
+        lines = self.record.order_line.filtered(
+            lambda x: x.product_id and x.product_id.weight > 0
+        )
+        for line in lines:
+            weight_line = line.product_id.weight
+            # Fix if product_dimension_volumetric_weight addon installed
+            if "volumetric_weight" in line.product_id._fields:
```
pedrobaeza: This is an ugly patch. If TNT works with volumetric height, then the dependency must be explicit.

delivery_tnt_oca/models/tnt_request.py
```diff
@@ -340,15 +358,22 @@ def tracking_state_update(self):
 
     # TntLabel
     def _prepare_label_address(self, partner):
-        return {
+        """Adapt to limit addressLine to 30 characters."""
+        address = "%s%s" % (partner.street, partner.street2 if partner.street2 else "")
```
pedrobaeza: Ugly final result

```suggestion
        address = partner.street or ""
        if partner.street2:
            address += " " + partner.street2
```

delivery_tnt_oca/models/tnt_request.py
```diff
@@ -192,32 +206,36 @@ def _prepare_sender(self):
...
+            p_length = sum(self.record.package_ids.mapped("length"))
         else:
-            lines = self.record.move_line_ids_without_package
-            for line in lines.filtered(lambda x: x.qty_done > 0):
-                volume += line.product_id.volume * line.qty_done
-                p_length += line.product_id.product_length * line.qty_done
-                height += line.product_id.product_height * line.qty_done
-                width += line.product_id.product_width * line.qty_done
+            weight = self.record.shipping_weight
+            # in cm, we need to convert to m
```
pedrobaeza, line 216: Where is stated that this is in cm?

pedrobaeza (changes requested):

delivery_tnt_oca/models/tnt_request.py
victoralmau: Actually this is used ONLY if packages are not used in `stock.picking` (`package_ids`) or you do not want to use the picking packages (`use_packages_from_picking` field of the carrier).
Really only cm is specified in the name of the `product.packing` (_TNT Default (cm)_) that will be used as the default package.

What other option do you propose?

delivery_tnt_oca/models/tnt_request.py
victoralmau: Finally, related changes to `product_dimension_volumetric_weight` has been removed (because it is a deformation of general use case).

pedrobaeza (approved):

cesar-tecnativa (approved):

PR #502, opened 2022-06-13
[14.0][ADD] delivery_cbl: CBL Integration

hildickethan:
Integrate CBL file generation in Odoo. As of the creation of this module, this carrier's services are available in
Spain and Portugal. It covers the 3rd section of the attached documentation. (It's in spanish)
[CBL. Especificacion. Integracion de envios y etiquetado.pdf](https://github.com/OCA/delivery-carrier/files/8892534/CBL.Especificacion.Integracion.de.envios.y.etiquetado.pdf)

Instead of working with a Web Service, this module generates the file necessary to create the CBL shipment.
As stated in the documentation, they do not offer a Web Service, but they offer to install a light software
that can be used to create the shipment with the generated file automatically, but this is external to Odoo.

In the future they may offer a web service that we can use to integrate properly, but for now this serves as an alternative when paired with their software to vaccuum files from a specified directory.


hildickethan:
I haven't tested it with production data yet, but I will notify when I can confirm the module works as is

victoralmau:
Very good contribution, thank you very much.

I think it would be important to add some test to validate, for example, that the content of the generated file is as expected.

Also note that there is the URL https://clientes.cbl-logistica.com/public/consultaenvios.aspx to obtain the status of a shipment (it would be necessary to select the html element afterwards).

hildickethan:
> I haven't tested it with production data yet, but I will notify when I can confirm the module works as is

The client ended up not using this carrier after the migration, so I could never confirm it working in production environment.
Due to this I won't continue working on the PR, but I'm not sure if I should just leave it open in case someone else can use it and superseed it to finish if need be

pedrobaeza:
The stale bot will close it next week, so let's close it and the PR will still be accessible if someone wants to resume it.

PR #509, opened 2022-07-04
[14.0][IMP] delivery_carrier_label_batch: tracking reference purge when regenerating labels

StephaneMangin:
Sometimes, when regenerating labels from third party API, we could not get the new tracking references. And then as we do not purge tracking references correctly, previous ones are kept and lead to incoherences (batch picking tracking references emptied but labels still existing). For this we need to manually purge references AND labels when asking for full regeneration.

Depends on:

- https://github.com/OCA/delivery-carrier/pull/481

### DIFF
--- a/delivery_carrier_label_batch/models/stock_batch_picking.py
+++ b/delivery_carrier_label_batch/models/stock_batch_picking.py
@@ -43,6 +43,7 @@ class StockBatchPicking(models.Model):
             available_options = self.carrier_id.available_option_ids
             default_options = self._get_options_to_add()
             self.option_ids = [(6, 0, default_options.ids)]
+            self.carrier_code = self.carrier_id.code
             return {
                 "domain": {
                     "option_ids": [("id", "in", available_options.ids)],
@@ -81,14 +82,17 @@ class StockBatchPicking(models.Model):
         return values
 
     def write(self, values):
-        """Set the default options when the delivery method is changed.
-
-        So we are sure that the options are always in line with the
-        current delivery method.
-
-        """
+        # - Set the default options when the delivery method is changed (So we
+        #   are sure that the options are always in line with the current
+        #   delivery method)
+        # - Purge all tracking references if a new carrier is applied
         values = self._values_with_carrier_options(values)
-        return super(StockBatchPicking, self).write(values)
+        result = super().write(values)
+        # If a carrier is removed, tracking references are kept until next
+        # carrier change
+        if values.get("carrier_id", False):
+            self.purge_tracking_references()
+        return result
 
     @api.model
     def create(self, values):
@@ -99,4 +103,20 @@ class StockBatchPicking(models.Model):
 
         """
         values = self._values_with_carrier_options(values)
-        return super(StockBatchPicking, self).create(values)
+        return super().create(values)
+
+    def purge_tracking_references(self):
+        """Purge tracking for each picking and destination package"""
+        for batch in self:
+            move_lines = batch.move_line_ids
+            packs = move_lines.result_package_id.filtered(lambda p: p.parcel_tracking)
+            if packs:
+                packs.write({"parcel_tracking": False})
+            pickings = self.env["stock.picking"].search(
+                [
+                    ("move_line_ids", "in", move_lines.ids),
+                    ("carrier_tracking_ref", "!=", False),
+                ]
+            )
+            if pickings:
+                pickings.write({"carrier_tracking_ref": False})
--- a/delivery_carrier_label_batch/tests/test_generate_labels.py
+++ b/delivery_carrier_label_batch/tests/test_generate_labels.py
@@ -1,6 +1,7 @@
 # Copyright 2013-2019 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 import base64
+from unittest.mock import patch
 
 import odoo.tests.common as common
 from odoo import exceptions
@@ -20,6 +21,7 @@ class TestGenerateLabels(common.SavepointCase):
         ShippingLabel = cls.env["shipping.label"]
         BatchPicking = cls.env["stock.picking.batch"]
         cls.DeliveryCarrierLabelGenerate = cls.env["delivery.carrier.label.generate"]
+        cls.PickingBatchApplyCarrier = cls.env["picking.batch.apply.carrier"]
         cls.stock_location = cls.env.ref("stock.stock_location_stock")
         cls.customer_location = cls.env.ref("stock.stock_location_customers")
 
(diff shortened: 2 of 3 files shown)


yvaucher (approved):
LGTM

I wonder if there are some use case out there where you don't want to purge the refs all the time, but I can't have one in mind.

StephaneMangin:
> 



> LGTM
> 
> I wonder if there are some use case out there where you don't want to purge the refs all the time, but I can't have one in mind.

I have none in mind too. And as the use of this module until now doesn't show any other situation, i presume it's ok.

delivery_carrier_label_batch/models/stock_batch_picking.py
```diff
@@ -100,3 +104,26 @@ def create(self, values):
         """
         values = self._values_with_carrier_options(values)
         return super(StockBatchPicking, self).create(values)
+
+    @contextmanager
+    @api.model
```
grindtildeath: Is this decorator required?

delivery_carrier_label_batch/models/stock_batch_picking.py
```diff
@@ -100,3 +104,26 @@ def create(self, values):
...
+        """Purge tracking for each picking and destination package"""
+        # We need a new env to ensure changes will be applied
+        with self._do_in_new_env() as new_env:
+            for batch in self.with_env(new_env):
+                for pack in batch.move_line_ids.mapped("result_package_id"):
+                    if pack.parcel_tracking:
+                        pack.write({"parcel_tracking": False})
+                for picking in batch.move_line_ids.mapped("picking_id"):
+                    if picking.carrier_tracking_ref:
+                        picking.write({"carrier_tracking_ref": False})
```
grindtildeath:
```suggestion
                packs = self.env["stock.quant.package"].search([("id", "in", batch.move_line_ids.mapped("result_package_id.ids")), ("parcel_tracking", "!=", False)]):
                if packs:
                    packs.write({"parcel_tracking": False})
                pickings = self.env["stock.picking"].search([("move_line_ids", "in", batch.move_line_ids.ids), ("carrier_tracking_ref", "!=", False)])
                if pickings:
                    pickings.write({"carrier_tracking_ref": False})
```
I feel like this could be improved with a `search` + single `write` to avoid the loops. (although it make more sense on the pickings than on the packs)

Not blocking anyway.

grindtildeath (approved):
Looks good, two non-blocking comments.

delivery_carrier_label_batch/models/stock_batch_picking.py
StephaneMangin: Testing it, but yes it seems optional as it doesn't manage a try/except statement

delivery_carrier_label_batch/models/stock_batch_picking.py
StephaneMangin: Why not, i test it.

delivery_carrier_label_batch/models/stock_batch_picking.py
StephaneMangin: Well no... it's mandatory for the with statement:
```
  File "/odoo/external-src/delivery-carrier/delivery_carrier_label_batch/tests/test_generate_labels.py", line 223, in test_batch_purge_tracking_reference
    self.batch.purge_tracking_references()
  File "/odoo/external-src/delivery-carrier/delivery_carrier_label_batch/models/stock_batch_picking.py", line 119, in purge_tracking_references
    with self._do_in_new_env() as new_env:
AttributeError: __enter__

```

StephaneMangin: The api.model is dispensable

delivery_carrier_label_batch/models/stock_batch_picking.py
StephaneMangin: With a little adaptation, it's working well.

delivery_carrier_label_batch/models/stock_batch_picking.py
```diff
@@ -43,6 +45,8 @@ def carrier_id_change(self):
             available_options = self.carrier_id.available_option_ids
             default_options = self._get_options_to_add()
             self.option_ids = [(6, 0, default_options.ids)]
+            self.carrier_code = self.carrier_id.code
+            self.purge_tracking_references()
```
simahawk: hmm such business logic should not happen on an onchange IMO.
I think it must happen on write instead.

delivery_carrier_label_batch/models/stock_batch_picking.py
```diff
@@ -100,3 +104,33 @@ def create(self, values):
...
+            yield self.env
+            return
+
+        with api.Environment.manage():
+            with registry(self.env.cr.dbname).cursor() as new_cr:
+                yield api.Environment(new_cr, self.env.uid, self.env.context)
+
+    def purge_tracking_references(self):
+        """Purge tracking for each picking and destination package"""
+        # We need a new env to ensure changes will be applied
```
simahawk: then it's better do it after commit IMO

simahawk (changes requested):
Also, please, rewrite the commit msg ;)

delivery_carrier_label_batch/models/stock_batch_picking.py
StephaneMangin: Further and as discussed, the new env is dispensable too as the write is executed before any kind of third party labels  generation API calls.

delivery_carrier_label_batch/models/stock_batch_picking.py
StephaneMangin: Done

delivery_carrier_label_batch/models/stock_batch_picking.py
```diff
@@ -1,5 +1,6 @@
 # Copyright 2013-2019 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
```
sebalix: To remove (to reduce diff not relevant for the fix applied)

delivery_carrier_label_batch/models/stock_batch_picking.py
```diff
@@ -81,14 +83,17 @@ def _values_with_carrier_options(self, values):
...
-        current delivery method.
-
-        """
+        # - Set the default options when the delivery method is changed (So we
+        #   are sure that the options are always in line with the current
+        #   delivery method)
+        # - Purge all tracking references if a new carrier is applied
         values = self._values_with_carrier_options(values)
-        return super(StockBatchPicking, self).write(values)
+        result = super(StockBatchPicking, self).write(values)
```
sebalix:
```suggestion
        result = super().write(values)
```

delivery_carrier_label_batch/tests/test_generate_labels.py
```diff
@@ -15,14 +16,14 @@ class TestGenerateLabels(common.SavepointCase):
...
-        Picking = cls.env["stock.picking"]
-        ShippingLabel = cls.env["shipping.label"]
-        BatchPicking = cls.env["stock.picking.batch"]
-        cls.DeliveryCarrierLabelGenerate = cls.env["delivery.carrier.label.generate"]
+        stock_move = cls.env["stock.move"]
+        stock_picking = cls.env["stock.picking"]
+        shipping_label = cls.env["shipping.label"]
+        batch_picking = cls.env["stock.picking.batch"]
+        cls.picking_batch_apply_carrier = cls.env["picking.batch.apply.carrier"]
+        cls.label_generate_wizard = cls.env["delivery.carrier.label.generate"]
```
sebalix: These changes are not mandatory and could lead to some conflicts when porting this fix on other branches, even if we don't like these names :)
Better to handle this during a migration.

delivery_carrier_label_batch/tests/test_generate_labels.py
```diff
@@ -15,14 +16,14 @@ class TestGenerateLabels(common.SavepointCase):
...
         cls.stock_location = cls.env.ref("stock.stock_location_stock")
         cls.customer_location = cls.env.ref("stock.stock_location_customers")
-
```
sebalix, line 25: To revert (reducing the diff)

StephaneMangin:
cc @simahawk @sebalix It's all done.

sebalix (approved):

PR #516, opened 2022-08-09
[15.0][MIG] delivery_state: Migration to 15.0

victoralmau:
Migration to 15.0

@Tecnativa TT36949


delivery_state/data/mail_template.xml
```diff
@@ -5,49 +5,64 @@
         <field name="model_id" ref="delivery.model_stock_picking" />
         <field
             name="partner_to"
-        >${object.sale_id.partner_id.id or object.partner_id.id}</field>
+        >{{object.sale_id.partner_id.id or object.partner_id.id}}</field>
```
pedrobaeza: I think this template is no longer needed since several versions, as Odoo includes a similar one, and the automatism.

chienandalu: Yes, I can't see the difference right now: https://github.com/odoo/odoo/blob/15.0/addons/stock/data/mail_template_data.xml#L3

victoralmau:
Changes done.

pedrobaeza:
The automation about sending the mail when the picking is validated already exists in core as well, so you should also remove it. I'm thinking that a valuable feature may be a mail template + configuration to send an automated email when the delivery is completed.

victoralmau:
Changes done and new feature added in a separate commit.

delivery_state/data/mail_template.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
```
chienandalu, line 2: Why do we need another template? We could choose the one in `stock`, isn't it?

victoralmau: Of course, as suggested here https://github.com/OCA/delivery-carrier/pull/516#issuecomment-1210379447 another template has been created and the mechanism to send an automatic email when picking has been delivered to the customer (in a separate commit).

chienandalu: Yes, it's only I can't see the difference with the core one :sweat_smile:

victoralmau: This would be the explanation / difference:
- **Core**: Send an email to the customer when picking is Done (leaves the warehouse).
- **Custom**: Send an email to the customer when the carrier has delivered the package (`delivery_state=customer_delivered`).

chienandalu: Well I meant the template

pedrobaeza: Yeah, the template should be short and stating that the delivery has reached the destiny or similar. Check for similar posts from other e-commerce.

victoralmau: Updated email template.

chienandalu (approved):

pedrobaeza (approved):

PR #519, opened 2022-08-12
[13.0][IMP] delivery_package_number: Hide Number of packages in Immediate Transfer wizard if picking have not carrier.

victoralmau:
Hide Number of packages in Immediate Transfer wizard if picking have not carrier.

@Tecnativa TT38549

### DIFF
--- a/delivery_package_number/wizard/stock_inmediate_transfer.py
+++ b/delivery_package_number/wizard/stock_inmediate_transfer.py
@@ -9,6 +9,11 @@ class StockImmediateTransfer(models.TransientModel):
     number_of_packages = fields.Integer(
         help="Set the number of packages for this picking(s)",
     )
+    ask_number_of_packages = fields.Boolean(compute="_compute_ask_number_of_packages")
+
+    def _compute_ask_number_of_packages(self):
+        for item in self:
+            item.ask_number_of_packages = bool(item.pick_ids.carrier_id)
 
     def process(self):
         if self.number_of_packages:
--- a/delivery_package_number/i18n/delivery_package_number.pot
+++ b/delivery_package_number/i18n/delivery_package_number.pot
@@ -6,12 +6,19 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 13.0\n"
 "Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2022-08-12 10:12+0000\n"
+"PO-Revision-Date: 2022-08-12 10:12+0000\n"
 "Last-Translator: \n"
 "Language-Team: \n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: \n"
 "Plural-Forms: \n"
+
+#. module: delivery_package_number
+#: model:ir.model.fields,field_description:delivery_package_number.field_stock_immediate_transfer__ask_number_of_packages
+msgid "Ask Number Of Packages"
+msgstr ""
 
 #. module: delivery_package_number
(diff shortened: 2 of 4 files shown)


delivery_package_number/wizard/stock_inmediate_transfer.py
```diff
@@ -9,6 +9,14 @@ class StockImmediateTransfer(models.TransientModel):
...
         help="Set the number of packages for this picking(s)",
     )
+    ask_number_of_packages = fields.Boolean(compute="_compute_ask_number_of_packages")
+
+    @api.depends("pick_ids")
+    def _compute_ask_number_of_packages(self):
+        for item in self:
+            item.ask_number_of_packages = fields.first(
+                item.pick_ids
+            ).picking_type_id.ask_number_of_packages
```
chienandalu:
```suggestion
        if len(self.pick_ids) != 1:
            return False
        return bool(self.pick_ids.carrier_id)
```

chienandalu (commented):
I'd say it would be enough to just check if there's a carrier assigned. Maybe there's no need to overbloat picking type config with this... for now :)

victoralmau:
> I'd say it would be enough to just check if there's a carrier assigned. Maybe there's no need to overbloat picking type config with this... for now :)

Changes done according to this (more optimal) approach.

delivery_package_number/wizard/stock_inmediate_transfer.py
```diff
@@ -9,6 +9,14 @@ class StockImmediateTransfer(models.TransientModel):
     number_of_packages = fields.Integer(
         help="Set the number of packages for this picking(s)",
     )
+    ask_number_of_packages = fields.Boolean(compute="_compute_ask_number_of_packages")
+
+    @api.depends("pick_ids")
```
chienandalu: Not sure if this is needed, as pick_ids won't change in the lifetime of the wizard...

delivery_package_number/static/description/index.html
```diff
@@ -3,7 +3,7 @@
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
 <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
-<meta name="generator" content="Docutils 0.15.1: http://docutils.sourceforge.net/" />
+<meta name="generator" content="Docutils: http://docutils.sourceforge.net/" />
```
pedrobaeza: Why this change?

delivery_package_number/wizard/stock_inmediate_transfer.py
```diff
@@ -9,6 +9,13 @@ class StockImmediateTransfer(models.TransientModel):
     number_of_packages = fields.Integer(
         help="Set the number of packages for this picking(s)",
     )
+    ask_number_of_packages = fields.Boolean(compute="_compute_ask_number_of_packages")
+
+    def _compute_ask_number_of_packages(self):
+        for item in self:
+            if len(item.pick_ids) != 1:
+                item.ask_number_of_packages = False
+            item.ask_number_of_packages = bool(item.pick_ids.carrier_id)
```
pedrobaeza: If there's more than one picking, no error will happen as they are joined together in a recordset

```suggestion
            item.ask_number_of_packages = bool(item.pick_ids.carrier_id)
```

chienandalu: The thing is to avoid showing the package number in those cases as well.

pedrobaeza: Yes, and that's the case for my suggestion.

victoralmau:
Changes done.

chienandalu (approved):

pedrobaeza (approved):

PR #528, opened 2022-08-26
[14.0][MIG] delivery_carrier_multi_zip: Migration to 14.0

rizzel:
Migration to 14.0.

Nothing had to be changed. The things I added are:
* added a `default_order` to the tree view
* added a German translation

If some of the commits should be squashed just tell me.


pedrobaeza:
Thanks for going on with this! Please check the tests to see why they are red. You can squash together the first 4 commits, and extract in a separate one the improvements.

rizzel:
I will do the rebase, however the linked tests fail on the `delivery_carrier_location` module - which I did not touch :( 

How best to proceed here?

rizzel:
It seems `delivery_carrier_location` also introduced a field `zip_ids` in the `delivery.carrier` module. I will have a look at that module and see what that does and if both can coexist somehow. Or the functionality is implemented by the other module somehow.

rizzel:
I think I'm done. If you need anything else changed just tell me.

rousseldenis:
@rizzel @pedrobaeza It seems this is weird. First commit does not look like in v13 branch.

victoralmau:
Superseed by https://github.com/OCA/delivery-carrier/pull/539

PR #163, opened 2018-08-27
[MIG] [8.0] base_delivery_carrier_files_document

thomaspaulb:

PR #529, opened 2022-08-29
[8.0][MIG] base_delivery_carrier_files_document

thomaspaulb:
Retake of https://github.com/OCA/delivery-carrier/pull/163


thomaspaulb:
@dreispt We already jumped through the review hoops in https://github.com/OCA/delivery-carrier/pull/163, and you tried to merge there but it failed; the last 2 commits should fix that, so could you try another merge with this PR? The old one got stale.

dreispt:
Hi @thomaspaulb , why are you creating a new PR instead of continuing the one that already had reviews?

thomaspaulb:
@dreispt Because the old one got stale and was closed. I wasn't quick enough. Hence I had to open a new one, or can I reopen the old one?

rousseldenis:
@thomaspaulb Yes, you can reopen the former one

thomaspaulb:
@rousseldenis I cannot, only the PSC has the right to do it it seems. I don't have the checkbox here:

image

But for the `server-tools` repo for example, where I am a PSC, I see a checkbox, and also the "Reopen Pull Request" button is visible for me there on the bottom of the thread, and here it is not.

rousseldenis:
> @rousseldenis I cannot, only the PSC has the right to do it it seems. I don't have the checkbox here:
> 
> image
> 
> But for the `server-tools` repo for example, where I am a PSC, I see a checkbox, and also the "Reopen Pull Request" button is visible for me there on the bottom of the thread, and here it is not.

No that's because you reuse the same branch. But ok we go forward here

rousseldenis:
/ocabot merge patch

Merging for records

PR #530, opened 2022-09-27
[13.0][FIX] delivery_tnt_oca: Set correct Sender (warehouse or company address).

victoralmau:
Set correct Sender (warehouse or company address).

@Tecnativa TT39666

### DIFF
--- a/delivery_tnt_oca/models/tnt_request.py
+++ b/delivery_tnt_oca/models/tnt_request.py
@@ -198,8 +198,12 @@ class TntRequest(object):
         }
 
     def _prepare_sender(self):
-        data = self._prepare_address(self.record.company_id.partner_id)
-        collection = self._prepare_collection(self.record.company_id.partner_id)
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

chienandalu (approved):
:+1: Out of curiosity: shouldn't this be outside the connector logic?

victoralmau:
> +1 Out of curiosity: shouldn't this be outside the connector logic?

Perhaps a connector base module with carriers with some logics would avoid applying the same changes to all carrier modules, but it would be a rather hard dependency that is not expressly necessary.

chienandalu:
> Perhaps a connector base module with carriers with some logics would avoid applying the same changes to all carrier modules

Well, I meant that this Odoo stuff should go in the delivery.carrier methods. The connector class (specific to each API) should be as thin as possible IMO. Anyway, not important right now

PR #533, opened 2022-09-30
[12.0][ADD] delivery_cttexpress: New module

chienandalu:
CTT Express Iberic Web Services API intration with Odoo. Valid for Portugal and Spain.

To do:

- [x] Tests
- [x] Translations
- [x] Functional tests

cc @Tecnativa TT39099

ping @pedrobaeza @victoralmau


delivery_cttexpress/models/cttexpress_request.py
```diff
@@ -0,0 +1,284 @@
...
+    """Interface between CTT Express SOAP API and Odoo recordset.
+       Abstract CTT Express API Operations to connect them with Odoo
+    """
+    def __init__(self, user, password, agency, customer, contract, prod=False):
+        self.user = user
+        self.password = password
+        self.agency = agency
+        self.customer = customer
+        self.contract = contract
+        self.history = HistoryPlugin(maxlen=10)
```
victoralmau, line 52: I prefer to simplify it (e.g. https://github.com/OCA/delivery-carrier/blob/13.0/delivery_tnt_oca/models/tnt_request.py#L20-L22).

chienandalu: IMO the connector class should be completely abstract. In its current form it could even be packaged as a generic connector library for CTT Express ready to use in any python project (and maybe we should do that to ease maintainance :thinking: across versions)

victoralmau (approved):
Good job, code review OK

pedrobaeza (approved):

pedrobaeza:
Switched to GitHub actions in #547, so now the error in CI should be gone:

/ocabot merge nobump

PR #535, opened 2022-10-05
[12.0][FIX] delivery_state: cron

chienandalu:
We were running the cron action over an empty recordset. Anyway code complexity can be reduced as we're repeating `tracking_state_update` method code.

cc @Tecnativa TT39099

### DIFF
--- a/delivery_state/models/stock_picking.py
+++ b/delivery_state/models/stock_picking.py
@@ -62,17 +62,7 @@ class StockPicking(models.Model):
             # These won't ever autoupdate, so we don't want to evaluate them
             ("delivery_type", "not in", [False, "fixed", "base_one_rule"]),
         ])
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Do you know the reason of the previous code splitting by delivery_type?

chienandalu:
> Do you know the reason of the previous code splitting by delivery_type?

Shame on me. To discard those without a carrier with the `tracking_state_update` wildcard method. But this is already done in the caller method, so there's no need to do it twice.

pedrobaeza:
But I think it still applies: there can be carriers that don't implement `<carrier>_tracking_state_update` method

victoralmau (approved):

pedrobaeza (approved):
Forget my last comment. I see it's checked later.

PR #541, opened 2022-10-28
[14.0][FIX] delivery_package_number: number_of_packages set in wizard overriden in compute

mariadforgeflow:
Steps to reproduce:
- Validate a picking that has a `carrier_id` and in the wizard set `number_of_packages` different than 1:
 image

- Check that the value hasn't been updated:
image

@pedrobaeza

### DIFF
--- a/delivery_package_number/models/stock_picking.py
+++ b/delivery_package_number/models/stock_picking.py
@@ -14,6 +14,8 @@ class StockPicking(models.Model):
     )
 
     @api.depends("package_ids")
+    @api.depends_context("force_write_number_of_packages")
     def _compute_number_of_packages(self):
         for picking in self:
-            picking.number_of_packages = len(picking.package_ids) or 1
+            if not self.env.context.get("force_write_number_of_packages", False):
+                picking.number_of_packages = len(picking.package_ids) or 1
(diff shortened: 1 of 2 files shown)


GuillemCForgeFlow (approved):
Nice work :) LGTM :+1:

LoisRForgeFlow (approved):
LGTM

pedrobaeza (changes requested):
Please indicate the steps to reproduce a problem.

And if you include a context in a compute, you should put depends_context decorator.

mariadforgeflow:
Hola @pedrobaeza! Done:)

pedrobaeza (commented):
I think this problem comes from the fact that you are using Odoo packages (Using "Put in pack"), and there's only one package there, so I don't have clear what's the behavior on that case:

- Forbid the modification of the packages
- Your PR having inconsistent information

What's your opinion?

I also invoke @chienandalu 's thoughts on that.

chienandalu (commented):
Yes, the idea was to be able to edit the number of packages just in the case of not having actual packages linked to your picking.

I'd agree to not show the field in the wizard for such cases anyway.

PR #542, opened 2022-11-03
[15.0][MIG] delivery_multi_destination: Migration to 15.0

victoralmau:
Migration to 15.0

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT40225


delivery_multi_destination/tests/test_delivery_multi_destination.py
```diff
@@ -0,0 +1,168 @@
...
+        return carrier_form.save()
+
+    def _create_sale_order(self):
+        order_form = Form(self.env["sale.order"])
+        order_form.partner_id = self.partner_1
+        with order_form.order_line.new() as line_form:
+            line_form.product_id = self.product
+        return order_form.save()
+
+    def _choose_deivery_carrier(self, order, carrier):
```
pedrobaeza:
```suggestion
    def _choose_delivery_carrier(self, order, carrier):
```

delivery_multi_destination/tests/test_delivery_multi_destination.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def _create_carrier(self, childs):
+        carrier_form = Form(self.env["delivery.carrier"])
+        carrier_form.name = "Test carrier multi"
+        carrier_form.product_id = self.product
+        carrier_form.destination_type = "multi"
+        carrier_form.delivery_type = "fixed"
+        carrier_form.fixed_price = 100
+        for child_item in childs:
+            with carrier_form.child_ids.new() as child_form:
```
pedrobaeza, line 91: Is this really needed to be created by `Form`? And why extract it as method if it's called only once?

pedrobaeza (commented):
Is this weird disposition due to the module?

Selección_004

delivery_multi_destination/tests/test_delivery_multi_destination.py
victoralmau: The idea of using `Form` is to check that UX process is correct. 
Extracting the method was to make it more readable than a very long setup. 
Do you prefer to leave it as it was before?

victoralmau:
> Is this weird disposition due to the module?
> 
> Selección_004

Changes done.

victoralmau:
Pre-commit error will be resolved with https://github.com/OCA/delivery-carrier/pull/543

pedrobaeza (approved):

delivery_multi_destination/tests/test_delivery_multi_destination.py
pedrobaeza: In this case it's not adding too much value.

CarlosRoca13 (approved):

PR #522, opened 2022-08-22
[14.0] Add delivery_carrier_package_measure_required

TDu:

PR #546, opened 2022-11-04
[14.0][FIX] delivery_carrier_package_measure_required for not done transfer

TDu:
The required dimensions on a package need not to be enforce before the picking is being set to done.
Because that blocks any preparation on the package before it is delivered.

But we still want to be able to check if the package is valid at will. This can be done with the `ignore_picking_state` context key.

### DIFF
--- a/delivery_carrier_package_measure_required/models/stock_quant_package.py
+++ b/delivery_carrier_package_measure_required/models/stock_quant_package.py
@@ -13,11 +13,17 @@ class StockQuantPackage(models.Model):
     height_required = fields.Boolean(related="packaging_id.package_height_required")
     weight_required = fields.Boolean(related="packaging_id.package_weight_required")
 
-    @api.constrains("length", "width", "height", "weight")
     # The boolean field use to check if a dimension is required are intentionally left out.
     # To not raise error when changing packaging configuration.
+    @api.constrains("length", "width", "height", "weight", "quant_ids")
     def _check_required_dimension(self):
+        ignore_pack_content = self.env.context.get(
+            "delivery_pkg_measure__ignore_package_content"
+        )
         for package in self:
+            if not ignore_pack_content and not package.quant_ids:
+                # Only validate a package when it contains goods
+                continue
             required_dimension = []
             if package.length_required and not package.pack_length:
                 required_dimension.append(_("length"))
--- a/delivery_carrier_package_measure_required/tests/test_delivery_carrier_package_measure_require.py
+++ b/delivery_carrier_package_measure_required/tests/test_delivery_carrier_package_measure_require.py
@@ -49,47 +49,59 @@ class TestDeliveryCarrierPackageMeasureRequire(TestPackingCommon):
             }
         )
         cls.pick.action_confirm()
-
-    def test_required_measurement_are_properly_set(self):
-        """Check required measurement are fullfilled on validation."""
-        pack_action = self.pick.action_put_in_pack()
+        pack_action = cls.pick.action_put_in_pack()
         pack_action_ctx = pack_action["context"]
         pack_wiz = (
-            self.env["choose.delivery.package"].with_context(pack_action_ctx).create({})
+            cls.env["choose.delivery.package"]
+            .with_context(**pack_action_ctx)
+            .create({})
         )
         pack_wiz.action_put_in_pack()
-        package = self.pick.move_line_ids.mapped("result_package_id")
-        package.packaging_id = self.packaging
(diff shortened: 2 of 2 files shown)


TDu:
This one is related to #522 merged yesterday. Who is being used by https://github.com/OCA/wms/pull/433 and testing the new changes with the constraint raised some new issue (see the commit message).
@simahawk @sbidoul

delivery_carrier_package_measure_required/models/stock_quant_package.py
```diff
@@ -13,11 +13,30 @@ class StockQuantPackage(models.Model):
     height_required = fields.Boolean(related="packaging_id.package_height_required")
     weight_required = fields.Boolean(related="packaging_id.package_weight_required")
 
-    @api.constrains("length", "width", "height", "weight")
     # The boolean field use to check if a dimension is required are intentionally left out.
     # To not raise error when changing packaging configuration.
+    @api.constrains("length", "width", "height", "weight")
     def _check_required_dimension(self):
+        ignore_picking_state = self.env.context.get("ignore_picking_state")
```
simahawk:
```suggestion
        ignore_picking_state = self.env.context.get("delivery_pkg_measure__ignore_picking_state")
```

simahawk: Is this something you need only for the case described at https://github.com/OCA/wms/pull/433?
If yes, I would simply move out of this method the check w/out raising an exception.

delivery_carrier_package_measure_required/tests/test_delivery_carrier_package_measure_require.py
```diff
@@ -49,47 +49,57 @@ def setUpClass(cls):
...
+        cls.package.packaging_id = cls.packaging
+
+    def test_force_check_required_measurement(self):
+        self.packaging.package_length_required = True
+        # Picking is not done no validation error
+        self.pick._check_required_package_measurement()
+        # Force validation error raised
+        self.pick = self.pick.with_context({"ignore_picking_state": True})
+        with self.assertRaises(ValidationError):
+            self.pick.button_validate()
```
simahawk:
```suggestion
        pick = self.pick.with_context(delivery_pkg_measure__ignore_picking_state=True)
        with self.assertRaises(ValidationError):
            pick.button_validate()
```

jbaudoux (changes requested):
Better look at the package content. Constrain should apply only if there is content

delivery_carrier_package_measure_required/models/stock_quant_package.py
TDu: It is not about not raising an exception, it is about forcing the check even if the package is still empty.

delivery_carrier_package_measure_required/models/stock_quant_package.py
```diff
@@ -13,11 +13,17 @@ class StockQuantPackage(models.Model):
     height_required = fields.Boolean(related="packaging_id.package_height_required")
     weight_required = fields.Boolean(related="packaging_id.package_weight_required")
 
-    @api.constrains("length", "width", "height", "weight")
     # The boolean field use to check if a dimension is required are intentionally left out.
     # To not raise error when changing packaging configuration.
+    @api.constrains("length", "width", "height", "weight")
```
jbaudoux: No need for quant_ids here for triggering it when the package is filled in ?

TDu: Sounds like a good idea. added.

jbaudoux (approved):

sebalix (approved):

sebalix:
@simahawk if it's OK for you, can we merge?

delivery_carrier_package_measure_required/tests/test_delivery_carrier_package_measure_require.py
```diff
@@ -49,47 +49,57 @@ def setUpClass(cls):
...
         cls.pick.action_confirm()
-
-    def test_required_measurement_are_properly_set(self):
-        """Check required measurement are fullfilled on validation."""
-        pack_action = self.pick.action_put_in_pack()
+        pack_action = cls.pick.action_put_in_pack()
         pack_action_ctx = pack_action["context"]
         pack_wiz = (
-            self.env["choose.delivery.package"].with_context(pack_action_ctx).create({})
+            cls.env["choose.delivery.package"].with_context(pack_action_ctx).create({})
```
simahawk: never replace the ctx, use `**`.

PR #552, opened 2022-11-09
[13.0][IMP] delivery_price_rule_volumetric_weight: Add highest_weight option.

victoralmau:
Add `highest_weight` option + improve tests.

@Tecnativa TT37577

### DIFF
--- a/delivery_price_rule_volumetric_weight/README.rst
+++ b/delivery_price_rule_volumetric_weight/README.rst
@@ -39,7 +39,7 @@ To configure this module, you need to:
 
 #. Go to *Sale -> Configuration -> Shipping Methods*.
 #. Create a new Shipping method or edit some record and set Based on rule providers.
-#. Add new rule and set Volumetric weight in condition and in variable factor.
+#. Add new rule and set Volumetric weight or Highest weight in condition and in variable factor.
 
 Usage
 =====
--- a/delivery_price_rule_volumetric_weight/models/delivery_carrier.py
+++ b/delivery_price_rule_volumetric_weight/models/delivery_carrier.py
@@ -1,4 +1,4 @@
-# Copyright 2021 Tecnativa - Víctor Martínez
+# Copyright 2021-2022 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import models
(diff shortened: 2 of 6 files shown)


delivery_price_rule_volumetric_weight/README.rst
```diff
@@ -39,7 +39,7 @@ To configure this module, you need to:
 
 #. Go to *Sale -> Configuration -> Shipping Methods*.
 #. Create a new Shipping method or edit some record and set Based on rule providers.
-#. Add new rule and set Volumetric weight in condition and in variable factor.
+#. Add new rule and set Volumetric weight or Ponderate weight in condition and in variable factor.
```
chienandalu: I'm not sure that **ponder** is the proper english term for this (https://www.merriam-webster.com/dictionary/pondered). It would be **pondered** in any case, as **ponderate** means **deliberate** (https://www.merriam-webster.com/dictionary/ponderate#dictionary-entry-2)

chienandalu: ping @LudLaf

pedrobaeza (approved):
Apart from David's comments

delivery_price_rule_volumetric_weight/README.rst
LudLaf: "Ponderate" has a meaning that doesn't fit in this description. An option could be "weighted average". Only thing is, it'll seem to be a redundancy ("weighted average weight"), but probably "weighted average load" instead of "Ponderate weight" will help.

ping @victoralmau @chienandalu

pedrobaeza: What we want to express is the maximum between the volumetric weight and the normal weight, but in the most condensed way. What a challenge... Any suggestion?

chienandalu: What about modulated_weight?

LudLaf: What about "higher weight"? Quite simple, and suggests the biggest option between volumetric weight and physical weight.

chienandalu: > What about "higher weight"?

Wouldn't it be highest weight?

LudLaf: Sure, highest weight.

victoralmau:
Changed to `highest_weight` option.

chienandalu (approved):
Great! :+1:

PR #557, opened 2022-11-10
[13.0][ADD] delivery_sending: New module

victoralmau:
New module

@Tecnativa TT40374


delivery_sending/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This module links the `Sending <https://www.senging.es>`_ booking and tracking
```
chienandalu:
```suggestion
This module links the `Sending <https://www.sending.es>`_ booking and tracking
```

delivery_sending/models/sending_request.py
```diff
@@ -0,0 +1,147 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+# Copyright 2022 Impulso Diagonal - Javier Colmeiro
+# Copyright 2022 Tecnativa - Víctor Martínez
+import logging
+
+from suds.client import Client
+from suds.sax.text import Raw
```
chienandalu, line 7: Wasn't `zeep` suitable for the task?

delivery_sending/models/sending_request.py
```diff
@@ -0,0 +1,147 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+class SendingRequest(object):
+    """Interface between Sending SOAP API and Odoo recordset
+       Abstract Sending API Operations to connect them with Odoo
+    """
+
+    def __init__(self, carrier):
```
chienandalu: Non blocking, but you know I like to abstract the request class away from any odoo object :grin:

chienandalu (commented):
Testing pending. Some little comments for the moment:

delivery_sending/models/sending_request.py
victoralmau: The current behavior works correctly. If there was a test user for the API I would change it to `zeep`.

delivery_sending/models/sending_request.py
victoralmau: Yes, I know, but I think just the opposite.

delivery_sending/models/sending_request.py
chienandalu: Ok, I didn't notice it was coauthored

delivery_sending/tests/test_delivery_sending.py
```diff
@@ -0,0 +1,83 @@
...
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.picking = cls.sale.picking_ids[0]
+        cls.picking.move_lines.quantity_done = 10
+
+    def test_delivery_carrier_sending_integration(self):
+        self.picking.action_confirm()
+        self.picking.action_assign()
+        with self.assertRaises(UserError):
+            self.picking.send_to_shipper()
```
chienandalu: IMO this test isn't very useful and adds a unnecesary delay in the CI until the connection timeout is raised.

chienandalu (approved):
Tested as far as I could. Thanks! :+1:

delivery_sending/tests/test_delivery_sending.py
victoralmau: Yes, the truth is that without a test user it is not very useful. What do you think @pedrobaeza?

pedrobaeza: If there's a connection timeout, then indeed it should go out. The value of this is that the data to send prepare methods at least works. Is there a way to circumvent this?

victoralmau: I have changed tests to simulate a complete process.

delivery_sending/models/sending_request.py
pedrobaeza: I'm with @chienandalu. This class is not ORM (`models.Model`), but Python one (`object`), so only Python arguments should be passed for eventually extracting this isolatedly from Odoo.

victoralmau: Ok, changes done.

pedrobaeza (approved):

PR #568, opened 2022-11-23
[16.0][MIG] migrate delivery_carrier_info

florian-dacosta:
Since Odoo 16, there is a field `carrier_description` on `delivery_carrier` : https://github.com/odoo/odoo/blob/16.0/addons/delivery/models/delivery_carrier.py#L58

So the module `delivery_carrier_info` do not need to add it anymore and just adds the field `code`.
That is why I have renamed it accordingly.

I made a migration script, I guess we we'll need to add the module in openupgrade's apriori file when merged.
@pedrobaeza Could you confirm me this is the good way to go ?

@sebastienbeau @fshah-initos review welcome


bealdav (approved):
Thanks for this contribution.
Good idea to have avoided sql in script to deal with translations
thanks

pedrobaeza:
I would still keep the name `delivery_carrier_info` for possible extra fields added in the future, although now it only contains the `code`.

florian-dacosta:
@pedrobaeza Ok, indeed, let's avoid useless renaming in case of future new fields.
I've updated.

pedrobaeza:
Thanks for reconsidering the rename. The README is still pointing to the renamed one. Can you please check it and see if other parts point to the `_code` one?

pedrobaeza (approved):

florian-dacosta:
@pedrobaeza I have deleted README/Index files and re-generated it, it should be good now, I did not notice anymention of _code that should not be.
I did not think it was mandatory to change the auto-generated files.

PR #569, opened 2022-11-23
[15.0][MIG] delivery_free_fee_removal: Migration to 15.0

victoralmau:
Migration to 15.0

@Tecnativa TT36945


delivery_free_fee_removal/tests/test_delivery_free_fee_removal.py
```diff
@@ -0,0 +1,131 @@
+# Copyright 2019 tecnativa - Vicent Cubells
```
pedrobaeza:
```suggestion
# Copyright 2019 Tecnativa - Vicent Cubells
# Copyright 2022 Tecnativa - Víctor Martínez
```

delivery_free_fee_removal/tests/test_delivery_free_fee_removal.py
```diff
@@ -60,6 +63,10 @@ def test_delivery_free_fee_removal_with_fee(self):
                 }
             ],
         )
+        res = self.report_obj._get_report_from_name(
```
pedrobaeza, line 67: Why do you need to test the report? The same question for the next ones.

victoralmau: The purpose of testing the report is: improve the coverage and check that the delivery line is indeed displayed or not (depending on the use case).

pedrobaeza: OK, I see now that there's a modification on the sales report.

pedrobaeza (approved):

chienandalu (approved):
Functional and code review :+1: Thanks!

Issue #537, opened 2022-10-11
Split base_delivery_carrier_label


florian-dacosta:
Hello,
`base_delivery_carrier_label` is not used as a dependency for all oca carrier module. I think it is a shame not to have some consistency between oca modules on this matter, because that becomes a problem when using multiple carriers that works differently (not user friendly + harder to override in case of custom behavior).

The most important example, IMHO, is the `carrier.account` model.
In some module like delivery_ups_oca (probably multiple others) the carrier account is set directly on the delivery method, on others, based on base_delivery_carrier_label, it is managed with the `carrier.account` model from `base_delivery_carrier_label`.

I believe it is important that all modules use `carrier.account` for the following reasons : 

- No `delivery.carrier` duplication in case we use multiple carrier accounts.
  In case of multi company instance for example, it allows you to share the `delivery.method `between companies
  In case of complicated use case you may have multiple account for a same carrier depending on custom criterias. In this case, you definitely don't want to duplicate the `delivery.carrier`
  
- no carrier account duplication : some times we have multiple delivery method for 1 carrier (depending on the service...) and it is easier to have a separated object for this, to avoid duplicating the account information on each delivery method.

- possibility tu customize the way to get the account information for all carrier instead of overriding the same for each carrier...

- we basically have at least an account number/password as common field for all carrier account, no duplicating this fields for each carrier make is easier, for instance we have a generic carrier_account_environment module instead of having one by carrier...

I am probably missing other pros arguments, and I don't really see any cons ?

The reason for not using `base_delivery_carrier_label` as a dependency I was told, if I remember well, was because it had too much features, it added too much maintenance burden, etc
That is why, I'd like to propose to split this module for version 16.

As I see it, base_delivery_carrier_label has 3 main features/areas : 
1) carrier account model and (overridable) method to choose the right one
2) Possibility to have option for a delivery carrier with possibility to choose specific option for a picking and not another...
3) miscellaneous helpers that could help/standardize the carrier modules (special model to store labels, store parcel number on pack, auto generate pack if missing, etc)

So we would have 3 modules
1) delivery_carrier_account => dependency of all OCA carrier modules
2) delivery_carrier_option => optional for carrier modules that want to use it
3) delivery_carrier_helper => optional for carrier modules (I would make it mandatory as well, but I guess it is more controversial so let's forget this)

What do you think ? If the module is splitted in this way, would you be ok to make `delivery_carrier_account` a mandatory dependency for all carrier modules ?
[...]

PR #570, opened 2022-11-23
[16][MIG]  base_delivery_carrier_label + extract delivery_carrier_account

florian-dacosta:
Depends on https://github.com/OCA/delivery-carrier/pull/568

I have extracted the code relative to the carrier accounts from `base_delivery_carrier_label` to a new module `delivery_carrier_account`

The goal is to make it easy for each specific carrier module to depend on use these carrier account, which I believe would be very beneficial in multiple case like I explain here : https://github.com/OCA/delivery-carrier/issues/537

There are no changes in the functionalities.

In the future I guess we may want to extract also the options system in a separated module, but it is not a priority for me because there is nothing to really standardize around this in the specific carrier modules, at the contrary of the carrier accounts.

It seems logical to have only 1 PR for the 2 modules since they both come from the same v15 module, I hope it is ok.


rousseldenis (commented):
@florian-dacosta This LGTM.

Would it deserves migration scripts for carrier account model?

florian-dacosta:
@rousseldenis I am not sure it is needed, since there are no change in the database structure.
That is just a new module but that is a dependency of `base_delivery_carrier_label` so it should install itself in case of update of `base_delivery_carrier_label` and go without issue.
Well I did not test it, not sure about this

rousseldenis (approved):
Code review. LGTM

lmignon (approved):

delivery_carrier_account/models/carrier_account.py
```diff
@@ -0,0 +1,44 @@
...
+        help=(
+            "This field may be used to link an account to specific delivery methods"
+            " It may be usefull to find an account with more precision than with "
+            "only the delivery type"
+        ),
+    )
+    account = fields.Char(string="Account Number", required=True)
+    password = fields.Char(string="Account Password", required=True)
+    company_id = fields.Many2one(comodel_name="res.company", string="Company")
+    file_format = fields.Selection(
```
pedrobaeza: This shouldn't be here, as the format will depend on the carrier, and having XML as a possibility when the carrier doesn't support it is misleading.

delivery_carrier_account/models/carrier_account.py
```diff
@@ -0,0 +1,44 @@
...
+    _order = "sequence"
+
+    @api.model
+    def _selection_file_format(self):
+        """To inherit to add label file types"""
+        return [("PDF", "PDF"), ("ZPL", "ZPL"), ("XML", "XML")]
+
+    name = fields.Char(required=True)
+    sequence = fields.Integer()
+    delivery_type = fields.Selection(
```
pedrobaeza, line 21: Is this working when a new module is installed? I have also doubts about the translations. I use this method:

https://github.com/OCA/sale-workflow/blob/6dd6786ce468f5ef33fa8e11cfd842ac5223ddee/sale_order_type/models/sale_order_type.py#L17

delivery_carrier_account/models/carrier_account.py
```diff
@@ -0,0 +1,44 @@
...
+        selection=lambda self: self.env["delivery.carrier"]
+        ._fields["delivery_type"]
+        .selection,
+        help="This field may be used to link an account to a carrier",
+    )
+    carrier_ids = fields.Many2many(
```
pedrobaeza: I don't like the complication on the design having a m2m instead of a many2one is worth. I know the case you have exposed, but this is an anti-UX design, and also not compatible with the rest of the system (you won't see which specific delivery carrier case will be used in for example the sales order). It's not too-much trade-off to create 2 delivery carriers linked to the same account.

pedrobaeza (changes requested):

delivery_carrier_account/models/carrier_account.py
florian-dacosta: I fully agree that this field is hardly usable. I removed it.

delivery_carrier_account/models/carrier_account.py
florian-dacosta: You are right, the translation are not working this way... I've updated following your example, thanks

delivery_carrier_account/models/carrier_account.py
florian-dacosta: This many2many is totally optional and should be used only for specific carriers.
I don't see how it can be replaced by a many2one though.

The most common case is to have one account by carrier, and possibly multiple delivery method by carrier.
In that case, you don't need to to fill this `carrier_ids` field, because the `delivery_type` you chose is enough to find the right account.
But in the case for a same carrier, you may have to choose a different account depending on the delivery method, then you have to fill the field, and only in this case.
It happens for DPD (different account for the relay point delivery method) and Geodis for example.

Since the case may happen, I think it is important to keep the possibility to link an account to delivery method this way, but what could be done is to show this field only for carrier that may need it.
For instance, adding an invisible computed field `show_carrier_ids` which would just check if a method `_get_delivery_type_allowing_carrier_ids` (I'd have to find a better name for the method..) contains the carrier account's delivery type.

Then, the modules delivery_dpd, delivery_geodis could just override `_get_delivery_type_allowing_carrier_ids` to add their own `delivery_type` and make the field visible.
Would it be better this way?

pedrobaeza: The anti-pattern I'm seeing is that you select the match on the account through delivery_type, instead of selecting the account on the delivery carrier. That is a lot more confusing. So I think we should put a many2one `delivery_account_id` in `delivery.carrier` for selecting the account, restricting the possible values to the ones with the same `delivery_type`.

florian-dacosta: But one of the goal of this module beeing to avoid a lot of data duplication but having the possibility to have multiple account for 1 delivery carrier, the many2one can't work.

The delivery type beeing a carrier and the delivery method beeing just a specific service of the carrier, it makes sense that we select the match on the account throught delivery type.
And we still keep possibility to filter by delivery method, only if needed, to keep some flexibility.

I can understand the preference to configure and show a direct link on the delivery method, but we need to cover all the uses cases and this does not...

Another idea could be to add a computed readonly m2o field on the side on the delivery_carrier, to show which account will be chosen for a given delivery method, so it is very clear of the users. But the field would only be informational.
And, if there are multiple accounts for a delivery order, which is a more advanced case, then it would stay empty.

I am really open to changes, but the uses cases stated must continue to be possible.

pedrobaeza: As said, adding several carrier accounts is an anti-pattern that makes you later to have to select the proper one on another step, complicating everything, and not being able to show the proper operation on the right moment (for example, on the sales order). The use case should be covered with several carriers, not with one.

florian-dacosta: @pedrobaeza Ok, clear enough.
Well, on my side, I am doing this migration for the customer that needs multiple account per carrier/delivery method (white label dropshipping case explained before).
They use 5 carriers/23 delivery methods / 5 accounts by carrier (5 white label company as customer).

If we add a many2one on the `delivery.carrier`, it will indeed make the multiple accounts per delivery method impossible, which was one of the initial goal of this module (model). It would force me to configure 115 `delivery.method` (instead of 23).
Probably even more in the future.
I won't use the module if it is based on a many2one on `delivery.carrier`.
Sadly, I see 2 options only : 
1) We keep the module as it is, since it is already a used OCA module from several version
I understand you'd continue not using it
2) Merge is blocked into the OCA until the change you ask is done => It won't be usable for me anymore, I'll use a fork or something and will leave someone else continue on this PR.

Don't see any lack of willingness in the proposal, I am just no seeing other solution.

pedrobaeza: A third option is to keep it with the many2one, and just add the m2m in a module over it with such functionality calling it `delivery_carrier_account_multi_carrier`.

florian-dacosta: I thought a bit about it.
What I dislike is that delivery_carrier_account is a small module, we'll override it with an even smaller module that will change the first module behavior and that it will be difficult to keep compatibility between carrier specific modules and our submodule `delivery_carrier_account_multi_carrier`

Indeed, for this to work well (compatibility with all carrier modules), we'd have to keep the method `_get_carrier_account` on `stock.picking`. In the base module ,the method would just return `self.carrier_id.carrier_account_id`.
In the submodule `delivery_carrier_account_multi_carrier`, an override would return what the method is returning today (a search between the accounts with an overridable domain).

But for compatibility of `delivery_carrier_account_multi_carrier` with all carrier specific modules, all of those should get the account doing `picking._get_carrier_account()`, if they by pass the method getting the account just doing `picking.carrier_id.carrier_account_id`, the submodule `delivery_carrier_account_multi_carrier` won't have any effect.

How can we ensure this in that case ?

pedrobaeza: I still don't see how do the user selects the proper account in the picking or whatever? Or do you tell me that it only relies on the method `_get_carrier_account` to return the proper one because the user knows how to avoid collisions in the configuration?

I still see a very bad idea to do this m2m.

florian-dacosta: It only relies on `_get_carrier_account` The normal user don't need to know which account is used, it does not matter to him

Carrier accounts basically never changes, the person in charge of its configuration have to get the information from the carrier, if for 1 carrier, there are multiple accounts, the person in charge is very well aware of the reason and he configures the account, once when the carrier start beeing used.
Basic case, you just create your 1 account with one delivery type, nothing difficult.
less basic case, you have different accounts depending on the delivery.method, you have to select the delivery type, and then since you know that you have multiple account for the carrier, you also select the delivery methods concerned.
Advanced case, you have multiple account for a delivery method, it depends on other criterias (in my case a sale channel)
The person know he has 1 account by sale channel, he goes and select delivery type and the sale  channel concerned (require additional module of course). In my experience, it never has been an issue, and, once it is done, it never changes. 

Anyway, I understand you don't deal with what I called "advanced case" so you don't see why we would not just have a many2one on carrier which would indeed be the easiest. But we have to leave a solution for advanced case IMO.
Like I said, without this possibility, I just won't have any use for the module for example.

In order to go forward with this subject, I propose we then go with this imperfect solution : 
Remove `carrier_ids` and remove `_get_carrier_account_domain`
Add the many2one on `delivery.carrier` (I guess mandatory if delivery_type is not fixed/based on rule)
Adapt `_get_carrier_account` to return the account set on the `delivery.carrier`

Then, each carrier specific modules must use the method `_get_carrier_account`.
If some module depends on `delivery_carrier_account` and don't use `_get_carrier_account`, we'll just make a PR to change that.
Not perfect because I guess some module won't use the method, but, in case someone needs it to call the method, it is an easy/fast PR.

For the advanced case management, it will be enough to : 
Hide the many2one on `delivery.carrier`
Add the required fields to manage multiple account on `carrier.account`
Override `_get_carrier_account` to call a new method `_get_carrier_account_domain`

It is quite easy, it does not complexify `carrier_account` 
The only issue is that we can't make carrier_specific module to use the `_get_carrier_account` method but I will add a comment and if it is a problem, we'll just adapt the specific carrier modules when the problem occurs.

What do you think ?

pedrobaeza: OK, that way seems reasonable, thanks.

florian-dacosta: I have made the changes, if you want to have a look.
I have added a dependency on web_domain_field, to manage the domain on the many2one on the `delivery.carrier `model.
Not sure we have another choice to manage the case, but the module since well maintained so it seems quite reasonable.

delivery_carrier_account/models/delivery_carrier.py
```diff
@@ -0,0 +1,24 @@
...
+import json
+
+from odoo import api, fields, models
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = "delivery.carrier"
+
+    carrier_account_id = fields.Many2one(
+        "carrier.account", string="Account", company_dependent=True
```
pedrobaeza: The domain can be put statically and don't require additional modules:

```suggestion
        "carrier.account",
        string="Account",
        company_dependent=True,
        domain="[('delivery_type', '=', delivery_type)]",
```

florian-dacosta: Thanks, I fixed it

delivery_carrier_account/models/carrier_account.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2014-TODAY Florian da Costa Akretion <http://www.akretion.com>.
```
pedrobaeza: Please don't put `TODAY`. You have to put `2022` or whatever, but you don't do things in the future.

pedrobaeza (approved):
For the delivery_carrier_account

lmignon:
Why a merge now. The initial implementation has been changed and my review is not for the current implementation. If the design of an addon is changed after an approval, this approval is no more valid.

pedrobaeza:
I didn't know that, sorry. I see 2 approvals, and the module split was already done, and it has been around a week.

lmignon:
But you requested changes to the data model. Therefore you should at least ping the previous reviewers to know if they agree with these changes. Your requested changes are not cosmetics nor typo changes.

pedrobaeza:
OK, anything more to comment, or is it just a rant?

lmignon:
This is not a rant, it's just a call for a little more collaboration on this type of decision that can impact the initial perception and call into question the current and past use cases or uses of the module.

pedrobaeza:
I haven't done it with intention, and it's noted for other times, but if finally nothing needs to be changed, why continue losing the time of both?

florian-dacosta:
Hi,
Sorry, my bad too, I should have pinged, before doing the changes, since there was a debate around the final solution.
I'll be more careful next time.
If anything is not good with the merged solution, let's discuss it and I am of course opened to make a new PR with possible changes we could all agree on.

pedrobaeza:
> If anything is not good with the merged solution, let's discuss it and I am of course opened to make a new PR with possible changes we could all agree on.

That's my point too.

lmignon:
> I haven't done it with intention, and it's noted for other times

:+1:

base_delivery_carrier_label/models/stock_picking.py
```diff
@@ -0,0 +1,190 @@
...
+            move_lines = picking.move_line_ids.filtered(
+                lambda s: not s.result_package_id
+            )
+            if move_lines:
+                picking._put_in_pack(move_lines)
+
+    def send_to_shipper(self):
+        self.ensure_one()
+        if self.env.context.get("set_default_package", True):
+            self._set_a_default_package()
```
rousseldenis, line 69: @florian-dacosta @glitchov Could you help me and explain why we need to set a new package ? For what purpose ? Thanks

rousseldenis: Ok, I see that's because we want to store the tracking on package side. But that should be optional IMHO and let by default the tracking on picking side.

florian-dacosta: @rousseldenis 
I think, the real goal was to ease the carrier modules. 
There were/are sometimes module with quite a lot of code to manage the case where we don't have any package on the picking and we have to generate a label, and the case where we have packages and so we have to generate a label with a pack... 
I guess the idea was kind of : the pack should be mandatory when generating a carrier label, but we don't want to force the user that don't need to use pack in Odoo to fill the pack on each picking

In reality there is always at least a pack, it is just not filled in Odoo for some users. And the carrier webservice usually expects some data relative to a pack.
I know the delivery_roulier is a lot simpler knowing it does not have to manage products outside a pack for generating carrier labels.
But... I guess it can be an issue, mainly in the case we don't want to manage pack in Odoo and we have action after the delivery (return or interwarehouse shipments, etc...)

rousseldenis: Thanks for your answer. I think I need to dig a little bit in this in order to identify corner cases.

PR #572, opened 2022-12-01
[15.0][FIX] delivery_cttexpress: translation keywords

chienandalu:
Dumb error

cc @Tecnativa

PR #573, opened 2022-12-01
[16.0][FIX] delivery_cttexpress: translation keywords

chienandalu:
fw of #572 

cc @Tecnativa


pedrobaeza (approved):
Straight-forward fw-port, so fast-tracking it:

/ocabot merge patch

PR #576, opened 2022-12-11
[14.0][ADD] delivery_line_qty_delivered

mt-software-de:
This modules changes the Delivered Quantity (qty_delivered) of a delivery sale.order.line
if one other sale.order.line is delivered
the qty_delivered of the delivery line is changed to its Quantity (product_uom_qty)

cc @jbaudoux


delivery_line_qty_delivered/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2022 Michael Tietz (MT Software) <mtietz@mt-software.de>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "delivery_line_qty_delivered",
+    "summary": "Changes the Delivered Quantity (qty_delivered) of a delivery sale.order.line"
+    "if one other sale.order.line is delivered the qty_deliverd of the delvivery line"
```
jbaudoux:
```suggestion
    "if one other sale.order.line is delivered the qty_delivered of the delivery line"
```

delivery_line_qty_delivered/models/sale_order_line.py
```diff
@@ -0,0 +1,31 @@
...
+        "order_id.delivery_state",
+    )
+    def _compute_qty_delivered(self):
+        super(SaleOrderLine, self)._compute_qty_delivered()
+        for line in self.filtered(lambda sol: sol.qty_delivered_method == "delivery"):
+            if (
+                line.product_id.invoice_policy == "delivery"
+                and line.order_id.delivery_state in ["done", "partially"]
+            ):
+                line.qty_delivered = line.product_uom_qty
```
jbaudoux: I would put that in a method `_get_delivery_qty_delivered()` as some may want a different value

delivery_line_qty_delivered/models/sale_order_line.py
```diff
@@ -0,0 +1,31 @@
...
+        for line in self:
+            if line._is_delivery() and line.product_id.type == "service":
+                line.qty_delivered_method = "delivery"
+
+    @api.depends(
+        "order_id.delivery_state",
+    )
+    def _compute_qty_delivered(self):
+        super(SaleOrderLine, self)._compute_qty_delivered()
+        for line in self.filtered(lambda sol: sol.qty_delivered_method == "delivery"):
```
jbaudoux: Put the check inside the loop to prevent double loop

delivery_line_qty_delivered/models/sale_order_line.py
```diff
@@ -0,0 +1,31 @@
...
+
+    @api.depends("product_id", "is_delivery")
+    def _compute_qty_delivered_method(self):
+        super(SaleOrderLine, self)._compute_qty_delivered_method()
+        for line in self:
+            if line._is_delivery() and line.product_id.type == "service":
+                line.qty_delivered_method = "delivery"
+
+    @api.depends(
+        "order_id.delivery_state",
```
jbaudoux: I think you should not add this depend otherwise you are creating a dependency loop between `_compute_qty_delivered` and `_compute_sale_delivery_state`

delivery_line_qty_delivered/models/sale_order.py
```diff
@@ -0,0 +1,16 @@
...
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _compute_sale_delivery_state(self):
+        super()._compute_sale_delivery_state()
+        for sale in self:
+            delivery_lines = sale.order_line.filtered(lambda line: line._is_delivery())
+            if delivery_lines:
+                sale.modified(["delivery_state"])
```
jbaudoux: Have you tried flagging `qty_delivered_manual` as modified instead?

jbaudoux (changes requested):
The dependencies seems to cause a non deterministic result

delivery_line_qty_delivered/models/sale_order_line.py
mt-software-de: Yes i know, but what i need is a dependency on each other line's qty_delivered

mt-software-de: What i can do is removing the dependency completely, and trigger the computation from sale.order

mt-software-de:
> The dependencies seems to cause a non deterministic result

I removed the dependency. 
@jbaudoux  or @rousseldenis could you maybe have a look at tests?
Because the tests are running locally.

rousseldenis:
@mt-software-de I would try to install all the modules installed before this one (see log details) and see which one is modifying standard behavior.

mt-software-de:
> @mt-software-de I would try to install all the modules installed before this one (see log details) and see which one is modifying standard behavior.

Thxs, Unfortunately I have to do that.

mt-software-de:
This was now moved to https://github.com/OCA/sale-workflow/pull/2717

PR #577, opened 2022-12-12
[13.0][FIX] delivery_tnt_oca: Set correct weight if not using packages

victoralmau:
Set correct weight if not using packages

@Tecnativa TT40767

### DIFF
--- a/delivery_tnt_oca/models/tnt_request.py
+++ b/delivery_tnt_oca/models/tnt_request.py
@@ -162,7 +162,7 @@ class TntRequest(object):
             "LENGTH": data_total["length"],
             "HEIGHT": data_total["height"],
             "WIDTH": data_total["width"],
-            "WEIGHT": data_total["weight"],
+            "WEIGHT": data_total["weight"] / self.record.number_of_packages,
         }
 
     def _prepare_address(self, partner):
--- a/delivery_tnt_oca/tests/test_delivery_tnt.py
+++ b/delivery_tnt_oca/tests/test_delivery_tnt.py
@@ -5,6 +5,8 @@ from datetime import timedelta
 from odoo.exceptions import UserError
 from odoo.tests import Form, common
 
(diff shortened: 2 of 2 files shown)


delivery_tnt_oca/models/tnt_request.py
```diff
@@ -162,7 +162,11 @@ def _quant_package_data_from_picking(self):
             "LENGTH": data_total["length"],
             "HEIGHT": data_total["height"],
             "WIDTH": data_total["width"],
-            "WEIGHT": data_total["weight"],
+            "WEIGHT": (
+                data_total["weight"] / self.record.number_of_packages
```
pedrobaeza: Shouldn't it be always the division?

pedrobaeza:
But this has to be confirmed by the carrier that it should be that way, isn't it?

delivery_tnt_oca/models/tnt_request.py
victoralmau: Yes, you are right, because it is not sent in PACKAGE the package listing (if used).

pedrobaeza:
It's a bit weird that we send the weight per package in the totals.

pedrobaeza:
OK, confirmed:

```
En la request de EC el peso que vuelca a nuestros sistemas es el del apartado PACKAGE y este se debe especificar el peso total como peso total dividido por el número de ITEMS por ejemplo:

        <PACKAGE>
          <ITEMS>2</ITEMS>
          <DESCRIPTION>MAY\OUT\78475</DESCRIPTION>
          <LENGTH>0.35</LENGTH>
          <HEIGHT>0.35</HEIGHT>
          <WIDTH>0.35</WIDTH>
          <WEIGHT>10</WEIGHT>
        </PACKAGE>

En este caso el peso total del envío es 20 KG.
```

pedrobaeza (approved):

pedrobaeza:
Let's fast track it to fix the problem:

/ocabot merge patch

PR #578, opened 2022-12-12
[13.0][FIX] delivery_ups_oca: Fix cancel_shipment method

victoralmau:
Fix `cancel_shipment` method.

@Tecnativa TT40776

### DIFF
--- a/delivery_ups_oca/models/ups_request.py
+++ b/delivery_ups_oca/models/ups_request.py
@@ -32,6 +32,7 @@ class UpsRequest(object):
     def _process_reply(
         self, url, data=None, method="post", query_parameters=None,
     ):
+        data = data or {}
         headers = {
             "AccessLicenseNumber": self.access_license_number,
             "Username": self.username,


delivery_ups_oca/models/ups_request.py
```diff
@@ -32,6 +32,7 @@ def __init__(self, carrier):
     def _process_reply(
         self, url, data=None, method="post", query_parameters=None,
     ):
+        data = data if data else {}
```
pedrobaeza: Why this can be `None`?

```suggestion
        data = data or {}
```

victoralmau: When trying to cancel a shipment without set this parameter https://github.com/OCA/delivery-carrier/blob/13.0/delivery_ups_oca/models/ups_request.py#L257 i have seen this problem.

pedrobaeza (approved):

chienandalu (approved):

PR #579, opened 2022-12-12
[13.0][IMP] delivery_state: new state to ignore further update calls

chienandalu:
It could happen that a carrier deletes tracking numbers from its DB and those states are not longer updatable. We don't want to be checking them forever, so we want a state to ignore them in the future.

cc @Tecnativa TT40774

### DIFF
--- a/delivery_state/models/stock_picking.py
+++ b/delivery_state/models/stock_picking.py
@@ -24,6 +24,7 @@ class StockPicking(models.Model):
             ("incidence", "Incidence"),
             ("customer_delivered", "Customer delivered"),
             ("warehouse_delivered", "Warehouse delivered"),
+            ("no_update", "No more updates from carrier"),
         ],
         string="Carrier State",
         tracking=True,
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
I think this shouldn't be handled as state level, but an extra boolean flag for that, for not losing last state tracked.

And on the update cron, if an error is raised, to mark such flag.

And finally another thing that I'm remembering now: on the log, all the state are repeated each time a new one is got. Can this be avoided?

chienandalu:
> I think this shouldn't be handled as state level, but an extra boolean flag for that, for not losing last state tracked. And on the update cron, if an error is raised, to mark such flag.

For the case presented, there's no exception. Just a message reporting that the tracking number doesn't exists. But if we were to flag the pickings which delivery state update ends in an exception (e. g.: temporary API failures), we could be flagging wrongly those.

> And finally another thing that I'm remembering now: on the log, all the state are repeated each time a new one is got. Can this be avoided?

That's formatted for every carrier request class. This module only adds the method plugs.

pedrobaeza (approved):
OK, let's go this way.

victoralmau (approved):

PR #584, opened 2022-12-14
[15.0][IMP] delivery_cttexpress: force delivery_price_method dependency

chienandalu:
CTT Express doesn't have a delivery prices methods in their API. Until now, it was enforced in the instructions to add delivery_price_method to the instance in order to be able to have price rules. We force the dependecy so the module functionality is complete out of the box.

cc @Tecnativa TT40759

### DIFF
--- a/delivery_cttexpress/README.rst
+++ b/delivery_cttexpress/README.rst
@@ -36,12 +36,8 @@ Installation
 ============
 
 This module needs the `zeep` python library. It depends on the modules
-`delivery_package_number` and `delivery_state` that can be found on
-OCA/delivery-carrier.
-
-CTT Express Iberic Web Services API doesn't provide shipping price calculation methods.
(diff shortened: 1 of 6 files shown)


pedrobaeza (commented):
You need to change INSTALL.rst to accommodate lastest dependency

chienandalu:
This is how it is currently. Wouldn't it be enough?: https://github.com/OCA/delivery-carrier/blob/4fbfc2cc97ca8d3afa218e2edf46f19772097a85/delivery_cttexpress/readme/INSTALL.rst

pedrobaeza:
Well, now it's mandatory, so better to put everything in the first paragraph, and on CONFIGURE.rst mention about that new part.

chienandalu:
Done

pedrobaeza (approved):

victoralmau (approved):

PR #585, opened 2022-12-14
[14.0][IMP] delivery_state: new state to ignore further update calls

chienandalu:
fw of - #579 

It could happen that a carrier deletes tracking numbers from its DB and those states are not longer updatable. We don't want to be checking them forever, so we want a state to ignore them in the future.

cc @Tecnativa TT40774


pedrobaeza (approved):
We can fast-track this fw-port:

/ocabot merge minor

PR #594, opened 2022-12-20
[16.0][ADD] delivery_carrier_max_weight_constraint

glitchov:
Add after migration of a never merged 10.0 PR


lmignon (changes requested):
@glitchov IMO it's not the right approach. The check should be made into the `ChooseDeliveryPackage` wizard. (By default a warning is displayed when the weight exceed). Moreover, the weight to consider is not the weight of one quant but the weight of all the moves added to the package.
An other point to take into consideration is that you must avoid to modify the initial behavior globally by installing a new addon. If a module, modify the initial behavior, this change must be activable by explicitly setting a parameter. In this case,
I propose to add:
* an option into the stock settings asking if overweight package must be prevented, (by company) 
* a computed `readonly=false` field on the carrier `depends('company_id')` for the same purpose
* a computed `readonly=false` field on `stock.package.type`  `depends('package_carrier_type')` for the same purpose

Once it's done, you can add an `api.constraints` into the `ChooseDeliveryPackage` wizard checking if the max_weight is exceeded only if the check must be done for the given `delivery_package_type_id`

lmignon (approved):
LGTM (Code review only since no carrier integration is available)

delivery_carrier_max_weight_constraint/wizards/choose_delivery_package.py
```diff
@@ -0,0 +1,25 @@
...
+from odoo import _, api, models
+from odoo.exceptions import ValidationError
+
+
+class ChooseDeliveryPackage(models.TransientModel):
+    _inherit = "choose.delivery.package"
+
+    @api.constrains("delivery_package_type_id", "shipping_weight")
+    def _check_strict_weight_package(self):
+        for package in self:
```
lmignon, line 13: @glitchov IMO,if `delivery_package_type_id` is not set, you should take the value for ` is_strict_weight`on the company.

lmignon (changes requested):

delivery_carrier_max_weight_constraint/models/delivery_carrier.py
```diff
@@ -0,0 +1,25 @@
...
+
+    is_strict_weight_package = fields.Boolean(
+        string="Overweight package forbidden",
+        help="The maximum weight of packages can't be exceeded.",
+        compute="_compute_is_strict_weight_package",
+        store=True,
+        readonly=False,
+    )
+
+    @api.depends("company_id")
```
rousseldenis: Shouldn't it be : `company_id.delivery_carrier_strict_weight_package` ?

delivery_carrier_max_weight_constraint/models/stock_package_type.py
```diff
@@ -0,0 +1,28 @@
...
+        help="The maximum weight of the package can't be exceeded.",
+        compute="_compute_is_strict_weight",
+        store=True,
+        readonly=False,
+    )
+
+    @api.depends("package_carrier_type")
+    def _compute_is_strict_weight(self):
+        for rec in self:
+            carrier_id = self.env["delivery.carrier"].search(
```
rousseldenis: Is this really correct ? Shouldn't it be evaluated on picking level ?

If there are several carriers ?

UPDATE: I have answered myself. The field on carrier is evaluated on company level. But one question remains: why doing a search in a loop to do that? Performances won't be good. I think we can use the company field directly.

delivery_carrier_max_weight_constraint/__manifest__.py
```diff
@@ -1,23 +1,19 @@
...
 # Copyright 2021 ACSONE SA/NV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 {
     "name": "Constrain package maximum weight",
-    "version": "10.0.1.0.0",
-    "author": "ACSONE SA/NV",
+    "version": "16.0.1.0.0",
+    "author": "ACSONE SA/NV, Odoo Community Association (OCA)",
     "maintener": "ACSONE SA/NV",
```
rousseldenis: `"maintainers": ["glitchov"]`

rousseldenis (changes requested):
Some questions. One change

delivery_carrier_max_weight_constraint/models/stock_package_type.py
glitchov: ok

lmignon:
ping @rousseldenis Can you update your review

rousseldenis:
> ping @rousseldenis Can you update your review

@lmignon I can, but comments haven't been attended :sweat_smile:

lmignon:
@rousseldenis done....

rousseldenis (approved):
Code review

rousseldenis:
@lmignon Could you approve this ?

lmignon (approved):

PR #597, opened 2023-01-10
[16.0][IMP] stock_picking_delivery_link

glitchov:
Force the 'choose package type' wizard in put in pack at a former picking step if shipment picking has a carrier_id set.


glitchov:
Redundant

PR #603, opened 2023-01-25
[16.0][ADD] delivery_package_number_theoretical

glitchov:
Computes the amount of packages a picking out should have depending on the weight of the products and the limit fixed by the carrier.
It's fully independent of the delivery_package_number module.


glitchov:
bad branch

PR #604, opened 2023-01-25
[16.0][ADD] delivery_estimated_package_quantity_by_weight

glitchov:
This module computes the amount of packages a picking out should have depending on the weight of the products and the limit fixed by the carrier.
It's fully independent of the delivery_package_number module.


delivery_package_number_theoretical/models/delivery_carrier.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class DeliveryCarrier(models.Model):
+
+    _inherit = "delivery.carrier"
+
+    maximum_weight_per_package = fields.Float(string="Maximum weight per package")
```
lmignon: @glitchov As it's done for the product form, you should display the uom name of the uom defined for volumes...

```python
    weight_uom_name = fields.Char(string='Weight unit of measure label', compute='_compute_weight_uom_name')

    def _compute_weight_uom_name(self):
        self.weight_uom_name = self.env["product.template"]._get_weight_uom_name_from_ir_config_parameter()
```

delivery_package_number_theoretical/views/delivery_carrier_views.xml
```diff
@@ -0,0 +1,31 @@
...
+    </record>
+    <record model="ir.ui.view" id="delivery_carrier_tree_view">
+        <field
+            name="name"
+        >delivery.carrier.tree (in delivery_number_package_theoretical)</field>
+        <field name="model">delivery.carrier</field>
+        <field name="inherit_id" ref="delivery.view_delivery_carrier_tree" />
+        <field name="arch" type="xml">
+            <field name="delivery_type" position="after">
+                <field name="maximum_weight_per_package" />
```
lmignon:
```suggestion
                 <label for="weight"/>
                 <div class="o_row" name="maximum_weight_per_package">
                     <field name="maximum_weight_per_package" class="oe_inline"/>
                     <field name="weight_uom_name"/>
                 </div>
```

lmignon (changes requested):
@glitchov Thank you for this addon. Reviewed technically and functionally.  LGTM but we need to display the volume uom configured into odoo next to the `maximum_weight_per_package` field into the delivery.carrier form view to help the user to provide the right value.

lmignon (approved):
Thank you @glitchov LGTM (Code review + functional tests)

rousseldenis (approved):
Code review

delivery_package_number_theoretical/models/stock_picking.py
```diff
@@ -0,0 +1,113 @@
...
+            ):
+                rec.is_number_of_packages_visible = True
+            else:
+                rec.is_number_of_packages_visible = False
+
+    @api.depends("is_number_of_packages_visible", "move_ids")
+    def _compute_theoretical_number_of_packages(self):
+        for rec in self:
+            if rec.is_number_of_packages_visible:
+                products_weights = rec.move_ids.mapped("product_id.weight")
```
jbaudoux: Add in README that product packagings are not supported

delivery_package_number_theoretical/models/stock_picking.py
```diff
@@ -0,0 +1,113 @@
...
+            theoretical_number_of_packages += 1
+            # Try to fit the heaviest product with the lightest.
+            # If it does not work, then the heaviest should have
+            # a box to himself
+            weight = products_weights_list[i] + products_weights_list[j]
+            while weight <= maximum_weight_per_package:
+                i += 1
+                if i < j:
+                    # While the weight of products does not exceed the limit,
+                    # continue adding products in the same package
```
jbaudoux: Add in README the chosen strategy

delivery_package_number_theoretical/models/delivery_carrier.py
```diff
@@ -0,0 +1,21 @@
...
```
jbaudoux: Add in README that you don't support multiple delivery packages (stock.package.type) with different max weight for a shipping method

jbaudoux (commented):
I find this module quite specific, especially that a specific strategy is applied on a max weight.
I find the module name misleading. Maybe rename to delivery_estimated_package_quantity_by_weight

PR #614, opened 2023-02-28
[15.0][MIG] delivery_auto_refresh: Migration to version 15.0

pilarvargas-tecnativa:
cc @Tecnativa TT36943

@CarlosRoca13 @chienandalu please review :)


delivery_auto_refresh/data/ir_config_parameter.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2018 Tecnativa - Pedro M. Baeza
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo noupdate="1">
+    <record id="set_default_carrier" model="ir.config_parameter">
```
chienandalu: Let's improve the design to set it via settings using a field with the `config_parameter` parameter. For example: https://github.com/odoo/odoo/blob/15.0/addons/base_setup/models/res_config_settings.py#L16

pilarvargas-tecnativa: I'm working on it

delivery_auto_refresh/models/sale_order.py
```diff
@@ -32,6 +32,7 @@ def _onchange_partner_id(self):
                         lambda x: x in order.available_carrier_ids._origin
                     )
                 )
+        return
```
chienandalu: If this is about the precommit warning this is just a trick with no effect. The proper way is to assign the `super` call to a variable that you're going to return at a given point.

delivery_auto_refresh/models/sale_order.py
```diff
@@ -101,6 +102,7 @@ def _remove_delivery_line(self):
         current_carrier = self.carrier_id
         super()._remove_delivery_line()
         self.carrier_id = current_carrier
+        return
```
chienandalu: Same here

delivery_auto_refresh/tests/test_delivery_auto_refresh.py
```diff
@@ -71,7 +71,7 @@ def setUpClass(cls):
         cls.param_name2 = "delivery_auto_refresh.refresh_after_picking"
         cls.param_name3 = "delivery_auto_refresh.auto_void_delivery_line"
         cls.param_name4 = "delivery_auto_refresh.set_default_carrier"
-        cls.env["ir.config_parameter"].sudo().set_param(cls.param_name4, 1)
+        cls.env["ir.config_parameter"].sudo().set_param(cls.param_name4, True)
```
chienandalu: Why not setting the res.config.settings field to do the full case?

delivery_auto_refresh/utils.py
```diff
@@ -7,4 +7,4 @@
 def get_bool_param(env, name, default="0"):
```
chienandalu: This tool should go away I think. Let's make it work without this

chienandalu (changes requested):

pilarvargas-tecnativa:
@chienandalu changes made, thank you

delivery_auto_refresh/models/sale_order.py
```diff
@@ -18,8 +16,13 @@ class SaleOrder(models.Model):
...
         if hasattr(super(), "_onchange_partner_id"):
-            super()._onchange_partner_id()
-        if get_bool_param(self.env, "set_default_carrier"):
+            return super()._onchange_partner_id()
+        default_carrier = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("delivery_auto_refresh.set_default_carrier")
+        )
+        if default_carrier:
```
chienandalu: Let's name these variables after their params
```suggestion
        set_default_carrier = (
            self.env["ir.config_parameter"]
            .sudo()
            .get_param("delivery_auto_refresh.set_default_carrier")
        )
        if set_default_carrier:
```

delivery_auto_refresh/models/sale_order.py
```diff
@@ -0,0 +1,151 @@
...
+        # When we have the context 'website_id' it means that we are doing the order from
+        # e-commerce. So we don't want to add the delivery line automatically.
+        if self.env.context.get("website_id"):
+            return False
+        auto_delivery = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("delivery_auto_refresh.auto_add_delivery_line")
+        )
+        return auto_delivery
```
chienandalu: There's no need to assign a variable that we're immediately returning back
```suggestion
        return = (
            self.env["ir.config_parameter"]
            .sudo()
            .get_param("delivery_auto_refresh.auto_add_delivery_line")
        )
```

delivery_auto_refresh/models/stock_picking.py
```diff
@@ -3,16 +3,19 @@
...
-
 
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
     def _add_delivery_cost_to_so(self):
         """Update delivery price in SO from picking data."""
         res = super()._add_delivery_cost_to_so()
-        if not get_bool_param(self.env, "refresh_after_picking"):
+        refresh_picking = (
```
chienandalu: Name it after the param as well

delivery_auto_refresh/models/stock_picking.py
```diff
@@ -53,7 +56,12 @@ def _action_done(self):
         # If configured, we want to set to 0 automatically the delivery line
         # when we have a returned picking that isn't invoiced so we don't have
         # it as invoiceable line. Otherwise, the salesman has to do it by hand.
-        if not get_bool_param(self.env, "auto_void_delivery_line"):
+        auto_void = (
```
chienandalu: Same as above

chienandalu (changes requested):
Some slight changes :)

delivery_auto_refresh/models/stock_picking.py
pilarvargas-tecnativa: thanks :)

delivery_auto_refresh/models/sale_order.py
```diff
@@ -18,8 +16,13 @@ class SaleOrder(models.Model):
     @api.onchange("partner_id", "partner_shipping_id")
     def _onchange_partner_id(self):
         if hasattr(super(), "_onchange_partner_id"):
-            super()._onchange_partner_id()
-        if get_bool_param(self.env, "set_default_carrier"):
+            return super()._onchange_partner_id()
```
chienandalu: This changes the behavior of the code that doesn't expect to return at this point.

CarlosRoca13 (approved):

chienandalu (changes requested):

delivery_auto_refresh/models/sale_order.py
pilarvargas-tecnativa: corrected, thanks :)

pedrobaeza (approved):
/ocabot merge nobump

PR #619, opened 2023-03-13
[16.0][MIG] delivery_postlogistics: Migration to 16.0

lehoangan:


StephaneMangin (approved):
LGTM  Well spotted about the context propagation ;)

delivery_postlogistics/models/stock_package_type.py
```diff
@@ -4,8 +4,8 @@
 from odoo import fields, models
 
 
-class ProductPackaging(models.Model):
-    _inherit = "product.packaging"
+class PackageType(models.Model):
+    _inherit = "stock.package.type"
```
simahawk, line 8: shouldn't we have a migration step to migrate current data from packaging to package type?

lehoangan: `package_carrier_type` is an Odoo native field from Odoo 15, I don't think we need to migrate data

simahawk (approved):

PR #622, opened 2023-03-24
[FIX] delivery_schenker: fix missing dependency

mt-software-de:
before there was no volume field for stock.picking
 which is used in 
 https://github.com/OCA/delivery-carrier/blob/efd269cdda33696b011f13d1a17f99b1027e7dbe/delivery_schenker/models/delivery_carrier.py#L378
 and
 https://github.com/OCA/delivery-carrier/blob/efd269cdda33696b011f13d1a17f99b1027e7dbe/delivery_schenker/models/delivery_carrier.py#L406
 
 depends on https://github.com/OCA/stock-logistics-warehouse/pull/1675


rousseldenis:
@mt-software-de Don't hesitate to include the PR dependency in test-requirements.txt in order to be able to review

mt-software-de:
> @mt-software-de Don't hesitate to include the PR dependency in test-requirements.txt in order to be able to review

Thx, could you share with me how to include it in the test-requirements.txt?

rousseldenis:
> > @mt-software-de Don't hesitate to include the PR dependency in test-requirements.txt in order to be able to review
> 
> Thx, could you share with me how to include it in the test-requirements.txt?

For sure: https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29

mt-software-de:
> > > @mt-software-de Don't hesitate to include the PR dependency in test-requirements.txt in order to be able to review
> > 
> > 
> > Thx, could you share with me how to include it in the test-requirements.txt?
> 
> For sure: https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29

Thanks. 
I added the PR.

pedrobaeza (changes requested):
You shouldn't modify the dependencies of a module in a stable version. Create an extra module for that `delivery_schenker_picking_volume` or similar.

mt-software-de:
> You shouldn't modify the dependencies of a module in a stable version. Create an extra module for that `delivery_schenker_picking_volume` or similar.

But the version of this module is not stable because of the missing volume field in stock.picking.
If you install this module on an empty database, there is no volume field for stock.picking and this raises at the described lines an error.

pedrobaeza:
Then remove such field from the communication

mt-software-de:
> Then remove such field from the communication

But this would change the behavior of this module, i don't think this would be good for someone who is using it already.
And also schenker declares this field as mandatory.
image

pedrobaeza:
If it's giving error, there's no such "regression".

mt-software-de:
> If it's giving error, there's no such "regression".

How we do continue here?

pedrobaeza:
I would compute the volume on the fly from the products for the main module, and create the other module for advanced volume management.

mt-software-de:
@chienandalu and @hildickethan-S73 
What do you think about it, since the module was developed by you. Also, how do you use this module without an existing volume field on the picking?

hildickethan:
> @chienandalu and @hildickethan-S73 What do you think about it, since the module was developed by you. Also, how do you use this module without an existing volume field on the picking?

:thinking: Hard to say exactly because it's been a while, but I think what happened here was that the volume field exists in 13.0 where the module was developed and when I migrated it to 14.0, I didn't stumble over this clearly missing field because the Odoos I've used it with would've had some other private module that adds a volume field, since it is a somewhat easy thing to want.

If this is the first time someone has run into this problem, it means that no one else has used this module in 14.0 or they also have some private volume field :sweat_smile: 

This module definitely should have either the dependency to use it or in it's absence, a basic sum of the products volume * product quantity

To remedy this and follow the "no dependencies to stable" rule it's what Pedro said, it's best to extract wherever `picking.volume` is used to a seperate method, have the sum of product volume be there in this module and then in the new "glue" module, use `picking.volume` in it's place

mt-software-de:
i haded a deeper look in the code, the volume is already computed, 
i refactored some lines to use the already calculated volume.
As well i created the glue module

delivery_schenker/models/delivery_carrier.py
```diff
@@ -165,6 +166,18 @@ class DeliveryCarrier(models.Model):
         help="If not delivery package or the package doesn't have defined the packaging"
         "it will default to this type",
     )
+    schenker_address_number = fields.Char(
+        "Address ID",
+        help="ID assigned by Schenker to you.\nWill be part of the sender or "
+        "if set, the invoice address.\nIf you don't want to send it, use 0",
```
pedrobaeza, line 172: According to the help:

```suggestion
        "if set, the invoice address.\nIf you don't want to send it, use 0",
        default=0,
```

delivery_schenker/models/delivery_carrier.py
```diff
@@ -241,20 +264,19 @@ def _prepare_schenker_address(
...
-            vals["email"] = partner.email
-        if partner.mobile:
-            vals["mobilePhone"] = partner.mobile
-        if partner.phone:
-            vals["phone"] = partner.phone
-        if partner.street2:
-            vals["street2"] = partner.street2
+        for schenker_key, expression in self._schenker_address_optional_fields():
+            fields = expression.split(".")
+            value = getattr(partner, fields[0])
```
pedrobaeza: Why doing it in 2 steps? It can be in the same loop.

delivery_schenker/models/delivery_carrier.py
```diff
@@ -269,10 +291,25 @@ def _schenker_shipping_address(self, picking):
             or picking.company_id.partner_id
         )
         consignee_address = picking.partner_id
-        return [
-            self._prepare_schenker_address(shipper_address, "SHIPPER"),
-            self._prepare_schenker_address(consignee_address),
+        shipper_address = self._prepare_schenker_address(shipper_address, "SHIPPER")
+        consignee_address = self._prepare_schenker_address(consignee_address)
+
```
pedrobaeza: Please avoid empty lines inside a method.

delivery_schenker/models/delivery_carrier.py
```diff
@@ -269,10 +291,25 @@ def _schenker_shipping_address(self, picking):
...
+            shipper_address,
+            consignee_address,
         ]
+        invoice_address = False
+        if self.schenker_partner_invoice_id:
+            invoice_address = self._prepare_schenker_address(
+                self.schenker_partner_invoice_id, "INVOICE"
+            )
+            result.append(invoice_address)
+
```
pedrobaeza: The same

delivery_schenker/views/delivery_schenker_view.xml
```diff
@@ -1,13 +1,14 @@
...
+Copyright 2023 Michael Tietz (MT Software) <mtietz@mt-software.de>
+License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
 <odoo>
-    <record id="view_delivery_carrier_form" model="ir.ui.view">
-        <field name="model">delivery.carrier</field>
-        <field name="inherit_id" ref="delivery.view_delivery_carrier_form" />
-        <field name="arch" type="xml">
-            <xpath expr="//notebook/page[1]" position='before'>
-                <page
+        <record id="view_delivery_carrier_form" model="ir.ui.view">
```
pedrobaeza: Remove this incorrect extra indentation

delivery_schenker/views/delivery_schenker_view.xml
```diff
@@ -60,15 +61,21 @@
...
-                        <group string="Logistics">
-                            <field
-                                name="schenker_default_packaging_id"
-                                attrs="{'required': [('delivery_type', '=', 'schenker')]}"
-                            />
-                            <field
-                                name="schenker_measure_unit"
-                                attrs="{'required': [('delivery_type', '=', 'schenker')]}"
-                            />
+                        <group>
```
pedrobaeza, line 64: You don't need this extra group AFAIK. Just add the other group.

delivery_schenker_picking_volume/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This is a glue module between delivery_schenker and stock_picking_volume
```
pedrobaeza, line 1: You can explain a bit more the motivation of this module.

pedrobaeza (changes requested):

delivery_schenker/views/delivery_schenker_view.xml
mt-software-de: Without this extra group the new block would be shown on left side, this would be resulting in a big empty block on the right side.

pedrobaeza: Well, I see it as no problem, but OK if you want to leave it that way.

hildickethan:
Could you rebase with the pre-commit changes?

hildickethan (approved):

delivery_schenker/models/delivery_carrier.py
mt-software-de: Sorry i don't know what you mean by that

mt-software-de:
> Could you rebase with the pre-commit changes?

done

pedrobaeza:
Can you split the PR for avoiding the dependency over the other module?

mt-software-de:
> Can you split the PR for avoiding the dependency over the other module?

The other PR is nearly done and should be mergeable today.

pedrobaeza:
I still think that it should be other PR for the glue module.

delivery_schenker/models/delivery_carrier.py
mt-software-de: ping @pedrobaeza :-)

pedrobaeza: You are repeating operations for index 0 and then a loop for [1:]. Do everything in the same loop for code readability and for saving lines.

pedrobaeza (approved):
/ocabot merge minor

PR #623, opened 2023-03-24
[ADD] stock_picking_delivery_package

rousseldenis:
This module allows to choose a delivery package type on other pickings than the delivery one (e.g.: pack). Customizable on operation type level.

This has been introduced because: some maybe want to add the delivery package type sooner (that's the important thing) than on delivery step if output process is at least using Pick + Ship flow.


rousseldenis:
In fact, the module stock_picking_delivery_link has to be enhanced... So, closing in favor of that one

PR #625, opened 2023-03-27
[14.0][IMP] delivery_price_rule_untaxed: fix website in manifest

Aldeigja:
fix website in manifest


rousseldenis:
Trivial.

/ocabot merge patch

@Aldeigja FYI, you could have added a slash at the end of the first line too.

PR #633, opened 2023-04-28
[FIX] delivery_schenker: Bugfix volume calculation

mt-software-de:
Sorry there were two bugs in the code from https://github.com/OCA/delivery-carrier/pull/622. 

I think we should add a simple test for it, :-) I am trying to do this today. 

cc @pedrobaeza @hildickethan-S73

### DIFF
--- a/delivery_schenker_picking_volume/models/delivery_carrier.py
+++ b/delivery_schenker_picking_volume/models/delivery_carrier.py
@@ -6,7 +6,7 @@ from odoo import models
 class DeliveryCarrier(models.Model):
     _inherit = "delivery.carrier"
 
-    def _schenker_shipping_information(self, picking):
-        res = super()._schenker_shipping_information(picking)
-        res["volume"] = picking.volume
-        return res
+    def _prepare_schenker_shipping(self, picking):
+        vals = super()._prepare_schenker_shipping(picking)
+        vals["shippingInformation"]["volume"] = picking.volume
+        return vals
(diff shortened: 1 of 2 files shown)


delivery_schenker_picking_volume/models/delivery_carrier.py
```diff
@@ -6,7 +6,7 @@
 class DeliveryCarrier(models.Model):
     _inherit = "delivery.carrier"
 
-    def _schenker_shipping_information(self, picking):
-        res = super()._schenker_shipping_information(picking)
-        res["volume"] = picking.volume
-        return res
+    def _prepare_schenker_shipping(self, picking):
+        vals = super()._schenker_shipping_information(picking)
```
hildickethan:
```suggestion
    def _prepare_schenker_shipping(self, picking):
        vals = super()._prepare_schenker_shipping(picking)
```

delivery_schenker_picking_volume/models/delivery_carrier.py
```diff
@@ -6,7 +6,7 @@
...
+        vals["volume"] = picking.volume
```
hildickethan:
```suggestion
        vals["shippingInformation"]["volume"] = picking.volume
```

mt-software-de: thanks

delivery_schenker_picking_volume/models/delivery_carrier.py
mt-software-de: :-) thanks

hildickethan (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #634, opened 2023-05-02
[13.0][FIX] delivery_multi_destination: Avoid error in creation of a multiple carrier with UX

victoralmau:
Avoid error in creation of a multiple carrier with UX

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT42837


CarlosRoca13 (approved):

pedrobaeza (approved):
This doesn't work for carriers created by code, but it's not the usual case

/ocabot merge patch

PR #637, opened 2023-05-03
[14.0][FIX] delivery_schenker: fix default of schenker_address_number

mt-software-de:


hildickethan:
what does this change?

mt-software-de:
> what does this change?

It removes the default value 0 from the new field schenker_address_number.
https://github.com/OCA/delivery-carrier/pull/622#discussion_r1166335778

This is wrong, because now the default value would be "0" and this is not False
like it is used here https://github.com/OCA/delivery-carrier/blob/a73f51da57b59556781eac110bbd217cbb62ed0c/delivery_schenker/models/delivery_carrier.py#L305

hildickethan:
> This is wrong, because now the default value would be "0" and this is not False like it is used here
> 
> https://github.com/OCA/delivery-carrier/blob/a73f51da57b59556781eac110bbd217cbb62ed0c/delivery_schenker/models/delivery_carrier.py#L305

0 should count as False for these sort of conditions so the functionality shouldn't change, but I do agree that a normal False is better suited for a case like this

hildickethan (approved):

mt-software-de:
> > This is wrong, because now the default value would be "0" and this is not False like it is used here
> > https://github.com/OCA/delivery-carrier/blob/a73f51da57b59556781eac110bbd217cbb62ed0c/delivery_schenker/models/delivery_carrier.py#L305
> 
> 0 should count as False for these sort of conditions so the functionality shouldn't change, but I do agree that a normal False is better suited for a case like this

Yes but it is not 0 the default value would be "0"
Because the field is a Char. 
https://github.com/OCA/delivery-carrier/blob/a73f51da57b59556781eac110bbd217cbb62ed0c/delivery_schenker/models/delivery_carrier.py#L169
image

As you can see it here on the bottom right.

hildickethan:
I see, that makes sense. I would rebase for the tests since they failed due to some other modules API, see if it works this time

mt-software-de:
> I see, that makes sense. I would rebase for the tests since they failed due to some other modules API, see if it works this time

Should i maybe move this one to draft, and add also unittests for it.

hildickethan:
If it is within your time budget, more test coverage is always welcome

mt-software-de:
> If it is within your time budget, more test coverage is always welcome

I added some simple tests

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
Oh, wait, the fail is in your PR. Please check.

mt-software-de:
> Oh, wait, the fail is in your PR. Please check.

Yes i know

mt-software-de:
> Oh, wait, the fail is in your PR. Please check.

Now i know the problem. 
In the test workflow run, the module base_delivery_carrier_label is getting installed before delivery_schenker.
And this is here the problem the base_delivery_carrier_label is creating an default package. And this is causing the error.
I can fix it within the test by deleting the previous created package, but i don't think it should be like that.

pedrobaeza:
You can make sure in your test that no package is assigned. Is that possible?

mt-software-de:
> You can make sure in your test that no package is assigned. Is that possible?

Yes i can do this. But i really don't like this at all.

delivery_schenker/tests/test_delivery_schenker.py
```diff
@@ -0,0 +1,193 @@
...
+        cls.sale, cls.picking = cls._create_sale_order(cls)
+        cls._process_picking(cls, cls.picking)
+
+    def _process_picking(self, picking):
+        picking.move_lines.quantity_done = 1
+        picking.button_validate()
+        # Delete result package, otherwise Unittests will fail
+        # because the addon base_delivery_carrier_label gets installed first
+        # which creates a default package within the method _set_a_default_package
+        picking.move_line_ids.result_package_id = False
```
hildickethan, line 69: looking at `base_delivery_carrier_label` code, if you send `set_default_package` context to False in the validate it should avoid this

mt-software-de: Yes i know, but i dont Like it to add Logic from an other module which is not in the dependency list. We only have to add this in the Tests because of the oca unitests behavior. If you Test it separate everything works fine.

mt-software-de:
@pedrobaeza could you rerun the Tests? And merge it then?

PR #638, opened 2023-05-10
[13.0][FIX] delivery_multi_destination: Get properly subcarrier price on based on rules

pedrobaeza:
If the destination carrier line is based on rules, the price is not correctly fetched, as it's hardcoded to call `_get_price_available` using picking's carrier, no matter the recordset from which you call it (the self argument).

Thus, the only solution to get the proper value is to temporarily replace the carrier on the picking on the calls chain, to restore it before returning.

@Tecnativa TT42862

### DIFF
--- a/delivery_multi_destination/models/delivery_carrier.py
+++ b/delivery_multi_destination/models/delivery_carrier.py
@@ -98,12 +98,19 @@ class DeliveryCarrier(models.Model):
                             break
                     else:
                         try:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@carmenbianca I think this is the solution for the problem you have in #574

chienandalu (approved):

victoralmau (approved):

PR #639, opened 2023-05-15
[14.0][MIG] delivery_ups_oca

HviorForgeFlow:
Standard migration

CC @ForgeFlow @JoanSForgeFlow


JordiBForgeFlow:
@victoralmau I can see that in v13 you proposed this logic https://github.com/OCA/delivery-carrier/commit/fec0e8fb66d41ba7d08d4980a0ca24576de43445#diff-b7c04acb978562ba955ae08facf529d4960361a2ffbd8db950468bd62a94993aR93 and I am now reverting in https://github.com/OCA/delivery-carrier/pull/639/commits/807d9037b45ab030c5cc26e7e9c57e443c9e3b88. I am finding that UPS responds with a GIF format, and the previous code was renaming the file to <tracking number>-GIF.txt, which does not make much sense. Perhaps there's a reason for putting .txt when the response is not PDF?

victoralmau:
> 

Remembering about this (not completely sure), that this was done so that when you click on the "_UPS Label_" button the file that is generated is a `.txt `so that when printed on a zebra printer it prints correctly (a .gif does not print it). I think this was happening because the response it gave when creating the shipment and getting the label (after) was not exactly the same.

Anyway, the correction you have made seems to make sense.

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -9,7 +10,12 @@
 class DeliveryCarrier(models.Model):
     _inherit = "delivery.carrier"
 
-    delivery_type = fields.Selection(selection_add=[("ups", "UPS")])
+    delivery_type = fields.Selection(
+        selection_add=[("ups", "UPS")],
+        ondelete={
+            "ups": lambda recs: recs.write({"delivery_type": "fixed", "fixed_price": 0})
```
victoralmau, line 16: I am not sure if this is necessary and/or correct.

victoralmau (commented):
Code review OK.

delivery_ups_oca/models/ups_request.py
```diff
@@ -0,0 +1,287 @@
...
+
+    def cancel_shipment(self, pickings):
+        url = "%s/ship/v1/shipments/cancel" % self.url
+        for item in pickings.filtered(lambda x: x.carrier_tracking_ref):
+            url = "{}/{}".format(url, item.carrier_tracking_ref)
+            status = self._process_reply(url=url, method="delete")
+            self._raise_for_status(status, False)
+        return True
+
+    def tracking_state_update(self, picking):
```
manuelregidor, line 266: This seems not to be working with production mode (it is working in test mode though). For some reason, the API sends an error. We have got in touch with UPS to try to sort it out, but we are still waiting for a solution. Are you dealing with this issue too? In that case, we can try to work together on this so it can be solved quickly. Thank you.

HviorForgeFlow: Regarding UPS documentation https://developer.ups.com/api/reference?loc=en_US#operation/getSingleTrackResponseUsingGET it seems that Oauth is required now. That's why maybe it is working on sandbox environment but not in production.

@manuelregidor 

https://developer.ups.com/api/reference/oauth/authorization-code?loc=en_US

manuelregidor: OK, thank you.

manuelregidor:
@HviorForgeFlow Are you planning to continue with this migration? Thank you.

HviorForgeFlow:
@manuelregidor I don't think so, but you have a review here if you go ahead with this migration.

PR #642, opened 2023-05-23
[16.0][ADD] delivery_automatic_package

rousseldenis:


delivery_package_automatic/models/res_company.py
```diff
@@ -0,0 +1,14 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+
+    _inherit = "res.company"
+
+    automatic_package_creation_default = fields.Boolean(
```
lmignon: s/automatic_package_creation_default/automatic_package_creation_at_delivery_default

delivery_package_automatic/models/res_config_settings.py
```diff
@@ -0,0 +1,14 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+
+    _inherit = "res.config.settings"
+
+    automatic_package_creation_default = fields.Boolean(
```
lmignon: s/automatic_package_creation_default/automatic_package_creation_at_delivery_default

lmignon (changes requested):
LGTM but I would use a field name that better reflect the context where it's used.

delivery_package_automatic/models/res_company.py
rousseldenis: Done

delivery_package_automatic/models/res_config_settings.py
rousseldenis: Done

delivery_automatic_package/models/delivery_carrier.py
```diff
@@ -0,0 +1,15 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class DeliveryCarrier(models.Model):
+
+    _inherit = "delivery.carrier"
+
+    automatic_package_creation_at_delivery = fields.Boolean(
```
bealdav, line 11: Nitpicking
Python code makes things automatic, then `automatic`  have no high value here and make field name longer useless.
```suggestion
    package_creation_at_delivery = fields.Boolean(
```

rousseldenis: Yes, but that reflects a change from standard behavior.

Apart that, are you ok with the change ?

lmignon (approved):
LGTM (Code review)

bealdav (approved):
Thanks a lot for this contribution, LGTM code review.

My own preference is to have only meaningful words in fields to make it shorter, but that's just my taste.

Then that's an unconditionally approve

florian-dacosta (approved):
Thanks a lot for this PR, I think it is a very good thing to make this feature in a separate module so it becomes optional.
LGTM code review

bguillot:
Thanks @rousseldenis !

PR #643, opened 2023-05-23
[IMP] base_delivery_carrier_label: Externalize automatic package creation

rousseldenis:
Depends on :

- [x] #642 

I've added delivery_postlogistics in rebel modules as it overrides base_delivery_carrier_label methods without calling super. @lehoangan @simahawk


rousseldenis:
@florian-dacosta Following our discussion about this.

I've added a separate module in order to isolate the functionnality and did it configurable.

So, this is a breaking change from former behaviour but easily configurable (see module README)

@bealdav

lmignon (approved):
LGTM (Code review)

florian-dacosta (commented):
@rousseldenis I wonder if we really want all this..

I mean, `base_delivery_carrier_package` has now no link at all with `delivery_automatic_package`
In the tests, you actually just test that `delivery_automatic_package` works, we could do the same in `delivery` module or any module depending on it. It is quite strange to add these tests in this module when there is really no link.

I understand the initial purpose of course, which would be not to remove a functionality from the module out of nowhere, but even with the dependency is is a breaking change, since configured with no package by default.

Couldn't we just remove drop the tests and the dependency on `delivery_automatic_package` and just remove `_set_a_default_package` method in this PR?
It would still be a breaking change quite easily solved with the intallation of the module... I mean, since the user will need to be aware of this change to make it continue to work...

Moreover, this feature is probably not really use yet in v16, at least willingly, for now...

IMO, the new module is mainly usefull for carrier modules which needs pack to work, and for customer that' don't want to manage PACK when there is only one PACK by picking, so it should be a dependency of those module but not of base_delivery_carrier_label.
For instance, delivery_roulier needs packs on the picking to be able to generate the labels. This one should/may depend on `delivery_automatic_package`. And in fact, I  think I won't add this dependency in delivery_roulier, I'd probably prefer to raise in case there is no pack and willingly install/configure delivery_automatic_package for customers that need it.

What do you think?

florian-dacosta:
@rousseldenis Any news ? 
I  really think it is strange to add test on this module for a feature added in a submodule, but if you think it is important, I'll approve as it is. Also, can you rebase?

sbidoul:
@rousseldenis what's the status of this one?

rousseldenis:
> @rousseldenis what's the status of this one?

I'll fix this now

delivery_roulier/tests/test_delivery_roulier.py
```diff
@@ -55,6 +55,7 @@ def setUp(self):
                 "delivery_type": "test",
                 "product_id": delivery_product.id,
                 "carrier_account_id": self.account.id,
+                "automatic_package_creation_at_delivery": True,
```
rousseldenis, line 58: @florian-dacosta I've updated the test here to get the same behavior as before.

rousseldenis:
@chienandalu Could you review this too as this is a breaking change ? I think we need to update your ctt express module too to get the automatic package creation with the new module.

chienandalu (commented):
Hi @rousseldenis thanks for the heads up :slightly_smiling_face: Anyway, we don't use `base_delivery_carrier_label` in our carrier integrations, so I can't judge properly

rousseldenis:
> Hi @rousseldenis thanks for the heads up 🙂 Anyway, we don't use `base_delivery_carrier_label` in our carrier integrations, so I can't judge properly

Ok, I haven't put my head in the code. But could you explain the test fail here ? Thanks

chienandalu:
Ok, I didn't see that :man_facepalming:

chienandalu:
I can't see how your changes might affect that integration. Could relaunch them to give them another shot?

rousseldenis:
> I can't see how your changes might affect that integration. Could relaunch them to give them another shot?

Done

rousseldenis:
@chienandalu @florian-dacosta That's because delivery_package_number uses a 'set_default_package' context... but does not depends on base_delivery_carrier_label...

@chienandalu So, your module was working by chance (I suppose you have base_delivery_carrier_label installed on your projects too :sweat_smile: )

rousseldenis:
> @chienandalu @florian-dacosta That's because delivery_package_number uses a 'set_default_package' context... but does not depends on base_delivery_carrier_label...
> 
> @chienandalu So, your module was working by chance (I suppose you have base_delivery_carrier_label installed on your projects too 😅 )

@florian-dacosta @chienandalu I propose to add package_automatic module in delivery_package_number dependencies

chienandalu:
> That's because delivery_package_number uses a 'set_default_package' context... but does not depends on base_delivery_carrier_label...

I see it's intentional and the only purpose is to avoid OCA's tests fail: https://github.com/OCA/delivery-carrier/blob/31b6df1f7f39b8b6ae6b1f15f65408437ce1d3a8/delivery_package_number/wizard/stock_inmediate_transfer.py#L13-L16

rousseldenis:
Mmmh ok, didn't see the False value. So, there is something that put a new package before this that makes your tests happy because cttexpress requires a package I see.

chienandalu:
> So, there is something that put a new package before this that makes your tests happy because cttexpress requires a package I see.

I'd say is the opposite, as the issue is that we're declaring 0 packages to the carrier (and this will happen for every carrier integration using this methods). Somehow this new chain of dependencies make that this compute is triggered after we've already set the number of packages: https://github.com/OCA/delivery-carrier/blob/31b6df1f7f39b8b6ae6b1f15f65408437ce1d3a8/delivery_package_number/models/stock_picking.py#L19-L23

My only guess is that somehow the company stays with the automatic package setting to `True`, but I didn't find where could that be set.

Things I'd try:

- Try out that theory printing that flag value when it fails so we can probe it at the moment of fail.
- Maybe a context to that compute that avoids being triggered when the number of packages is confirmed from the wizard.

delivery_roulier/tests/test_delivery_roulier.py
mt-software-de: A migration script is missing to set the value also on existing carriers,
otherwise it is a breaking change.

mt-software-de (changes requested):
Can you cherry-pick 
https://github.com/OCA/delivery-carrier/pull/786/commits/1984571fc791ead775c659fca2f9c9447d594932
https://github.com/OCA/delivery-carrier/pull/786/commits/86b0685f4456390391c9253e480b79cb6fb67a14
https://github.com/OCA/delivery-carrier/pull/786/commits/f38d3d3094778ce370402a82cc009293f8a4375b

mt-software-de:
Which methods are called in delivery_postlogistics from base_delivery_carrier_label without super?
Because there is not dependency on delivery_postlogistics for base_delivery_carrier_label

mt-software-de:
> Which methods are called in delivery_postlogistics from base_delivery_carrier_label without super? Because there is not dependency on delivery_postlogistics for base_delivery_carrier_label

@rousseldenis could you give an update on this?
My PR https://github.com/OCA/delivery-carrier/pull/786 is waiting. 
We should refactor the module or run the tests in a separate flow like i did it in 786.

hparfr:
I'm ok with @florian-dacosta: don't add a new dependency to base_delivery_label (opposit of this pr)
Remove: `  def _set_a_default_package(self): ` (already in the pr)

In delivery_roulier package have to exist and it has to fail if no packages are defined. 
How the packages are created is not in the scope of delivery_roulier

PR #644, opened 2023-05-23
Update __manifest__.py

andreampiovesana:
"application": False

### DIFF
--- a/delivery_postlogistics/__manifest__.py
+++ b/delivery_postlogistics/__manifest__.py
@@ -25,5 +25,5 @@
     ],
     "installable": True,
     "auto_install": False,
-    "application": True,
+    "application": False,
 }


andreampiovesana:
is really an app?

rousseldenis:
@jbaudoux @jcoux @yvaucher

andreampiovesana:
@rousseldenis can you fix?

jbaudoux (changes requested):
All delivery_* connector modules in enterprise are `application=True`

So this is consistent and there is no reason to change that

andreampiovesana:
change all

Il giorno lun 5 feb 2024 alle ore 09:31 Jacques-Etienne Baudoux <
***@***.***> ha scritto:

PR #647, opened 2023-05-23
[15.0][OU-ADD] delivery_auto_refresh: Migration script to delete deactivated parameters

pilarvargas-tecnativa:
In v15 the module does not set the default parameters to any value as was the case in v14. In v15 when a parameter is activated from settings a record is created in the database and when it is deactivated the record is deleted. For this reason and to avoid possible problems with the configuration of the parameters, those that are deactivated are deleted from the database.

cc @Tecnativa TT43515

@chienandalu @CarlosRoca13 please review

### DIFF
--- a/delivery_auto_refresh/migrations/15.0.1.0.0/pre-migration.py
+++ b/delivery_auto_refresh/migrations/15.0.1.0.0/pre-migration.py
@@ -1,0 +1,17 @@
+# Copyright 2023 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """Remove system parameter records of 'delivery_auto_refresh' with value 0 or False,
+    to avoid their activation during migration due to their presence in the database."""
+    openupgrade.logged_query(
+        env.cr,
+        """
+        DELETE FROM ir_config_parameter
+        WHERE key LIKE '%delivery_auto_refresh%'
+        AND (value = '0' OR value = 'False')
+        """,
+    )


delivery_auto_refresh/migrations/15.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,15 @@
...
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """
+DELETE FROM ir_config_parameter
+WHERE ir_config_parameter.key LIKE '%delivery_auto_refresh%'
```
chienandalu: - All former parameters started with `delivery_auto_refresh`...
- You don't need to specify the table in a query from a single table where there can be no ambiguity
```suggestion
WHERE key LIKE 'delivery_auto_refresh%'
```

delivery_auto_refresh/migrations/15.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,15 @@
...
+AND (ir_config_parameter.value = '0' OR ir_config_parameter.value = 'False')
```
chienandalu:
```suggestion
AND value = '0' OR value = 'False'
```

chienandalu (commented):
LGTM. Just some tiny remarks

delivery_auto_refresh/migrations/15.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2023 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
```
chienandalu, line 7: Document

delivery_auto_refresh/migrations/15.0.1.0.0/pre-migration.py
pilarvargas-tecnativa: thanks! :smile:

CarlosRoca13 (approved):

chienandalu (approved):

delivery_auto_refresh/migrations/15.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,17 @@
...
+    """Remove system parameter records of 'delivery_auto_refresh' with value 0 or False,
+    to avoid their activation during migration due to their presence in the database."""
+    openupgrade.logged_query(
+        env.cr,
+        """
+DELETE FROM ir_config_parameter
+WHERE key LIKE '%delivery_auto_refresh%'
+AND (value = '0' OR value = 'False')
```
pedrobaeza: Although valid, visually it's better to indent the same as the code:

```suggestion
        DELETE FROM ir_config_parameter
        WHERE key LIKE '%delivery_auto_refresh%'
        AND (value = '0' OR value = 'False')
```

pedrobaeza (commented):
The commit message should be with tag `[OU-ADD]`

delivery_auto_refresh/migrations/15.0.1.0.0/pre-migration.py
pilarvargas-tecnativa: thanks :)

pedrobaeza (approved):

PR #649, opened 2023-05-25
[16.0][MIG] delivery_tnt_oca

hildickethan:
- Migration to 16.0
- Added PDF label format in seperate commit, I made it based on the ZPL one, not 1:1 but close enough
- Added multiple package support for both reports, now using the weight returned for the piece rather than the picking weight. I did this trying to maintain the sparse fields and with minimal changes, so I saved the multi-piece data in a string and then split it in the report
- Added Manifest, International manifest and Consignment note reports by @miguel-S73


RabbitJon-S73:
I changed message when for some reason TNT server is not reachable to specify that is a TNT fail instead Odoo's fail

PlantBasedStudio:
Hello, 
I'd like to know the details that caused this RP to fail and, if possible, try it again:
Can it be put into an open state again? 

@hildickethan-S73  can we have some information? 

Were the tests good?

hildickethan:
> Hello, I'd like to know the details that caused this RP to fail and, if possible, try it again: Can it be put into an open state again?
> 
> @hildickethan-S73 can we have some information?
> 
> Were the tests good?

From what I remember this migration worked fine, it's just that the only client we had that used this module stopped using it so when the branch went stale I just let it close. I didn't delete the branch in case it was needed again, if you need it I suggest cherry-picking off https://github.com/Studio73/delivery-carrier/tree/16.0-mig-delivery_tnt_oca and opening a new PR

PR #654, opened 2023-05-31
[16][MIG] delivery_roulier

florian-dacosta:
Standard migration

@hparfr @bealdav @bguillot


delivery_roulier/models/stock_picking.py
```diff
@@ -0,0 +1,282 @@
...
+            )
+            if move_line_no_pack:
+                # For automatic package creation, for use cases where we only have one
+                # pack per picking and we don't want to force the user to use pack
+                # consider installing delivery_automatic_package OCA module
+                raise UserError(
+                    _(
+                        "Some products have no destination package in picking %s, "
+                        "please add a destination package in order to be able to "
+                        "generate the carrier label."
```
bealdav, line 94: You might suggest to install `delivery_automatic_package` in the raise ?
Functionnal users won't read source code.

EDIT not blocking suggestion, I'll let you decide

florian-dacosta: This is not really an information for a functional/final user, I don't thunk it should be in the odoo error. Also because it can be consitent for some use cases and for other, they just want it to raise when user forgot to put a pack and it makes no sense to always tell him to install a new module!

Anyway, I am aware it is not very visible but that was just to put the information somewhere. Maybe I could put it in the know issue part instead

rousseldenis: After thinking of it, wouldn't it be more correct to add a new module like delivery_carrier_package_required ?

florian-dacosta: It is really 3 lines of code, I am not sure it is worth a module ?
Why would that help ? (I am not specially against, I am just wondering the gain ?)

rousseldenis: I would say it sounds like a transversal requirement for several modules.

BUT, as the implementations can vary in each of them, it could become too harsh to raise the error at the good moment.

So, let's the error implementation in the modules that require it.

florian-dacosta: @bealdav I have removed the comment and added the information in the README.

florian-dacosta:
@bealdav Comment attended
@hparfr please review when you got some time

hparfr (approved):

bguillot (approved):

PR #656, opened 2023-06-02
[IMP] delivery_auto_refresh: Compatibility with sale_tier_validation

ilyasProgrammer:
Check  https://github.com/OCA/server-ux/pull/612  fo reference

### DIFF
--- a/delivery_auto_refresh/models/sale_order.py
+++ b/delivery_auto_refresh/models/sale_order.py
@@ -53,15 +53,18 @@ class SaleOrder(models.Model):
         if self.state in {"draft", "sent"}:
             # Context added to avoid the recursive calls and save the new
             # value of carrier_id
-            self.with_context(auto_refresh_delivery=True)._remove_delivery_line()
+            self.with_context(
+                auto_refresh_delivery=True, skip_validation_check=True
+            )._remove_delivery_line()
         if self._get_param_auto_add_delivery_line() and self.carrier_id:
             if self.state in {"draft", "sent"}:
                 price_unit = self.carrier_id.rate_shipment(self)["price"]
-                if not self.is_all_service:
-                    self._create_delivery_line(self.carrier_id, price_unit)
-                self.with_context(auto_refresh_delivery=True).write(
-                    {"recompute_delivery_price": False}
+                self.with_context(skip_validation_check=True)._create_delivery_line(
+                    self.carrier_id, price_unit
                 )
+                self.with_context(
+                    auto_refresh_delivery=True, skip_validation_check=True
+                ).write({"recompute_delivery_price": False})
 
     @api.model
     def create(self, vals):


francesco-ooops (approved):
Functional ok!

@LoisRForgeFlow what do you think of this implementation?

delivery_auto_refresh/models/sale_order.py
```diff
@@ -53,14 +53,18 @@ def _auto_refresh_delivery(self):
         if self.state in {"draft", "sent"}:
             # Context added to avoid the recursive calls and save the new
             # value of carrier_id
-            self.with_context(auto_refresh_delivery=True)._remove_delivery_line()
+            self.with_context(
+                auto_refresh_delivery=True, skip_validation_check=True
```
rousseldenis, line 57: @ilyasProgrammer Nope. You cannot add code that this module is not aware of.

A better implementation should require a hook method to retrieve self for instance.

Then, in a glue module override that method and add context key

ilyasProgrammer: @rousseldenis but parent module supports this context option: https://github.com/OCA/server-ux/pull/612/commits/b487a53f7e2e183b9b8dc98757dde51be43d30fe#diff-5b72a073de187894d5d5401d6574d08eb57b1fcf07fcaa0ad1682b09dd65bc65R277

rousseldenis: Not a parent as this module depends on delivery only : 

image

GSLabIt:
can you please rebase?

ilyasProgrammer:
@GSLabIt done

elvise:
@ilyasProgrammer can you rebase ?

ilyasProgrammer:
@elvise done

delivery_auto_refresh/models/sale_order.py
HekkiMelody: Hello @rousseldenis ,

the "problem" with `base_tier_validation` is that it can be very strict, and it's hardly feasible to create a glue module for every possible module that would require it. The issue was also discussed here https://github.com/OCA/server-ux/pull/612#discussion_r1140034514 and that PR was approved based precisely on the principle of not wanting a proliferation of glue modules.

I see the addition of this context key as quite harmless, and a clean solution (maybe with a comment mentioning why it's being added) for the issue at hand.

It also wouldn't be the first time a module is modified to add code that is only relevant if another OCA module is also present.

Some examples
- https://github.com/OCA/project/blob/14.0/project_timesheet_time_control/wizards/hr_timesheet_switch.py#L116-L128
- https://github.com/OCA/account-invoicing/blob/14.0/sale_order_invoicing_queued/wizards/sale_advance_payment_inv.py#L23
- https://github.com/OCA/multi-company/blob/14.0/account_invoice_inter_company/models/account_move.py#L273
- https://github.com/OCA/multi-company/blob/14.0/account_invoice_inter_company/models/account_move.py#L390
- https://github.com/OCA/multi-company/blob/14.0/account_multicompany_easy_creation/wizards/multicompany_easy_creation.py#L15
- https://github.com/OCA/pos/blob/14.0/pos_payment_change/models/pos_config.py#L34

What do you think?

cc @LoisRForgeFlow

HekkiMelody: Hello @pedrobaeza , sorry to bother you, but you're very wise and knowledgeable. Do you have any advice regarding this topic?

pedrobaeza: As both modules are in the OCA ecosystem, and they are wide-spread used, I think it's reasonable to put the context without requiring a glue module. It's true that this context can't be tested automatically (as you don't have the module installed in the CI), but it's not too much.

Not sure if there can be a generic mechanism though for ignoring "special lines" in `base_tier_validation`.

HekkiMelody: thank you :pray:

HekkiMelody (approved):

francesco-ooops:
@pedrobaeza can this be merged based on existing reviews? 🎄🎄🎄

PR #657, opened 2023-06-07
[16.0][ADD] delivery_deliverea

OscarIndias:
The module delivery deliverea has been added.


asierneiradev (approved):

OscarIndias:
Good morning @pedrobaeza, I don't know if you remember, but in the oca days we talked about a new integration of a carrier, Deliverea. For that I created this PR, could you check it?
Thanks!

delivery_deliverea/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+    "summary": "Delivery Carrier implementation for Deliverea using their API",
+    "version": "16.0.1.0.0",
+    "category": "Stock",
+    "website": "https://github.com/OCA/delivery-carrier",
+    "author": "FactorLibre, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "base_delivery_carrier_label",
```
pedrobaeza: We talked about not depending on this one, and for me is a totally blocker.

delivery_deliverea/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+        "delivery_state",
+        "queue_job",
```
pedrobaeza: This is also not very convenient. A more abstract `delivery_queue_job` should be done that intercepts any API call (being SEUR, Deliverea, or whatever, and enqueue it). Anyway, I don't think this is a good idea, as you don't have immediate feedback if the sending is registered, and thus this can lead to missing packages not being informed to the carrier, provoking business problems.

delivery_deliverea/README.rst
```diff
@@ -0,0 +1,151 @@
...
+In order to select the type of carrier and the type of service to be used for this
+shipping method, click on the Import carriers and services button.
+
+Once the distribution centers and services are imported, we can select them in the
+fields.
+
+Known issues / Roadmap
+======================
+
+* No specified roadmap and no known problems in operation.
```
pedrobaeza: If none, don't put it.

delivery_deliverea/README.rst
```diff
@@ -0,0 +1,151 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/delivery-carrier-16-0/delivery-carrier-16-0-delivery_deliverea
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runboat-Try%20me-875A7B.png
+    :target: https://runboat.odoo-community.org/builds?repo=OCA/delivery-carrier&target_branch=16.0
+    :alt: Try me on Runboat
+
+|badge1| |badge2| |badge3| |badge4| |badge5|
+
+Deliverea API intration with Odoo.
```
pedrobaeza: Do you mean `integration`? Can you explain a bit what Deliverea is? Please don't save words here, which is very important.

pedrobaeza (changes requested):
Very blocking points.

delivery_deliverea/__manifest__.py
OscarIndias: You are right, I changed the code but I forgot to remove the dependency.

delivery_deliverea/__manifest__.py
OscarIndias: It's ok, I put it to avoid possible concurrency problems, but it's not mandatory so I remove it.

delivery_deliverea/README.rst
OscarIndias: removed.

delivery_deliverea/README.rst
OscarIndias: I have explained what deliverea is so you can understand a little bit how it works.

delivery_deliverea/__manifest__.py
pedrobaeza: Please remove this dependency then. As said, you can make a generic `delivery_queue_job` module that intercepts the delivery methods (`_send_shipper` and company), and enqueue them instead of doing it directly. Check https://github.com/OCA/bank-payment/blob/8.0/account_payment_transfer_reconcile_batch/models/payment_order.py for a similar technique.

OscarIndias: Okaay thanks.
I removed the dependency.

hugo-cordoba:
Hi, @pedrobaeza! How are you doing?

Is there anything on our end that we can do to move this PR forward?  Can we somehow help merge it?

Regards!

delivery_deliverea/controllers/http.py
```diff
@@ -0,0 +1,27 @@
...
+                request.httprequest.data.decode(request.httprequest.charset or "utf-8")
+            )
+        }
+        request.httprequest.data = request.httprequest._cached_data = json.dumps(
+            data
+        ).encode(request.httprequest.charset or "utf-8")
+    __original___init__(self, request=request)
+
+
+JsonRPCDispatcher.__init__ = __init__
```
pedrobaeza: Monkeypatching is not very advisable. Why do you need this? It should be done at other level or using proper inheritance.

delivery_deliverea/models/deliverea_shipping_label.py
```diff
@@ -0,0 +1,17 @@
+# © 2023 - FactorLibre - Oscar Indias <oscar.indias@factorlibre.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+
+
+class DelivereaShippingLabel(models.Model):
+    _name = "deliverea.shipping.label"
```
pedrobaeza: This model doesn't have too much sense. Just put the attachment in the picking.

delivery_deliverea/models/deliverea_distribution_center.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2023 FactorLibre - Jorge Martínez <jorge.martinez@factorlibre.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class DelivereaDistributionCenter(models.Model):
+    _name = "deliverea.distribution.center"
```
pedrobaeza, line 7: Why not using standard `res.partner`?

delivery_deliverea/models/carrier_deliverea.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2022 FactorLibre - Jorge Martínez <jorge.martinez@factorlibre.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class CarrierDeliverea(models.Model):
+    _name = "carrier.deliverea"
```
pedrobaeza: This model should disappear and put this data in the `delivery.carrier`.

delivery_deliverea/readme/CREDITS.rst
pedrobaeza, line 1: Don't put empty files.

delivery_deliverea/readme/INSTALL.rst
```diff
@@ -0,0 +1,8 @@
+This module needs the `unidecode` python library.
+
+It depends on the modules:
+
+* `base_delivery_carrier_label <https://github.com/OCA/delivery-carrier/tree/16.0/base_delivery_carrier_label>`_ repositorio OCA/delivery-carrier
```
pedrobaeza, line 5: Remove this reference and the queue_job one.

delivery_deliverea/static/description/icon.png
pedrobaeza, line 1: Please put square images of 140x140.

delivery_deliverea/models/deliverea_distribution_center.py
OscarIndias: It may make sense to use the res.partner model, but a new field would have to be created to identify it from the other distribution centers.

It would be the uuid field.

delivery_deliverea/static/description/icon.png
pedrobaeza, line 1: Still not 140x140 squared

pedrobaeza (commented):
Some points still unattended.

delivery_deliverea/models/deliverea_distribution_center.py
OscarIndias: Good afternoon @pedrobaeza 

Yes I have not yet made all the changes, because I have the doubt if the distribution centers model should disappear and add the new field in res.partner or leave that model.

I think that if it could be changed, but creating the field would be mandatory, because you have to differentiate them by id with the new field.

What is your opinion?

Thanks in advance.

pedrobaeza: Is this populated manually or automatically? If automatically, then it's reasonable to keep it isolated in this model.

OscarIndias: Is automatic, there are a button in the delivery.carrier model to import all the distribution centers.

pedrobaeza: OK, then let's not pollute the partner table for that. It's OK this way :)

delivery_deliverea/static/description/icon.png
OscarIndias: The new image is 140x140.

delivery_deliverea/readme/INSTALL.rst
OscarIndias: I removed this reference.

delivery_deliverea/models/carrier_deliverea.py
OscarIndias: This model was to be able to filter the services by the different carriers, but you were right, I have put them in the delivery.carrier.

delivery_deliverea/models/delivery_carrier.py
```diff
@@ -0,0 +1,538 @@
...
+        string="Return Proof Delivery",
+        help="Whether or not to a proof of delivery should be returned"
+        " back to origin (availability depends on carrier and service)",
+    )
+    deliverea_hide_sender = fields.Boolean(
+        string="Hide Sender",
+        help="Whether or not to hide sender information in the printed label"
+        " (availability depends on carrier and service)",
+    )
+    deliverea_carrier_service = fields.Many2one(
```
pedrobaeza: All m2o should end with `_id` by convention.

OscarIndias: Changed the field name.

OscarIndias:
Hello, @pedrobaeza , how are you?
I made the changes you asked for, could you check it please?
Thank you!

pedrobaeza (approved):
Thanks. Next time, please ping me as soon as you have completed the requested changes, so that I can be aware.

/ocabot merge nobump

pedrobaeza:
@OscarIndias please fix es.po file, that contains a translation error:
```
 delivery_deliverea/i18n/es.po:491 Translation string couldn't be parsed correctly using str%variables TypeError('not enough arguments for format string') - [po-python-parse-printf]
```

OscarIndias:
Hello @pedrobaeza, done!

I fixed the error in the es.po file.

PR #659, opened 2023-06-08
[14.0] Fix access to not-existing record self after rollback

guewen:
If a stock.picking is created and in the same transaction we generate a label, when the webservice returns an error, the module rollbacks the transaction, so 'self' is empty and will fail with:

Record does not exist or has been deleted. (Record: stock.picking(8211067,), User: 6)

When trying to write the tracking number, and will therefore not raise the details of the error.

Out of scope here, but I wonder if we really need to commit / rollback, if we rollback and we "lose" a label, we can generate a new one, there is no fee per generated label.


Ricardoalso (approved):

simahawk (approved):
Seems legit

PR #660, opened 2023-06-09
[14.0][MIG][IMP] delivery automatic package

jbaudoux:
Backport from 16.0

Add feature to create a package per smallest product packaging. The purpose is that each picked box is automatically put in a package so that we can store a tracking reference on each box.

cc @rousseldenis @bealdav


jbaudoux:
@rousseldenis This module is badly designed. You cannot modify a done picking. So you cannot hook on `.send_to_shipper`. I close this PR and the module should be dropped in v16. I'll push a replacement module.

rousseldenis:
> @rousseldenis This module is badly designed. You cannot modify a done picking. So you cannot hook on `.send_to_shipper`. I close this PR and the module should be dropped in v16. I'll push a replacement module.

This was done in purpose replacing the function here : https://github.com/OCA/delivery-carrier/blob/14.0/base_delivery_carrier_label/models/stock_picking.py#L53 without any other change in order to keep module behavior.

PR #663, opened 2023-06-21
[16.0][IMP] add button Create Shipping Label

khoivo1601:


TDu (approved):

delivery_postlogistics/views/stock.xml
```diff
@@ -6,6 +6,16 @@
         <field name="model">stock.picking</field>
         <field name="inherit_id" ref="delivery.view_picking_withcarrier_out_form" />
         <field name="arch" type="xml">
+            <field name="state" position="before">
+                <button
+                    name="action_generate_carrier_label"
+                    help="Create Shipping Label 🚚"
+                    attrs="{'invisible': [('state', '!=', 'done')]}"
```
mmequignon: you also have to hide the button when `carrier_id.delivery_type != 'postlogistics'`

delivery_postlogistics/models/stock_picking.py
```diff
@@ -236,3 +236,9 @@ def generate_postlogistics_shipping_labels(self, package_ids=None):
         """Add label generation for PostLogistics"""
         self.ensure_one()
         return self._generate_postlogistics_label(package_ids=package_ids)
+
+    def action_generate_carrier_label(self):
+        self.ensure_one()
+        if not self.carrier_id:
+            raise exceptions.UserError(_("Please, set a carrier."))
+        self.env["delivery.carrier"].postlogistics_send_shipping(self)
```
mmequignon, line 244: what if carrier isn't postlogistics?

mmequignon (changes requested):

grindtildeath (commented):
Also, what happens if we already have a shipping label? Is it replaced? Do we get a second one?

delivery_postlogistics/models/stock_picking.py
khoivo1601: if isn't postlogistics, shipping label will not be created and attached ([https://github.com/OCA/delivery-carrier/blob/16.0/delivery_postlogistics/models/stock_picking.py#L61](https://github.com/OCA/delivery-carrier/blob/16.0/delivery_postlogistics/models/stock_picking.py#L61))

delivery_postlogistics/views/stock.xml
khoivo1601: hello @mmequignon, i updated this domain

delivery_postlogistics/models/stock_picking.py
mmequignon: I'm more concerned about carrier calling `_set_a_default_package`
https://github.com/OCA/delivery-carrier/blob/16.0/delivery_postlogistics/models/delivery_carrier.py#L139C18-L139C40

nilshamerlinck: hello @mmequignon this comes from `base_delivery_carrier_label` that has made packages mandatory since its [12.0 migration](https://github.com/Timo17100-c2c/delivery-carrier/pull/1); `delivery_postlogistics` was created based on `delivery_carrier_label_postlogistics` 12.0 (-> `base_delivery_carrier_label`) and kept same logic

nilshamerlinck:
hello @grindtildeath in that case, we would get a second label; this PR emerged from a need to make things more flexible for the users so it didn't focus on adding restrictions; let us know if you think this would need to be revisited cc @meredith-alana

PR #665, opened 2023-06-27
[16.0][FIX]  delivery_postlogistics: correct customer and recipient depending on picking code

ajaniszewska-dev:
customer and recipient on shipping label are wrong for return picking (should be swapped in places)

### DIFF
--- a/delivery_postlogistics/postlogistics/web_service.py
+++ b/delivery_postlogistics/postlogistics/web_service.py
@@ -102,6 +102,12 @@ class PostlogisticsWebService(object):
                 _("Phone number is required for phone call notification.")
             )
 
+        if not partner.street:
+            raise exceptions.UserError(_("Partner street is required."))
+
+        if not partner.name and not partner.parent_id.name:
+            raise exceptions.UserError(_("Partner name is required."))
+
         partner_name = partner.name or partner.parent_id.name
         sanitized_partner_name = self._sanitize_string(partner_name)
         partner_street = self._sanitize_string(partner.street)
--- a/delivery_postlogistics/postlogistics/web_service.py
+++ b/delivery_postlogistics/postlogistics/web_service.py
@@ -79,6 +79,13 @@ class PostlogisticsWebService(object):
 
         """
         partner = picking.partner_id
+        if picking.picking_type_id.code != "outgoing":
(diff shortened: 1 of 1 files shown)


delivery_postlogistics/postlogistics/web_service.py
```diff
@@ -78,7 +78,12 @@ def _prepare_recipient(self, picking):
         :return a dict containing data for ns0:Recipient
 
         """
-        partner = picking.partner_id
+        if picking.picking_type_id.code == "outgoing":
+            partner = picking.partner_id
+        elif picking.picking_type_id.code == "incoming":
+            location_dest = picking.location_dest_id
+            partner = location_dest.company_id.partner_id or self.env.user.company_id
```
sebalix:
```suggestion
            partner = location_dest.company_id.partner_id or self.env.user.company_id.partner_id
```

But here the `partner` is not defined if the transfer is not `outgoing` or `incoming` (probably not real use cases but better to not crash if we want to use SwissPost on internal tranfers). So I would keep the original line as a default value (we often use this module to ship goods, not to receive them) and complete the code:
```python
partner = picking.partner_id
if picking.picking_type_id.code != "outgoing":
    location_dest = picking.location_dest_id
    partner = location_dest.company_id.partner_id or self.env.user.company_id.partner_id
```

delivery_postlogistics/postlogistics/web_service.py
```diff
@@ -144,7 +149,10 @@ def _prepare_customer(self, picking):
 
         """
         company = picking.company_id
-        partner = company.partner_id
+        if picking.picking_type_id.code == "outgoing":
+            partner = company.partner_id
+        elif picking.picking_type_id.code == "incoming":
+            partner = picking.partner_id
```
sebalix: Same remark. Add new method to return the relevant partner to not copy the same logic.

sebalix (approved):

JuMiSanAr (approved):

sebalix:
@ajaniszewska-dev can you remove the `[16.0]` text from the first commit message?

ajaniszewska-dev:
> @ajaniszewska-dev can you remove the `[16.0]` text from the first commit message?

@sebalix  done

PR #668, opened 2023-07-03
Add delivery_carrier_return_barcode_pattern

mmequignon:


delivery_carrier_return_barcode_pattern/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+{
+    "name": "Delivery Carrier Return Barcode Pattern",
+    "version": "14.0.1.0.0",
```
simahawk, line 7: summary missing...

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
```diff
@@ -0,0 +1,35 @@
...
+import re
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = "delivery.carrier"
+
+    return_barcode_pattern = fields.Char(help="explain here")
```
simahawk: explain! :)

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
```diff
@@ -0,0 +1,35 @@
...
+            if record.return_barcode_pattern:
+                try:
+                    re.compile(record.return_barcode_pattern)
+                except (re.error):
+                    raise UserError(
+                        _("Return Barcode Pattern must be a valid regular expression")
+                    )
+
+    @api.model
+    def _get_origin_from_barcode(self, barcode):
```
simahawk, line 34: don't we know the carrier when we receive the goods? Do we really have to load all carriers all the times?

simahawk (changes requested):
tests?

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
mmequignon: No, incoming goods might be coming from any available carrier.
I could select distinct patterns though, as I suppose most of the time it'll be empty.

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
mmequignon: oops, forgot

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
mmequignon: I improved this, by only using selecting distinct patterns.

mmequignon:
@simahawk can you make another pass please?

mmequignon:
> @simahawk can you make another pass please?

@simahawk ping :)

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
```diff
@@ -0,0 +1,51 @@
...
+    return_barcode_pattern = fields.Char(help=HELP_RETURN_BARCODE_PATTERN)
+
+    @api.constrains("return_barcode_pattern")
+    def _constrains_return_barcode_pattern(self):
+        for record in self:
+            if record.return_barcode_pattern:
+                try:
+                    re.compile(record.return_barcode_pattern)
+                except (re.error):
+                    raise UserError(
```
rousseldenis: constrains should raise ValidationError instead

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
```diff
@@ -0,0 +1,51 @@
...
+                        _("Return Barcode Pattern must be a valid regular expression")
+                    )
+
+    @api.model
```
rousseldenis: Shouldn't this method use ormcache ?

rousseldenis (changes requested):

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
mmequignon: Hmm.
This depends on the result of `self.search([]).mapped("return_barcode_pattern")`.
Is there a way to do depend on that?
I mean, what happens if a `delivery.carrier.return_barcode_pattern` is edited/added/removed between two calls of this method?
And also, this is quite unlikely that the same barcode would be used multiple times.

rousseldenis: So, in write(), if return_barcode_pattern is in vals, clear cache.

mmequignon: What about this?

> And also, this is quite unlikely that the same barcode would be used multiple times.

As I understand, result would be fetched from the cache only when we already called this method for a given barcode.
When receiving a package, how many times do you need to scan its barcode?
I'd say once.
I'm wondering then if using ormcache here is relevant.
What do you say?

delivery_carrier_return_barcode_pattern/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+{
+    "name": "Delivery Carrier Return Barcode Pattern",
+    "summary": "Allow to define a return barcode pattern on delivery methods",
```
sebalix, line 6: This feature could be used outside of a return workflow I guess?
This is all about matching a barcode originated from a carrier, should we enforce the use of "Return" here?

sebalix (commented):
LG, one question

sebalix (approved):

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
rousseldenis: Yes indeed.

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
jbaudoux: I don't like this `@api.model`. If you don't know the carrier, it is up to the caller to build self with all carriers.
If you know the possible carriers, you want to load only the corresponding patterns

delivery_carrier_return_barcode_pattern/__manifest__.py
jbaudoux: It is for processing what comes back from the carrier, so somehow yes

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
```diff
@@ -0,0 +1,51 @@
...
+            FROM delivery_carrier;
+        """
+        self.flush()
+        self.env.cr.execute(pattern_query)
+        patterns = [row[0] for row in self.env.cr.fetchall()]
+        res = []
+        for pattern in patterns:
+            match = re.match(pattern, barcode)
+            if match:
+                res.append(match.groups()[0])
```
jbaudoux: The pattern could contain several information that can be useful. I would split this method: one method to load the matches, and one method to return all origins from all the matches.

With such regex `Y(?P<origin>.*)` you can type your groups.

```suggestion
                res.append(match.group("origin"))
```

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
```diff
@@ -0,0 +1,51 @@
...
+        the barcode itself.
+        """
+        pattern_query = """
+            SELECT DISTINCT(COALESCE(return_barcode_pattern, '(.*)'))
+            FROM delivery_carrier;
+        """
+        self.flush()
+        self.env.cr.execute(pattern_query)
+        patterns = [row[0] for row in self.env.cr.fetchall()]
+        res = []
```
jbaudoux:
```suggestion
        res = set([])
```

jbaudoux (changes requested):
Remove `@api.model` from `_get_origin_from_barcode`

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def _get_origin_from_barcode(self, barcode):
+        """Return a list of strings, parsed with carrier return barcode patterns.
+
+        Even if no pattern matched the barcode, it will return a list containing
+        the barcode itself.
+        """
+        pattern_query = """
+            SELECT DISTINCT(COALESCE(return_barcode_pattern, '(.*)'))
+            FROM delivery_carrier;
```
jbaudoux: @JuMiSanAr WHERE ID in self.ids

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
```diff
@@ -0,0 +1,50 @@
...
+                    )
+
+    def _get_origin_from_barcode(self, barcode):
+        """Return a list of strings, parsed with carrier return barcode patterns.
+
+        Even if no pattern matched the barcode, it will return a list containing
+        the barcode itself.
+        """
+        pattern_query = """
+            SELECT DISTINCT(COALESCE(return_barcode_pattern, '(.*)'))
```
jbaudoux: I would exclude the carriers where nothing is defined. You better set a default value on the field return_barcode_pattern than adding it in the read
```suggestion
            SELECT DISTINCT(trim(return_barcode_pattern))
            ...
            WHERE coalesce(trim(return_barcode_pattern), '') != ''
```

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
```diff
@@ -0,0 +1,50 @@
...
+            FROM delivery_carrier;
+        """
+        self.flush()
+        self.env.cr.execute(pattern_query)
+        patterns = [row[0] for row in self.env.cr.fetchall()]
+        res = set()
+        for pattern in patterns:
+            match = re.match(pattern, barcode)
```
jbaudoux, line 45:
```suggestion
            try:
                match = re.match(pattern, barcode)
            except re.error:
                continue
```

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
```diff
@@ -0,0 +1,50 @@
...
+
+import re
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+HELP_RETURN_BARCODE_PATTERN = """
+    This pattern can be used to parse a return barcode from the carrier.
+    This has to be a valid regular expression.
+    See https://docs.python.org/3/library/re.html
```
jbaudoux, line 14:
```suggestion
    Example for barcode containing only the PO: (?P<origin>.*)
    
    See https://docs.python.org/3/library/re.html
```

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
```diff
@@ -0,0 +1,50 @@
...
+            FROM delivery_carrier;
+        """
+        self.flush()
+        self.env.cr.execute(pattern_query)
+        patterns = [row[0] for row in self.env.cr.fetchall()]
+        res = set()
+        for pattern in patterns:
+            match = re.match(pattern, barcode)
+            if match:
+                res.add(match.groups()[0])
```
jbaudoux:
```suggestion
                res.add(match.group("origin"))
```

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
jbaudoux: No need for this as there is already a validation of the configuration

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
```diff
@@ -0,0 +1,58 @@
...
+import re
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+HELP_RETURN_BARCODE_PATTERN = """
+    This pattern can be used to parse a return barcode from the carrier.
+    This has to be a valid regular expression.
+
+    Example for barcode containing only the PO: (?P<origin>.*)
```
jbaudoux: My bad, it's SO and not PO
```suggestion
    Example for barcode containing only the sales order reference: (?P<origin>.*)
```

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
```diff
@@ -0,0 +1,58 @@
...
+        return list(res)
+
+    def _get_pattern_query(self):
+        pattern_query = """
+            SELECT DISTINCT(trim(return_barcode_pattern))
+            FROM delivery_carrier
+            WHERE COALESCE(trim(return_barcode_pattern), '') != ''
+        """
+        if self.ids:
+            pattern_query += "AND id IN {ids}".format(ids=tuple(self.ids))
```
jbaudoux: safer to return the query and query_args and use the 2 in the call to cr.execute

jbaudoux (approved):

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
```diff
@@ -0,0 +1,60 @@
...
+        pattern_query = """
+            SELECT DISTINCT(trim(return_barcode_pattern))
+            FROM delivery_carrier
+            WHERE COALESCE(trim(return_barcode_pattern), '') != ''
+        """
+        if self.ids:
+            pattern_query += "AND id IN %s"
+            ids = tuple(self.ids)
+        pattern_query += ";"
+        return (pattern_query, [ids])
```
jbaudoux, line 61: `ids` not defined if not self.ids

jbaudoux (changes requested):

jbaudoux (approved):

jbaudoux:
@rousseldenis @simahawk Can you update your review?

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
```diff
@@ -0,0 +1,61 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+import re
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+HELP_RETURN_BARCODE_PATTERN = """
```
rousseldenis, line 9: @mmequignon Would this be translated as is ?

rousseldenis (approved):
Code review. Minor comment

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
simahawk: Yes, why not?

simahawk (approved):

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
rousseldenis: That's what I said.

Here under, the commit that update the pot file:

https://github.com/OCA/delivery-carrier/commit/aaffa4de50fa8bf187538720eeee4b8412c8f859

@mmequignon You should call explicitly translation function for strings that are not directly defined on fields property but through a variable.

Could you do the fix ?

PR #612, opened 2023-02-21
[14.0][FIX] base_delivery_carrier_files: Reset cursor back for StringIO()

Kiplangatdan:

PR #609, opened 2023-02-20
[14.0][MIG] delivery_postlogistics_dangerous_goods

mmequignon:
follow up of https://github.com/OCA/delivery-carrier/pull/441

Those prs have been merged:
- https://github.com/OCA/community-data-files/pull/109
- https://github.com/OCA/community-data-files/pull/137

PR #673, opened 2023-07-10
[DEL] [14.0] delivery_cttexpress: non-mocked tests 

thomaspaulb:
Remove these tests, which are failing upon remote connection, ruining pipeline for other PRs. Has to be redone using mock.


thomaspaulb:
Caused failing pipeline on https://github.com/OCA/delivery-carrier/pull/609, https://github.com/OCA/delivery-carrier/pull/659, https://github.com/OCA/delivery-carrier/pull/612 and possibly others.

@chienandalu

rousseldenis:
@thomaspaulb I totally agree on that point. The problem is the author is not understanding why not using real url in tests could cause failing runs in shared repositories...

At least, removing tests as you do is not good. We should keep them with mock at least

thomaspaulb:
Can't we delete them first, or even delete the whole module, and re-merge it back in later? Otherwise we might end up waiting for mocked tests for months, which will be blocking all the other PR

pedrobaeza (changes requested):
Of course I don't agree, as they are transient problems, and mocking them is like no having tests, but the tests are adding real value. You have already merged some of the linked PRs, so there's no such problem.

Please don't rant on something that is not really blocking you.

thomaspaulb:
@pedrobaeza Agreed about the ranting aspect! So lets not rant, but I do see conflicting ideas here:

- you think that mocked tests are no good
- other people think that tests that call the real API are not good

Who decides the policy on this, and how? The PSC's, i guess - is this worth a discussion on the mailinglist perhaps?

pedrobaeza:
I think it should be the module's maintainer, unless it interferes severely on the repository working, which is not the case.

rousseldenis:
> severely on the repository working, which is not the case.

That is the subjective part. I tend to think it should not at all. Not relaunching the debate but I give my opinion to the contributor's growing list that are annoyed by.

thomaspaulb:
As for me, I think I overestimated the "blocking" part, so I'm closing this PR for now - it's not "blocking", just "delaying" things to be merged, and I guess I can live with that.

Should it come to a debate later on, I'm rooting for the mock-tests team :-)

pedrobaeza:
I agree that is not comfortable when this happens, but put it on perspective: when other "online" components like pre-commit download fails (like the infamous `tell ^^^^` error), we have to retry as well. This problem with CTT has happened about 4 times in about 9 months, which is not a bad rate.

If the debate arises in a central topic, not as side one, I'll tell you why mocking doesn't serve most of the time IMO.

thomaspaulb:
Thanks @pedrobaeza :)

PR #674, opened 2023-07-11
[16.0][IMP] delivery_postlogistics: shipping labels and returns

QuocDuong1306:
- This is a takeover of this PR: https://github.com/OCA/delivery-carrier/pull/663
- It includes the same commit, but renamed to respect conventions (no odoo version in commits title)
- Also, it adds 2 things to improve the flow for Returns:
  - Returns will use same carrier as the original picking by default
  - native button `Print Return Labels` is shown for Returns instead of custom `Create Shipping Label`, and Return picking is properly identified as a return, with `is_return_picking = True`

### DIFF
--- a/delivery_postlogistics/wizard/stock_picking_return.py
+++ b/delivery_postlogistics/wizard/stock_picking_return.py
@@ -1,0 +1,12 @@
+from odoo import models
+
+
+class ReturnPicking(models.TransientModel):
+    _inherit = "stock.return.picking"
+
+    def _create_returns(self):
+        new_picking, pick_type_id = super(ReturnPicking, self)._create_returns()
+        if self.picking_id.delivery_type == "postlogistics":
+            picking = self.env["stock.picking"].browse(new_picking)
+            picking.write({"carrier_id": self.picking_id.carrier_id.id})
+        return new_picking, pick_type_id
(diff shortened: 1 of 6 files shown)


delivery_postlogistics/wizard/stock_picking_return.py
```diff
@@ -0,0 +1,12 @@
+from odoo import models
+
+
+class ReturnPicking(models.TransientModel):
+    _inherit = "stock.return.picking"
+
+    def _create_returns(self):
+        new_picking, pick_type_id = super(ReturnPicking, self)._create_returns()
+        if self.picking_id.delivery_type == "postlogistics":
+            picking = self.env["stock.picking"].browse(new_picking)
```
simahawk, line 10: so new_picking is an ID in fact?

simahawk (commented):
Overall LG but there are some unaddressed questions on https://github.com/OCA/delivery-carrier/pull/663

delivery_postlogistics/wizard/stock_picking_return.py
nilshamerlinck: fair point, could be named `new_picking_id`
but we tried to be consistent with upstream :) https://github.com/odoo/odoo/blob/16.0/addons/delivery/models/stock_picking.py#L329

nilshamerlinck:
hi @simahawk sorry for the delay, I've just answered to both pending questions directly on https://github.com/OCA/delivery-carrier/pull/663

grindtildeath (approved):

gurneyalex (approved):

PR #675, opened 2023-07-11
[14.0][IMP] delivery schenker and delivery_schenker_picking_volume

mt-software-de:
This changes are refactoring the code as well fixing delivery_schenker_picking_volume's auto_install. 

- deliver_schenker: Use packages also even if there is not stock.package_level set
- deliver_schenker_picking_volume: Set auto_install to False to only get the estimated volume if it is really desired
- delivery_schenker: Refactor shipping_information methods to make it easier hookable

This depends on https://github.com/OCA/stock-logistics-warehouse/pull/1790


mt-software-de:
cc @jbaudoux @raumschmiede-joshuaL

delivery_schenker/models/delivery_carrier.py
```diff
@@ -341,20 +341,40 @@ def _schenker_pickup_dates(self, picking):
...
+                )
+                for q in package.quant_ids
+            ]
+        )
+
+    def _schenker_shipping_information_round_volume(self, volume):
+        return round(volume, 2)
+
+    def _schenker_shipping_information_default_volume(self):
+        return 0.01
```
jbaudoux: Not linked to your refactoring but is it really better to declare a fake volume instead of having an error?

I would not do that.

delivery_schenker/models/delivery_carrier.py
```diff
@@ -341,20 +341,40 @@ def _schenker_pickup_dates(self, picking):
...
+
+    def _schenker_shipping_information_package(self, picking, package):
+        weight = package.shipping_weight or package.weight
         return {
             # Dangerous goods is not supported
             "dgr": False,
             "cargoDesc": picking.name + " / " + package.name,
             "grossWeight": round(weight, 2),
             # Default to 1 if no volume informed
```
jbaudoux: This comment is now misplaced. Also the comment value is not inline with the default value 1 != 0.01

delivery_schenker/models/delivery_carrier.py
```diff
@@ -341,20 +341,40 @@ def _schenker_pickup_dates(self, picking):
...
+            [
+                self._schenker_shipping_information_product_volume(
+                    q.product_id, q.quantity
+                )
+                for q in package.quant_ids
+            ]
+        )
+
+    def _schenker_shipping_information_round_volume(self, volume):
+        return round(volume, 2)
```
jbaudoux: volumes & weights are usually rounded to 3 digits. If it is in the schenker api requirements to round to 2, a comment would be appreciated

jbaudoux (approved):
LGTM

I just placed a few questions not linked to your refactoring but as you're touching this

delivery_schenker_picking_volume/__manifest__.py
```diff
@@ -11,5 +11,5 @@
     "author": "MT Software, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "depends": ["delivery_schenker", "stock_picking_volume"],
-    "auto_install": True,
+    "auto_install": False,
```
jbaudoux: I don't get why this change since the refactoring where you don't use anymore the picking volume

jbaudoux (changes requested):

delivery_schenker_picking_volume/models/delivery_carrier.py
```diff
@@ -6,7 +6,5 @@
 class DeliveryCarrier(models.Model):
     _inherit = "delivery.carrier"
 
-    def _prepare_schenker_shipping(self, picking):
-        vals = super()._prepare_schenker_shipping(picking)
-        vals["shippingInformation"]["volume"] = picking.volume
-        return vals
+    def _schenker_shipping_information_product_volume(self, product, qty):
+        return product._get_volume_for_qty(qty)
```
jbaudoux, line 10: This depends on https://github.com/OCA/stock-logistics-warehouse/pull/1790 so it should fail as it is not merged yet. Can you add a unittest for this?

delivery_schenker/models/delivery_carrier.py
jbaudoux: I would prefer float_round

delivery_schenker_picking_volume/__manifest__.py
mt-software-de: Because i am using the method _get_volume_for_qty instead of picking.volume. 
I would not install the module automatically, that should always be an explicit choice.

delivery_schenker/models/delivery_carrier.py
mt-software-de: I would also prefer to throw an error.

delivery_schenker/models/delivery_carrier.py
mt-software-de: i will have a look at the api

delivery_schenker_picking_volume/models/delivery_carrier.py
mt-software-de: I added this PR to the test-requirements.txt so it works, but yes there should be a test for it.

mt-software-de:
@hildickethan-S73 
could you please answer the comments from @jbaudoux?

delivery_schenker/models/delivery_carrier.py
hildickethan: if it's an Odoo UserError window it should be OK, because the error traceback when Schenker errors is unreadable to a common user

delivery_schenker/models/delivery_carrier.py
hildickethan: Schema says 3 digits, so it should be fine for the "volume" field in the package. `measureUnitVolume` is the only one restricted to 2 digits

delivery_schenker/models/delivery_carrier.py
mt-software-de: I removed the method and added the UserError into the method _schenker_get_total_shipping_volume

delivery_schenker_picking_volume/__manifest__.py
jbaudoux: But `_get_volume_for_qty` in `stock_picking_volume` is returning the same as the base module here https://github.com/OCA/delivery-carrier/pull/675/files#diff-f7e2841d64dbfbb11160cdc9a3518621837855957cbfc81446649cb15755f37eR346

So nothing is changing and auto_install can be True in my opinion. What is optional is to install https://github.com/OCA/stock-logistics-warehouse/tree/14.0/stock_picking_volume_packaging that improves the computation and could lead to a different result

mt-software-de: Ah yes you are right.

delivery_schenker_picking_volume/models/delivery_carrier.py
mt-software-de: There is now a unittest for it

jbaudoux (approved):

delivery_schenker/models/delivery_carrier.py
```diff
@@ -408,9 +445,21 @@ def _schenker_measures(self, picking, vals):
         :returns dict values for the proper unit key and value
         """
         if self.schenker_measure_unit == "VOLUME":
-            return {"measureUnitVolume": vals["shippingInformation"]["volume"]}
+            return {
+                "measureUnitVolume": self._schenker_shipping_information_round_volume(
+                    vals["shippingInformation"]["volume"], 2
```
jbaudoux:
```suggestion
                    vals["shippingInformation"]["volume"]
```

delivery_schenker/models/delivery_carrier.py
```diff
@@ -408,9 +445,21 @@ def _schenker_measures(self, picking, vals):
...
+                )
+            }
         return {}
 
+    def _schenker_get_total_shipping_volume(self, shipping_information):
+        volume = sum(info["volume"] for info in shipping_information)
+        if float_compare(volume, 0, 3) == 0:
```
jbaudoux:
```suggestion
        if float_compare(volume, 0, 3) <= 0:
```

delivery_schenker/models/delivery_carrier.py
mt-software-de: We had a look at the api, it says 2 digits for both.
image
image

mt-software-de: so i will change it to 2 digits

delivery_schenker/models/delivery_carrier.py
mt-software-de: The api requires 2 decimal points

hildickethan (approved):

PR #677, opened 2023-07-17
[IMP] delivery_multi_destination: Introduce 'based on destination' delivery type

carmenbianca:
Replaces #540

---

Internal task: https://gestion.coopiteasy.be/web#id=9184&model=project.task&view_type=form&menu_id=


carmenbianca:
@remytms Can you re-approve? This was approved in the closed PR.

PR #678, opened 2023-07-20
[14.0][FIX] delivery_schenker: round up package volume + weight

hildickethan:
- Round up Schenker package volume
  - The Schenker API requires a volume to be non-zero, but also forces 2 decimal places.
  - The 0.01 minimum was removed here https://github.com/OCA/delivery-carrier/pull/675#discussion_r1260265410 to avoid fake volumes when incorrectly configured, but we must still be able to deliver cases where the volume is non-zero but rounds to 0.00.
  - I think it's also safer to give an overestimate on the volume to the carrier, rather than an underestimate (though its only going to be miniscule volume differences)
  - I added `float(float_repr())` to the `float_round` to avoid trailing decimals that can happen sometimes, in my case it was rounding 2.0086 to 2.01 would result in 2.0100000000000002 which errored
- Same for weight

### DIFF
--- a/delivery_schenker/models/delivery_carrier.py
+++ b/delivery_schenker/models/delivery_carrier.py
@@ -5,7 +5,7 @@ from lxml import etree
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
-from odoo.tools import float_compare, float_round
+from odoo.tools import float_compare, float_repr, float_round
 
 from .schenker_request import SchenkerRequest
 
@@ -358,12 +358,27 @@ class DeliveryCarrier(models.Model):
             ]
         )
 
+    def _schenker_float_round(self, value, precision_digits):
+        """
+        The schenker api requires 2 decimal points.
+        Round up to avoid 0 if <0.005.
+        round -> repr -> float to avoid extra decimals that error with the Schema.
+        float_round(2.086, precision_digits=2) = 2.0100000000000002
+        """
+        return float(
+            float_repr(
+                float_round(
+                    value, precision_digits=precision_digits, rounding_method="UP"
+                ),
+                precision_digits=precision_digits,
+            )
+        )
+
     def _schenker_shipping_information_round_weight(self, weight, precision_digits=2):
-        return float_round(weight, precision_digits=precision_digits)
+        return self._schenker_float_round(weight, precision_digits=precision_digits)
 
     def _schenker_shipping_information_round_volume(self, volume, precision_digits=2):
-        """The schenker api requires 2 decimal points"""
-        return float_round(volume, precision_digits=precision_digits)
+        return self._schenker_float_round(volume, precision_digits=precision_digits)
 
     def _schenker_shipping_information_package(self, picking, package):
         weight = package.shipping_weight or package.weight


mt-software-de:
@hildickethan-S73 https://github.com/OCA/delivery-carrier/pull/710/files :-)

delivery_schenker/models/delivery_carrier.py
```diff
@@ -359,11 +359,36 @@ def _schenker_shipping_information_package_volume(self, picking, package):
...
+        float_round(2.086, precision_digits=2) = 2.0100000000000002
+        """
+        return float(
+            float_repr(
+                float_round(
+                    volume, precision_digits=precision_digits, rounding_method="UP"
+                ),
+                precision_digits=precision_digits,
+            )
+        )
```
mt-software-de: Could you create a own method for this?
Something like
"def _float_round"?

mt-software-de (approved):
Minor remarks

delivery_schenker/models/delivery_carrier.py
hildickethan: @mt-software-de done

mt-software-de:
@jbaudoux can you review and merge this one?

delivery_schenker/models/delivery_carrier.py
```diff
@@ -358,12 +358,27 @@ def _schenker_shipping_information_package_volume(self, picking, package):
...
         )
 
+    def _schenker_float_round(self, value, precision_digits):
+        """
+        The schenker api requires 2 decimal points.
+        Round up to avoid 0 if <0.005.
+        round -> repr -> float to avoid extra decimals that error with the Schema.
+        float_round(2.086, precision_digits=2) = 2.0100000000000002
+        """
+        return float(
```
jbaudoux, line 368: Why reconvert back to float? You will get back extra decimals:
```
print('{0:.16f}'.format(float("2.09")))
2.0899999999999999

```
SOAP is transmitting in XML, so converting it all to text, so once you have the representation with 2 decimals, you better send that, you don't need a real float in your dict.

delivery_schenker/models/delivery_carrier.py
```diff
@@ -358,12 +358,27 @@ def _schenker_shipping_information_package_volume(self, picking, package):
             ]
         )
 
+    def _schenker_float_round(self, value, precision_digits):
+        """
+        The schenker api requires 2 decimal points.
+        Round up to avoid 0 if <0.005.
+        round -> repr -> float to avoid extra decimals that error with the Schema.
+        float_round(2.086, precision_digits=2) = 2.0100000000000002
```
jbaudoux, line 366: Your example doesn't look right
```suggestion
        float_round(2.086, precision_digits=2) = 2.0899999999999999
```

jbaudoux, line 366:
```suggestion
        Returns the string representation with the given precision
```

mt-software-de:
> @hildickethan-S73 https://github.com/OCA/delivery-carrier/pull/710/files :-)
> 

@jbaudoux we can reopen this of course.

delivery_schenker/models/delivery_carrier.py
hildickethan: Seems like I wrote the comment wrong, in the PR description the issue was 2.0086, not 2.086
image
I will change it

delivery_schenker/models/delivery_carrier.py
hildickethan: I think this was just to maintain the same type as before, but I see what you mean with it being pointless

jbaudoux:
@hildickethan-S73 can we close this PR in favor of https://github.com/OCA/delivery-carrier/pull/710 as it solves it properly ?

hildickethan:
> @hildickethan-S73 can we close this PR in favor of #710 as it solves it properly ?

If we can merge it yes, I need the change suggested in https://github.com/OCA/delivery-carrier/pull/710#discussion_r1424313178 though since it was in this PR

hildickethan:
closed in favor of https://github.com/OCA/delivery-carrier/pull/710

PR #680, opened 2023-07-27
[ADD] delivery_packaging_archive: 

mt-software-de:
Add an active field on delivery packaging

cc @jbaudoux :-)


delivery_packaging_archive/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module allows archival of delivery packaging.
```
jbaudoux:
```suggestion
This module allows archiving a delivery packaging.
```

jbaudoux (approved):

mt-software-de:
@jaredkipe maybe do you want to review this?

jaredkipe:
Looks complete.

Would it make sense to prevent archiving if the package has contents, or maybe to unpack on archive?

mt-software-de:
> Looks complete.
> 
> Would it make sense to prevent archiving if the package has contents, or maybe to unpack on archive?

I think it is not needed, because there is no content on a product.packaging only on stock.quant.package.
Also you are only able to archive it within the delivery packaging view, which are used mainly for the delivery.carriers.

jaredkipe:
Of course! Then in that case, should we prevent archiving if it is linked as a default packaging type on a carrier?

mt-software-de:
> Of course! Then in that case, should we prevent archiving if it is linked as a default packaging type on a carrier?

As there is no direct link to the carrier, we can not do this. 
https://github.com/odoo/odoo/blob/78f6557ea3ed27506de8cdeec6c31ea06081f5d2/addons/delivery/wizard/choose_delivery_package_views.xml#L10
More like this.

jaredkipe:
There is a direct link on carrier, unfortunately this link is per-carrier which I didn't think about.  (meaning things like `ups_default_package_type_id` per carrier)

rousseldenis (approved):

PR #688, opened 2023-08-23
[14.0][MIG] backport from 15.0 partner_delivery_schedule

sergiocorato:


partner_delivery_schedule/tests/test_partner_delivery_schedule.py
```diff
@@ -4,11 +4,10 @@
 from odoo.tests import Form, TransactionCase
 
 
-class TestStockPikingReturnRefundOption(TransactionCase):
-    @classmethod
-    def setUpClass(cls):
```
simahawk: Pls do not change this. Keep setUpClass and use SavepointCase instead.
Also, it would be nice to disable tracking if you don't need it: `cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))`

simahawk (changes requested):
Question (maybe for @Tecnativa): what about using https://github.com/OCA/stock-logistics-workflow/tree/14.0/stock_partner_delivery_window for such cases?

partner_delivery_schedule/tests/test_partner_delivery_schedule.py
sergiocorato: Done, thanks!

sergiocorato:
> Question (maybe for @Tecnativa): what about using https://github.com/OCA/stock-logistics-workflow/tree/14.0/stock_partner_delivery_window for such cases?

This module is present in v. 12.0, 13.0 and 15.0, so that would mean losing data and recreating them.

sergiocorato:
(probably the 2 modules should be merged in a future version, maybe 16.0)

simahawk (approved):

sergiocorato:
@OCA/logistics-maintainers-psc-representative merge?

PR #689, opened 2023-08-23
[16.0][FIX] postlogistics: Set customer name mandatory

mmequignon:
if name1 is empty, postlogistics api returns an error.
Catch this before this is sent.

### DIFF
--- a/delivery_postlogistics/postlogistics/web_service.py
+++ b/delivery_postlogistics/postlogistics/web_service.py
@@ -168,6 +168,8 @@ class PostlogisticsWebService(object):
         if picking.picking_type_id.code != "outgoing":
             partner = picking.partner_id
 
+        if not partner.name:
+            raise exceptions.UserError(_("Customer name is required."))
         customer = {
             "name1": self._sanitize_string(partner.name),
             "street": self._sanitize_string(partner.street),


delivery_postlogistics/postlogistics/web_service.py
```diff
@@ -167,8 +167,11 @@ def _prepare_customer(self, picking):
         if picking.picking_type_id.code != "outgoing":
             partner = picking.partner_id
 
+        partner_name = partner.name
+        if not partner_name:
+            raise exceptions.UserError(_("Customer name is required."))
```
TDu, line 172: Should the raised error be more explicit ?
My thinking is, I mark a transfer as done, the label is printed in the process, this error comes on my screen...

mmequignon: bah, this is what's done for every other field required by postlogistics.
What would you do?

TDu: Look the other way, this was just a random comment.

TDu (approved):

delivery_postlogistics/postlogistics/web_service.py
mmequignon: :smile:

delivery_postlogistics/postlogistics/web_service.py
```diff
@@ -167,8 +167,11 @@ def _prepare_customer(self, picking):
         if picking.picking_type_id.code != "outgoing":
             partner = picking.partner_id
 
+        partner_name = partner.name
+        if not partner_name:
```
sebalix: Can't we write `if not partner.name`? (avoiding the `partner_name` used here and below).

grindtildeath (approved):

delivery_postlogistics/postlogistics/web_service.py
mmequignon: I'm not sure what's the point here.
The idea is to avoid fetching data from the record twice, as intuitively, i'd say it's slower than storing the value in a variable and then reusing it.

sebalix: Odoo already caches all basic data as soon as you access one of its field (prefetch mechanism):
```python
>>> partner.name    # First call: generate relevant queries to fetch data and store them in the cache
>>> partner.name    # Second call: no queries generated, it simply returns the cached value (unless you called `clear_caches` meanwhile)
```
So a variable is useless here, it brings only noise.

sebalix: done

PR #692, opened 2023-08-28
[14.0] delivery_postlogistics: fix 'sanitize_strings' to always return strings

sebalix:
Still an open question about this one: should the city of the partner be mandatory when it comes to send data to the delivery carrier about a shipment? (in a general manner, not only for PostLogistics)

ref 4054

### DIFF
--- a/delivery_postlogistics/postlogistics/web_service.py
+++ b/delivery_postlogistics/postlogistics/web_service.py
@@ -455,9 +455,9 @@ class PostlogisticsWebService(object):
 
     def _sanitize_string(self, value):
         """Removes disallowed chars ("|", "\", "<", ">", "’", "‘") from strings."""
-        if isinstance(value, str):
-            for char, repl in DISALLOWED_CHARS_MAPPING.items():
-                value = value.replace(char, repl)
+        value = value or ""
+        for char, repl in DISALLOWED_CHARS_MAPPING.items():
+            value = value.replace(char, repl)
         return value
 
     def generate_label(self, picking, packages):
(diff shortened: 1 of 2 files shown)


delivery_postlogistics/postlogistics/web_service.py
```diff
@@ -455,9 +455,9 @@ def get_access_token(cls, picking_carrier):
 
     def _sanitize_string(self, value):
         """Removes disallowed chars ("|", "\", "<", ">", "’", "‘") from strings."""
-        if isinstance(value, str):
-            for char, repl in DISALLOWED_CHARS_MAPPING.items():
-                value = value.replace(char, repl)
+        value = value or ""
```
JuMiSanAr, line 458: Wouldn't it be a tiny bit prettier to set it as default in the parameter?

`def _sanitize_string(self, value=""):`

JuMiSanAr (approved):

delivery_postlogistics/postlogistics/web_service.py
sebalix: The issue is that this method is called with `value=False` so even by setting a default value that won't work.

TDu (approved):

simahawk (approved):

simahawk:
> Still an open question about this one: should the city of the partner be mandatory when it comes to send data to the delivery carrier about a shipment? (in a general manner, not only for PostLogistics)
> 
> ref 4054

Good point. As a "gentle" workaround we could have a config flag on the company to make it required when needed.

/ocabot merge patch

Issue #676, opened 2023-07-14
delivery_package_number bug module when creating a backorder


angelgarciadelachica:
## Module
delivery_package_number

## Describe the bug
We have detected a bug in the delivery_package_number module when creating a backorder.

## To Reproduce
**Affected versions**:

1. Create a sales order containing at least one storable product and quantity greater than 1, example 10.
2. Confirm the sales order.
3. Go to the created delivery.
4. Enter the 'Number of Packages' in the 'Additional Info' tab, example 2.
5. Enter the packages in the 'Done' column less than 'Reserved', example 2.
6. Click on Validate and click on 'Generate Backorder' or 'No backorder' .

**Expected behavior**
If we go again to the tab 'Additional Info' the number of packages is 1 (incorrect) instead of 2 (correct).

**Additional context**
While trying to solve it we have detected that the problem is caused by the repeated execution of the _compute_number_of_packages function that recalculates the number of packages based on package_ids and sets the Number of Packages to 1 if there are no package_ids.

## Proposal
We propose the following changes to the delivery_package_number module to fix the bug:

- Create a boolean field that will appear next to the 'Number of Packages' to be able to switch between inserting the number manually or having it automatically calculated based on the package_ids. This way we avoid recalculating the 'Number of Packages' if it is in manual mode.
- On the other hand, if the company has disabled the 'Packages' in the Inventory settings, by default the option will be manual.
- In addition, we will also add a Boolean field in the 'Operation Type' to set whether it is calculated automatically or manually by default depending on the 'Type of Operation'. In this way, we will always be able to choose from the picking whether the 'Number of Packages' is inserted manually or automatically but, depending on the 'Type of Operation', there will be a default option.

angelgarciadelachica:
@ValentinVinagre @HaraldPanten @luis-ron

HaraldPanten:
@pedrobaeza @chienandalu that might be interesting for you 👍

pedrobaeza:
My colleague is on holidays. I can't say right now.

HaraldPanten:
> My colleague is on holidays. I can't say right now.

No problem. We'll wait few weeks. THX!

HaraldPanten:
@chienandalu @pedrobaeza Is it OK for you having a look at this?

We'll manage the IMP/FIX

THX!

pedrobaeza:
Reading the issue, I would see 2 problems:

[...]

PR #694, opened 2023-09-05
[15.0][BUG] delivery_package_number: bug when creating a backorder.

angelgarciadelachica:
This PR creates a Test to show the problem detected in the issue #676
A sales order with 10 units is confirmed
In the generated delivery it is established:
Quantities done = 2
Number of packages = 2
A backorder is created and it is verified that the number of packages in the validated delivery is 1 instead of 2.

### DIFF
--- a/delivery_package_number/tests/test_delivery_package_number.py
+++ b/delivery_package_number/tests/test_delivery_package_number.py
@@ -1,5 +1,7 @@
 # Copyright 2020 Tecnativa - David Vidal
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from datetime import datetime
+
 from odoo.tests import Form
 from odoo.tests.common import TransactionCase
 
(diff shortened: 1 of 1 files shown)


HaraldPanten:
@angelgarciadelachica 

- Please set [FIX] instead of [BUG] in the commit message. On the other hand, you can make the same change in this PR title.
- Are the tests fails related to the PR that you're creating here? Will them be solved with your PR?

Update:

- Set a proper commit message, for example: [FIX] delivery_package_number: fixing number of packages issue when creating a backorder.
- Odoo version is not needed in the commit message when we create a [FIX]. It's OK leaving it in the PR title.

THX

HaraldPanten:
> @angelgarciadelachica
> 
> * Please set [FIX] instead of [BUG] in the commit message. On the other hand, you can make the same change in this PR title.
> * Are the tests fails related to the PR that you're creating here? Will them be solved with your PR?
> 
> Update:
> 
> * Set a proper commit message, for example: [FIX] delivery_package_number: fixing number of packages issue when creating a backorder.
> * Odoo version is not needed in the commit message when we create a [FIX]. It's OK leaving it in the PR title.
> 
> THX

angelgarciadelachica:
@HaraldPanten

For now in this PR we develop some tests that fail to show and analyze the error commented in the issue #676.
When we solve the bug and it passes all the tests, I will make the changes you request for the PR.

Thanks!

pedrobaeza:
Thanks, with this I can work on checking alternative solutions.

rousseldenis:
What should be solved too is this: 

image


Why setting 1 if no packages ??? Nothing in README for that case

pedrobaeza:
The idea behind setting 1 as fallback is that most delivery carriers will need to set a minimum quantity of packages, and if you don't set any packaging in Odoo, it's because you are packing this in one shot (so 1 package), but you don't want to waste time defining packaging in the software (or you don't even show such feature).

pedrobaeza:
Solution in #696 

Several tips for tests:

- Don't use demo data. They are not reliable. Use data created in your test.
- Don't test things that your module is not adding. Testing if confirming a sales order creates a picking is not the duty of your module, so adding a lot of asserts about that things only confuses someone reading the test. In case of a failing test, you will debug if the problem is that the picking didn't generate.
- Don't add comments when the code explains itself thanks to good variable names.
- Extra (but this one is personal taste): readability decreases adding empty lines inside a method.

PR #695, opened 2023-09-06
[14.0][FIX] delivery_auto_refresh

solo4games:
Fix delivery_auto_refresh system parameter


francesco-ooops (approved):
Functional ok!

Before this PR, enabling delivery_auto_refresh.auto_add_delivery_line and creating a SO with only service products would add shipping expenses, while standard odoo does not allow shipping expenses to be added if consumable/storable products are not present in SO

@solo4games squash commits please

rousseldenis:
@solo4games It should be great if you add the corresponding test case. Thanks

rousseldenis (approved):
Makes sense.

pedrobaeza:
Please fw-port it to upper versions:

/ocabot merge patch

PR #696, opened 2023-09-07
[15.0][FIX] delivery_package_number: Don't reset number on backorder

### LINKED ISSUE #676: delivery_package_number bug module when creating a backorder

pedrobaeza:
Fixes #676

@Tecnativa

### DIFF
--- a/delivery_package_number/models/stock_picking.py
+++ b/delivery_package_number/models/stock_picking.py
@@ -1,4 +1,5 @@
 # Copyright 2020 Tecnativa - David Vidal
+# Copyright 2023 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import api, fields, models
 
@@ -11,9 +12,12 @@ class StockPicking(models.Model):
         compute="_compute_number_of_packages",
         readonly=False,
         store=True,
+        default=0,
+        copy=False,
     )
 
     @api.depends("package_ids")
     def _compute_number_of_packages(self):
         for picking in self:
-            picking.number_of_packages = len(picking.package_ids) or 1
+            if picking.package_ids:
+                picking.number_of_packages = len(picking.package_ids)
--- a/delivery_package_number/tests/test_delivery_package_number.py
+++ b/delivery_package_number/tests/test_delivery_package_number.py
@@ -1,4 +1,5 @@
 # Copyright 2020 Tecnativa - David Vidal
+# Copyright 2023 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo.tests import Form
 from odoo.tests.common import TransactionCase
@@ -11,6 +12,7 @@ class TestDeliveryPackageNumber(TransactionCase):
         cls.product = cls.env["product.product"].create(
             {"name": "Test product", "type": "product"}
         )
+        cls.partner = cls.env["res.partner"].create({"name": "Test partner"})
         cls.wh1 = cls.env["stock.warehouse"].create(
             {"name": "TEST WH1", "code": "TST1"}
         )
@@ -35,8 +37,8 @@ class TestDeliveryPackageNumber(TransactionCase):
         cls.ml2 = cls.ml1.copy({"qty_done": 0})
 
     def test_number_of_packages(self):
-        # By default it's computed to 1
-        self.assertEqual(self.picking.number_of_packages, 1)
+        # By default it's computed to 0
+        self.assertEqual(self.picking.number_of_packages, 0)
         # We can edit the number of packages as there aren't delivery packages
         picking_form = Form(self.picking)
         picking_form.number_of_packages = 3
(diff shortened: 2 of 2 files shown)


chienandalu (approved):

delivery_package_number/models/stock_picking.py
```diff
@@ -11,9 +12,12 @@ class StockPicking(models.Model):
         compute="_compute_number_of_packages",
         readonly=False,
         store=True,
+        default=1,
```
rousseldenis: @pedrobaeza Would you want to say there is always a package (which is not the case...)?

ValentinVinagre:
Functional Review.
When a partial delivery note is created, it does not ask for the number of packages. This case is what the issue refers to, if I'm not mistaken.

Selección_045
image


On the other hand, if you set the number of packages to 1 and you have a carrier, the user will never be asked if it is validated without entering the quantities manually (I think the number of packages would have to be 0 by default, the same as before).

pedrobaeza:
OK, having the prompt on the dialog to enter the number of packages seems enough. Put default=0.

HaraldPanten (approved):
LGTM tested in runboat 👍

HaraldPanten (changes requested):
Wait, it's not working. Try:

1- Create a sale order with 2 lines. 2 storable products with quantity 10 (for example), each one.
2- Add shipping.
3- Validate the SO.
4- Set 2 packages in the additional info field.
5- Set quantity 2 to the delivered quantity, only for 1 of the products.
6- Create backorder, or don't create backorder.

Result: The number of packages that we set before is changed to 1. 🤔

Tested in runboat.

pedrobaeza:
I have tried in local and it's working fine, and I have also made a test case in local for your steps and they are working, so one of these 2 is happening:

- The runboat that this is running doesn't contain the latest code.
- There's another module that is colliding with this one. This second seems improbable, as the CIs with the tests are green.

HaraldPanten (approved):
My colleagues told me the same. In a local environment seems to be working fine.

@angelgarciadelachica WDYT?

pedrobaeza:
Let's merge it then:

/ocabot merge patch

angelgarciadelachica:
In local it works and the tests are passed so it is probably what pedro says.
LGTM

pedrobaeza:
@sbidoul is https://github.com/OCA/delivery-carrier/pull/696#issuecomment-1719968270 something to worry about?

sbidoul:
@pedrobaeza sorry I don't really understand that comment. Can you elaborate?

pedrobaeza:
The problem is that the runboat linked to this PR didn't contain the patch we were applying. It's like the image taken was the one from the main branch.

sbidoul:
> The problem is that the runboat linked to this PR didn't contain the patch we were applying. It's like the image taken was the one from the main branch.

By patch you mean this PR ? I don't see how that could happen.

HaraldPanten:
Seems that when we were testing the feature in runboat (https://github.com/OCA/delivery-carrier/pull/696#pullrequestreview-1621632186) It was not working as expected. But it was working fine in a local environment.

On the other hand tests were correct, so it means that this branch (using this PR feature) was suposed to be working correctly as well (but it didn't in runboat).

Due to that, Pedro was wondering if some of these two options might be occuring: https://github.com/OCA/delivery-carrier/pull/696#issuecomment-1719968270

Taking in consideration that tests were OK, our doubts are about if runboat could be running without the latest code. Is that possible?

THX @sbidoul

sbidoul:
> doubts are about if runboat could be running without the latest code. Is that possible?

I very much doubt runboat would start a build with the wrong code. The runboat part that fetches the code is here https://github.com/sbidoul/runboat/blob/90513736535b20df9afba4560d5e345eba973c3b/src/runboat/kubefiles/runboat-clone-and-install.sh#L16C60-L16C75, where RUNBOAT_GIT_REF is the commit sha.

The only thing that I could imagine are the the status updates would be delayed and you tested with a previous version of the PR?

pedrobaeza:
> The only thing that I could imagine are the the status updates would be delayed and you tested with a previous version of the PR?

Yes, this is possible, but the link in GH status was only showing one build (not live now: https://runboat.odoo-community.org/builds/b97ff1cbb-2d82-49fa-8f06-ae5aff85a820?live)

HaraldPanten:
Same problem is happening here (https://github.com/OCA/delivery-carrier/pull/701), if you need another example.

THX

PR #699, opened 2023-09-13
[15.0][FIX] delivery_auto_refresh: fix auto_add_delivery_line setting

solo4games:
Now when creating a new SO consisting of product services, the delivery line is not added


pedrobaeza (approved):
Fast-tracking as already merged in previous version:

/ocabot merge patch

PR #700, opened 2023-09-18
[16.0][FIX] delivery_auto_refresh: fix auto_add_delivery_line setting

solo4games:
Now when creating a new SO consisting of product services, the delivery line is not added.


pedrobaeza (approved):
Forward-port of #695 and #699, so fast-tracking it:

/ocabot merge patch

PR #701, opened 2023-09-18
[16.0][FIX] delivery_package_number: Don't reset number on backorder

pedrobaeza:
Forward-port of #696 

@Tecnativa

### DIFF
--- a/delivery_package_number/models/stock_picking.py
+++ b/delivery_package_number/models/stock_picking.py
@@ -1,4 +1,5 @@
 # Copyright 2020 Tecnativa - David Vidal
+# Copyright 2023 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import api, fields, models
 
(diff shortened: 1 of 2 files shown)


HaraldPanten (approved):
Functionally tested in local environment. LGTM.

angelgarciadelachica (approved):
Reviewed in local and it works correctly. 
On the other hand, as in the PR https://github.com/OCA/delivery-carrier/pull/696, it does not work in the runbot.

sbidoul:
You can see in this [log](https://runboat.odoo-community.org/api/v1/builds/b84e3c687-289a-4add-b0f9-d24713bb6463/init-log) that the correct commit is obtained from github (look for the curl command at the beginning).

I have looked in the pod and the correct code has been downloaded. So the fault is not with runboat. There must be something else going on.

HaraldPanten:
Thanks @sbidoul ! Then, probably a "test evasive" use case is creating inconsistance.

pedrobaeza:
Very weird indeed. Let's merge it anyway for finishing the fw-port chain:

/ocabot merge patch

PR #702, opened 2023-09-20
[14.0][FIX] delivery_package_number: Don't reset number on backorder

### LINKED ISSUE #676: delivery_package_number bug module when creating a backorder

angelgarciadelachica:
Fixes #676

### DIFF
--- a/delivery_package_number/models/stock_picking.py
+++ b/delivery_package_number/models/stock_picking.py
@@ -1,4 +1,5 @@
 # Copyright 2020 Tecnativa - David Vidal
+# Copyright 2023 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import api, fields, models
 
@@ -11,9 +12,12 @@ class StockPicking(models.Model):
         compute="_compute_number_of_packages",
         readonly=False,
         store=True,
+        default=0,
+        copy=False,
     )
 
     @api.depends("package_ids")
     def _compute_number_of_packages(self):
         for picking in self:
-            picking.number_of_packages = len(picking.package_ids) or 1
+            if picking.package_ids:
+                picking.number_of_packages = len(picking.package_ids)
--- a/delivery_package_number/tests/test_delivery_package_number.py
+++ b/delivery_package_number/tests/test_delivery_package_number.py
@@ -1,4 +1,5 @@
 # Copyright 2020 Tecnativa - David Vidal
+# Copyright 2023 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo.tests import Form
 from odoo.tests.common import SavepointCase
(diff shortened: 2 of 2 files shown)


angelgarciadelachica:

ValentinVinagre (commented):
LGTM 👍🏻

ValentinVinagre (approved):

pedrobaeza (approved):

pedrobaeza:
Ouch, TNT module also relies on having 1 as default. You should change it.

HaraldPanten:
> Ouch, TNT module also relies on having 1 as default. You should change it.

Yes, we'll FIX it, don't worry. We'll ping you once we prepare the PR. THX Pedro.

HaraldPanten (approved):

HaraldPanten:
We were rebasing just 1 minute before merge command. I supose it'll not be a problem  🤔

pedrobaeza:
You don't need to rebase. The merge command rebases your branch over latest changes on target branch, so better to not waste a build. A rebase is needed for manual verifications or using latest code on runboat, but here, which is clear that the CI is fixed and the code of this PR is OK, we just need to launch the merge command.

HaraldPanten:
> You don't need to rebase. The merge command rebases your branch over latest changes on target branch, so better to not waste a build. A rebase is needed for manual verifications or using latest code on runboat, but here, which is clear that the CI is fixed and the code of this PR is OK, we just need to launch the merge command.

Yes, we know that the new branch will run the tests as well. And it would be so strange to fail, as few minutes ago, the other FIX PR has been merged in the same branch without any problem. But you know... just in case...

pedrobaeza:
It's not just that they run the CI. It rebases the branch, so that's the key.

HaraldPanten:
> It's not just that they run the CI. It rebases the branch, so that's the key.

We know that. But, as I said... just in case 🤷

PR #705, opened 2023-09-22
[IMP] delivery_package_number: number of packages in the backorder wizard

angelgarciadelachica:
The number of packages is added to the backorder wizard in the same way as it appears in the wizard to validate all picking.

PR #708, opened 2023-09-27
[IMP] delivery_package_number: number of packages in the backorder wizard

angelgarciadelachica:
Same IMP as in v14 https://github.com/OCA/delivery-carrier/pull/705

### DIFF
--- a/delivery_package_number/wizard/stock_backorder_confirmation.py
+++ b/delivery_package_number/wizard/stock_backorder_confirmation.py
@@ -1,0 +1,23 @@
+# Copyright 2023 Ángel García de la Chica Herrera <angel.garcia@sygel.es>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class StockBackorderConfirmation(models.TransientModel):
+    _inherit = "stock.backorder.confirmation"
+
+    number_of_packages = fields.Integer(
+        help="Set the number of packages for picking to be validated",
+    )
+    ask_number_of_packages = fields.Boolean(compute="_compute_ask_number_of_packages")
+
+    @api.depends("pick_ids")
+    def _compute_ask_number_of_packages(self):
+        for item in self:
+            item.ask_number_of_packages = bool(item.pick_ids.carrier_id)
+
+    def process(self):
+        if self.number_of_packages:
+            self.pick_ids.write({"number_of_packages": self.number_of_packages})
+        return super().process()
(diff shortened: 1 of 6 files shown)


angelgarciadelachica:

ValentinVinagre:
@angelgarciadelachica The wizard does not save the number of packages on the delivery note in the runboat.

HaraldPanten:
@pedrobaeza one question:

- The module is working fine in local.
- It seems that another module (don't know which one) might be interfering with this improvement and causing a malfunction.
- Would it be better to find (and fix) the "guilty code" first, before merging this IMP?

THX

HaraldPanten:
> @pedrobaeza one question:
> 
> * The module is working fine in local.
> * It seems that another module (don't know which one) might be interfering with this improvement and causing a malfunction.
> * Would it be better to find (and fix) the "guilty code" first, before merging this IMP?
> 
> THX

pedrobaeza:
Yes, digging in the problem would be better. Check override of the wizards to detect the guilty.

angelgarciadelachica:
Hi, 
We have found the module that is causing the problem. The module is [base_delivery_carrier_label](https://github.com/OCA/delivery-carrier/tree/15.0/base_delivery_carrier_label). Specifically in the function _set_a_default_package. The function is called when the order is shipped with a carrier. What this function does is to pack all the lines that are not packed in a package. Therefore, the number that prevails is the number of packages, which is correct.

https://github.com/OCA/delivery-carrier/blob/97f8a7e00ec84da0537d5d73b3ce0e1f1e7e3bf6/base_delivery_carrier_label/models/stock_picking.py#L62

We can report this incompatibility in the README or open an issue. What do you think @pedrobaeza?

THX!

pedrobaeza:
For me it's a mistake to auto-create one package that way unconditionally. I see the context `set_default_package` to inhibit it. You can do a soft dependency passing `set_default_package=False` in the context when the number of packages is set, putting a comment about the relation.

delivery_package_number/wizard/stock_backorder_confirmation.py
```diff
@@ -15,9 +15,13 @@ class StockBackorderConfirmation(models.TransientModel):
...
+                item.pick_ids.carrier_id and not item.pick_ids._origin.package_ids
+            )
 
     def process(self):
+        ctx = self._context.copy()
         if self.number_of_packages:
+            ctx.update({"set_default_package": False})
             self.pick_ids.write({"number_of_packages": self.number_of_packages})
-        return super().process()
+        return super(StockBackorderConfirmation, self.with_context(**ctx)).process()
```
pedrobaeza: No need of copying the full context, and put a comment why you are doing this:

```suggestion
        # put context key for avoiding `base_delivery_carrier_label` auto-packaging feature
        return super(StockBackorderConfirmation, self.with_context(set_default_package=False)).process()
```

pedrobaeza (approved):
/ocabot merge minor

PR #710, opened 2023-10-01
[14.0][FIX] delivery_schenker: correct float_round return

mt-software-de:
Because float_round as an example returns for 1.88 -> 1.880...01
we have to round it properly with float_repr

### DIFF
--- a/delivery_schenker/models/delivery_carrier.py
+++ b/delivery_schenker/models/delivery_carrier.py
@@ -5,7 +5,7 @@ from lxml import etree
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
-from odoo.tools import float_compare, float_round
+from odoo.tools import float_compare, float_repr, float_round
 
 from .schenker_request import SchenkerRequest
 
@@ -362,11 +362,21 @@ class DeliveryCarrier(models.Model):
             ]
         )
 
+    def _float_round_repr(self, value, precision_digits, rounding_method="UP"):
+        return float_repr(
+            float_round(
+                value,
+                precision_digits=precision_digits,
+                rounding_method=rounding_method,
+            ),
+            precision_digits=precision_digits,
+        )
+
     def _schenker_shipping_information_round_dimension(
         self, dimension, precision_digits=2
     ):
         """The schenker api requires 2 decimal points"""
-        return float_round(dimension, precision_digits=precision_digits)
+        return self._float_round_repr(dimension, precision_digits)
 
     def _schenker_shipping_information_package(self, picking, package):
         weight = package.shipping_weight or package.weight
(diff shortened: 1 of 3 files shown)


mt-software-de:
Closed. Duplicate of https://github.com/OCA/delivery-carrier/pull/678

delivery_schenker/models/delivery_carrier.py
```diff
@@ -358,12 +358,18 @@ def _schenker_shipping_information_package_volume(self, picking, package):
             ]
         )
 
+    def _float_round(self, value, precision_digits):
```
jbaudoux:
```suggestion
    def _float_round_repr(self, value, precision_digits):
```

jbaudoux (approved):

delivery_schenker/models/delivery_carrier.py
```diff
@@ -358,12 +358,18 @@ def _schenker_shipping_information_package_volume(self, picking, package):
...
+        return float_repr(
+            float_round(value, precision_digits=precision_digits),
```
hildickethan:
```suggestion
            float_round(value, precision_digits=precision_digits, rounding_method="UP"),
```
you need this in case the volume is <0.005, it will error due to rounding to 0.0 when the minimum is 0.01

jbaudoux: I don't think we should always round up, but better set a min on the volume

jbaudoux: does this min applies also to weight ?

hildickethan: I'm not sure, looking at the schema I would say yes

The difference is that with a minimum of 0.01 then cases where volume is misconfigured to 0 would get through, it was something you mentioned a few months ago when the minimum was removed https://github.com/OCA/delivery-carrier/pull/675#discussion_r1260265410.

I think it's ok to error when the volume isn't configured properly, but the issue I encountered was that even when configured properly, sometimes the volume was just very small (<0.005) and it rounded to 0 and errored, hence I changed it to always rounding up with the idea that giving an overestimate would always be better for the carrier than an underestimate but the difference being at most 0.01 makes it feel pretty indifferent to the carrier

mt-software-de: I think rounding up is fine for volume and weight

mt-software-de:
The failing Tests are not related to this Change, there is as interference through an other addon

jbaudoux:
@mt-software-de Can you rebase ?

mt-software-de:
> @mt-software-de Can you rebase ?

yes but this will not fix the tests.
This is waiting on https://github.com/OCA/delivery-carrier/pull/786 or https://github.com/OCA/delivery-carrier/pull/643

hildickethan (approved):

Raul-S73:
@mt-software-de Can you rebase?

mt-software-de:
> @mt-software-de Can you rebase?

done

hildickethan:
@mt-software-de Could you adapt this part of the tests in `delivery_schenker_quant_package_dimension`? 
image
Due to changing the values to be strings in this PR it fails. Just noticed it in our CI that uses this PR

mt-software-de:
> @mt-software-de Could you adapt this part of the tests in `delivery_schenker_quant_package_dimension`? image Due to changing the values to be strings in this PR it fails. Just noticed it in our CI that uses this PR

Thx. I forgot that.

mt-software-de:
The 

> @mt-software-de Could you adapt this part of the tests in `delivery_schenker_quant_package_dimension`? image Due to changing the values to be strings in this PR it fails. Just noticed it in our CI that uses this PR

done

hildickethan:
> done

:thinking: It seems to still be failing, I think just changing the test itself in that module to compare vs strings is OK, like you did with delivery_schenker

mt-software-de:
> > done
> 
> 🤔 It seems to still be failing, I think just changing the test itself in that module to compare vs strings is OK, like you did with delivery_schenker

now :-)

adrip-s73:
@mt-software-de now test is working fine thx

mt-software-de:
@jbaudoux could you please merge this one?

mt-software-de:
> /ocabot merge patch

???

mt-software-de:
@jbaudoux could please try to merge this one?

PR #711, opened 2023-10-02
[16.0][FIX] delivery_cttexpress: add zeep as external_dependencies in maniest file

RodrigoBM:
add zeep as external_dependencies in maniest file


pedrobaeza:
It's already in core:

https://github.com/odoo/odoo/blob/16.0/requirements.txt#L64

RodrigoBM:
Yes @pedrobaeza, but "pre-commit" gives this message 

Selección_527

pedrobaeza:
That should be fixed at pre-commit config file AFAIK

PR #712, opened 2023-10-03
[14.0][FIX] delivery_package_number: inhibit auto-create packets from base_delivery_carrier_label module

angelgarciadelachica:
Same FIX as in v15 PR https://github.com/OCA/delivery-carrier/pull/708

### DIFF
--- a/delivery_package_number/wizard/stock_backorder_confirmation.py
+++ b/delivery_package_number/wizard/stock_backorder_confirmation.py
@@ -15,9 +15,16 @@ class StockBackorderConfirmation(models.TransientModel):
     @api.depends("pick_ids")
     def _compute_ask_number_of_packages(self):
         for item in self:
-            item.ask_number_of_packages = bool(item.pick_ids.carrier_id)
+            # we use ._origin because if not, a NewId is used for the checks and the returned
+            # value of package_ids is wrong.
+            item.ask_number_of_packages = bool(
+                item.pick_ids.carrier_id and not item.pick_ids._origin.package_ids
+            )
 
     def process(self):
         if self.number_of_packages:
             self.pick_ids.write({"number_of_packages": self.number_of_packages})
-        return super().process()
+        # put context key for avoiding `base_delivery_carrier_label` auto-packaging feature
+        return super(
+            StockBackorderConfirmation, self.with_context(set_default_package=False)
+        ).process()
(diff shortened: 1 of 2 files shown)


angelgarciadelachica:

HaraldPanten (approved):
Functional review in runboat. LGTM

HaraldPanten:
@pedrobaeza I think that this one can be merged. THX 👍

delivery_package_number/wizard/stock_backorder_confirmation.py
```diff
@@ -15,9 +15,14 @@ class StockBackorderConfirmation(models.TransientModel):
     @api.depends("pick_ids")
     def _compute_ask_number_of_packages(self):
         for item in self:
-            item.ask_number_of_packages = bool(item.pick_ids.carrier_id)
+            item.ask_number_of_packages = bool(
+                item.pick_ids.carrier_id and not item.pick_ids._origin.package_ids
```
pedrobaeza, line 21: Why this?

pedrobaeza: It's incorrect to use `_origin` in a compute, as it's using what was stored last time, which is nothing that can be compared now in the current record state.

angelgarciadelachica: I use `_origin` because otherwise it does not get the `picking_ids`. 

That is, if I use `item.picking_ids` it returns `stock.picking(<NewId origin=XX>,)` and `item.picking_ids.package_ids` always returns `stock.quant.package()` even if the `stock.picking` does have `package_ids`.

How else can I do it?

pedrobaeza: OK, being a transient model which is populated only for this and no more changes are done over `picking_ids`, it's acceptable, but please put a code comment explaining why that "design decission" for your future you and for the rest to know why.

angelgarciadelachica: I have put a comment in the code, what do you think?
THX

pedrobaeza (approved):
OK, it would be interesting to say also that it's not a problem, as the field is only set once on the initialization of the transient, but let's not block this more:

/ocabot merge patch

PR #714, opened 2023-10-03
[16.0][REF] Use copier v1.17.1 (previously v1.14.2)

RodrigoBM:
run copier update.


RodrigoBM:
Copier update to v1.17.1 clean false positive example https://github.com/OCA/delivery-carrier/pull/711

PR #719, opened 2023-10-10
[16][FIX] base_delivery_carrier_label : qty from operation for any uom

florian-dacosta:
For stock move, the weight is computed no matter of the uom of the product.
I think there is no point on doing differently for `stock.move.line`
Also fix uom qty conversion as qty_done and reserved_qty are not necessarly in the same UOM

@hparfr @bealdav


base_delivery_carrier_label/models/stock_move_line.py
```diff
@@ -17,32 +17,23 @@ class StockMoveLine(models.Model):
...
-                cant_calc_total = True
             # reserved_qty may be 0 if you don't set move line
             # individually but directly validate the picking
-            qty = operation.qty_done or operation.reserved_qty
+            qty = (
+                operation.qty_done
+                and operation.product_uom_id._compute_quantity(
+                    operation.qty_done, operation.product_id.uom_id
+                )
+                or operation.reserved_qty
```
hparfr, line 34: the quantity is converted only if it's done ; if it's only reserved there is no convertion ?

hparfr (commented):
my memory is a bit rusty on this topic; why removing the test instead of fixing it ?

florian-dacosta:
> my memory is a bit rusty on this topic; why removing the test instead of fixing it ?

Because the test was testing something that does not seem to be the right behavior, so there is nothing to fix IMO.

If the uom of the product is the cm and the weight is 1, it means the wieght of the product is 1kg per cm, no?
The only case where we can have an inconsistency is if the uom of the product is of the weight type. We could have a product with uom the gram but its weight beeing 10 gram for instance, but this is a confiuration issue/inconsistency, not sure we really need to solve this here (Odoo does not on the stock move weight).

base_delivery_carrier_label/models/stock_move_line.py
florian-dacosta: qty done is meant to be consistent with the uom of the stock move (it should named uom_qty_done actually) while reserved_qty consistent with the uom of the product. (reserved_uom_qty exists and is consistent with the uom of the stock move).

PR #725, opened 2023-10-30
[14.0][MIG] delivery_purchase: Migration to 14.0

victoralmau:
Migration to 14.0

@Tecnativa TT45656


delivery_purchase/models/purchase_order.py
```diff
@@ -0,0 +1,32 @@
...
+    _inherit = "purchase.order"
+
+    carrier_id = fields.Many2one(
+        comodel_name="delivery.carrier",
+        string="Delivery Method",
+    )
+    delivery_price = fields.Float()
+
+    @api.onchange("partner_id")
+    def onchange_partner_id_delivery_purchase(self):
```
pedrobaeza: Better to convert both fields to computed writable.

victoralmau: Ok, changes done.

delivery_purchase/tests/test_delivery_purchase.py
```diff
@@ -6,7 +6,7 @@
 class TestDeliveryPurchase(common.SavepointCase):
     @classmethod
     def setUpClass(cls):
-        super(TestDeliveryPurchase, cls).setUpClass()
+        super().setUpClass()
```
chienandalu, line 9: nitpick: shut mail stuff by context if possible

delivery_purchase/models/purchase_order.py
```diff
@@ -0,0 +1,35 @@
...
+from odoo import api, fields, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    carrier_id = fields.Many2one(
+        comodel_name="delivery.carrier",
+        compute="_compute_carrier_id",
+        store=True,
```
chienandalu, line 16:
```suggestion
        store=True,
        readonly=False,
```

delivery_purchase/models/purchase_order.py
```diff
@@ -0,0 +1,35 @@
...
+        string="Delivery Method",
+    )
+    delivery_price = fields.Float(compute="_compute_delivery_price", store=True)
```
chienandalu:
```suggestion
    delivery_price = fields.Float(compute="_compute_delivery_price", store=True, readonly=False)
```

chienandalu (commented):
I don't know the context of the module. Should a line with de delivery charges be created in the PO as it is in the SO?

pedrobaeza:
> Should a line with the delivery charges be created in the PO as it is in the SO?

That's one of the expansions to do (you can check in the TT).

victoralmau:
> I don't know the context of the module. Should a line with de delivery charges be created in the PO as it is in the SO?

Changes done.

delivery_purchase/models/purchase_order.py
```diff
@@ -25,14 +26,71 @@ def _compute_carrier_id(self):
...
 
     @api.model
     def _prepare_picking(self):
         res = super()._prepare_picking()
         if self.carrier_id:
             res["carrier_id"] = self.carrier_id.id
         return res
+
+    @api.depends("order_line.is_delivery")
+    def _get_invoice_status(self):
```
pedrobaeza: Instead of overwriting this method at purchase order level, can you please override the `invoice_status` at line status and disable deliveries one if no other are invoiceable?

delivery_purchase/models/purchase_order.py
```diff
@@ -25,14 +26,71 @@ def _compute_carrier_id(self):
...
+
+    def _create_delivery_line(self, carrier, price_unit):
+        product = carrier.product_id
+        if product.description_purchase:
+            po_description = "%s: %s" % (carrier.name, product.description_purchase)
+        else:
+            po_description = carrier.name
+
+        taxes = product.supplier_taxes_id
+        if self.partner_id and self.fiscal_position_id:
```
pedrobaeza: Why doing this manually instead of relying in standard? Call `onchange` for getting the proper values.

delivery_purchase/models/purchase_order.py
victoralmau: The `invoice_status` field does not exist in lines, another option would be to overwrite the `qty_to_invoice` field.

The `invoice_status` overwrite option is the one that `delivery` with sale orders.

pedrobaeza: This line https://github.com/OCA/delivery-carrier/pull/725/files/ab4625a182b4f57ff128ecd652b2c011119e87a1..6cda513556c0f2e54378b566d6085a0adbe3d1a2#diff-65ffd2f9c25409d29856d4366222d99ca9291f92663574680e428ba5e551fafeR58 is using `invoice_status` in lines.

victoralmau: Ok sorry, changed the approach to use `qty_to_invoice`

delivery_purchase/models/purchase_order.py
victoralmau: I have changed the way of set the values to avoid all this with another approach, do you see it correct now?

chienandalu (approved):
Tested :+1:

pedrobaeza (approved):
/ocabot merge nobump

PR #726, opened 2023-10-30
[14.0][IMP] Update copier template

victoralmau:
Update copier template

@Tecnativa


pedrobaeza:
There's already #724, which I'm fixing.

PR #684, opened 2023-08-06
[16.0][ADD] sale_order_warehouse_from_delivery_carrier

geomer198:
This module allows to set warehouse in Sales Order based on selected shipping method.

PR #731, opened 2023-11-09
[16.0][FIX]  Tests are fixed.

geomer198:
Related with this [RP](https://github.com/OCA/delivery-carrier/pull/684).


rousseldenis:
@geomer198 Mmmhh, 16.0 branch does not seem to be broken.

geomer198:
> @geomer198 Mmmhh, 16.0 branch does not seem to be broken.

But [here](https://github.com/OCA/delivery-carrier/pull/684) without this changes are tests is broken.

PR #737, opened 2023-12-12
[13.0][IMP] delivery_cttexpress: configurable label delay

chienandalu:
fw of https://github.com/OCA/delivery-carrier/pull/735

When we recod a shipping in the CTT Express backend the label API could not be ready in the moment to retrieve the labels. As CTT Express stated, a little delay would add the needed time for the API to be ready.

cc @Tecnativa TT46443

### DIFF
--- a/delivery_cttexpress/models/delivery_carrier.py
+++ b/delivery_cttexpress/models/delivery_carrier.py
@@ -1,5 +1,7 @@
 # Copyright 2022 Tecnativa - David Vidal
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+import time
+
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
 
@@ -38,6 +40,13 @@ class DeliveryCarrier(models.Model):
         string="Document format",
     )
     cttexpress_document_offset = fields.Integer(string="Document Offset")
+    cttexpress_label_delay = fields.Float(
+        string="Label Fetch Delay (seconds)",
+        help="The server might not be ready to deliver the label right after the "
+        "shipping is saved. We need to introduce a little delay. Tipacally 1 s will be "
+        "enough",
+        default=1,
+    )
 
     def _ctt_request(self):
         """Get CTT Request object
@@ -253,6 +262,7 @@ class DeliveryCarrier(models.Model):
         if not reference:
             return False
         ctt_request = self._ctt_request()
+        time.sleep(self.cttexpress_label_delay)
         try:
             error, label = ctt_request.get_documents_multi(
                 reference,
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

victoralmau (approved):

dreispt:
@chienandalu Can you rebase to fix pre-commit, please?

chienandalu:
Done @dreispt

delivery_cttexpress/models/delivery_carrier.py
```diff
@@ -253,6 +262,7 @@ def cttexpress_get_label(self, reference):
         if not reference:
             return False
         ctt_request = self._ctt_request()
+        time.sleep(self.cttexpress_label_delay)
```
rousseldenis, line 265: @chienandalu Don't do this.

Instead, I suggest you to use the session timeout argument of the zeep client.

See : https://docs.python-zeep.org/en/master/transport.html

pedrobaeza: It's not for that. The problem is that CTT is returning immediately the response, but preparing on background the label, so asking very soon for it will return a "404" (well, the equivalent). This is the only method.

rousseldenis: > This is the only method.

Nope.

Further, in the get_documents_multi, the client - if timeout defined correctly - will behave as expected without sleep() needs: https://github.com/Tecnativa/delivery-carrier/blob/e55397fda15d1d2845d760f23f6e7fd8df262875/delivery_cttexpress/models/cttexpress_request.py#L181

rousseldenis: The better will be to implement a 'retry' mechanism for the '404' errors. As you cannot do a back and forth with the sleep timeout for label...

chienandalu: > Further, in the get_documents_multi, the client - if timeout defined correctly - will behave as expected without sleep() needs

Yes, but when that method is invoked you already have a shipping reference...

And that's the issue precisely. As CTT explained us their system is quite clunky (well that's our conclusion :smile: ).

The main issue here is that the backend part for the label generation has some kind of cron and you can't expect to have the label inmidiately. If I remember well, there's no error, you'll only get an empty response.

Could we do a retry mechanism and retry forever? or after a `n` number o loops? Should we wait a little between loops to avoid issues with hundreds of successive calls against a weakly engineered API...?  😏️

pedrobaeza: I think it's enough in this case. It's true that is not the most elegant thing, but this is not active by default and we can't invest more here.

PR #740, opened 2023-12-13
[14.0][FIX] delivery_purchase: Change the position of the carrier to the purchase order header

victoralmau:
Change the position of the carrier to the purchase order header.

**v13**:
purchase-order-v13

**Before**:
purchase-order-antes

**After**:
purchase-order-ahora

@Tecnativa TT45656


pedrobaeza:
Sorry, for pickings, it's correct that is not on the header, as this is how it goes also for delivery orders AFAIK.

victoralmau:
> Sorry, for pickings, it's correct that is not on the header, as this is how it goes also for delivery orders AFAIK.

Removed the change of picking.

chienandalu (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #741, opened 2023-12-14
[14.0][IMP] delivery_purchase: Set carrier_price when creating a picking

victoralmau:
Set `carrier_price` when creating a picking

If we have a delivery price of 10 on the order, it is expected to show 10 on the picking (without needing to be set to that amount when set the picking as done).

@Tecnativa TT45656


chienandalu (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #745, opened 2023-12-20
[14.0][FIX] delivery_postlogistics: fix date computation

TDu:
If there is multiple commitment dates when creating a picking. Use the date the furthest in the future.

### DIFF
--- a/delivery_postlogistics/models/stock_move.py
+++ b/delivery_postlogistics/models/stock_move.py
@@ -10,13 +10,14 @@ class StockMove(models.Model):
     def _get_new_picking_values(self):
         vals = super(StockMove, self)._get_new_picking_values()
 
-        order_commitment_date = (
-            self.sale_line_id and self.sale_line_id.order_id.commitment_date
-        )
-
-        if order_commitment_date:
+        order_commitment_dates = [
+            date
+            for date in self.sale_line_id.order_id.mapped("commitment_date")
+            if date
+        ]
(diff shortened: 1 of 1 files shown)


delivery_postlogistics/models/stock_move.py
```diff
@@ -10,13 +10,14 @@ class StockMove(models.Model):
     def _get_new_picking_values(self):
         vals = super(StockMove, self)._get_new_picking_values()
 
-        order_commitment_date = (
-            self.sale_line_id and self.sale_line_id.order_id.commitment_date
-        )
-
-        if order_commitment_date:
+        order_commitment_dates = [
```
mmequignon, line 13: We have to consider the `shipping_policy`.
If `all at once` use `max()`, if `asap` use `min()`

<s>Also, commitment date might be False in some cases. This isn't a mandatory field</s>
<s>
```python3
>>> max([False, datetime.now()])
  File "<stdin>", line 1, in <module>
TypeError: '>' not supported between instances of 'datetime.datetime' and 'bool'
```
</s>

edit: I missed the `if date` in the comprehension list

mmequignon (changes requested):

rousseldenis:
@TDu What's the status of this ?

Could you rebase ?

TDu:
@rousseldenis Rebase done. Status: it is being used, but need to investigate that comment.

PR #746, opened 2023-12-20
[16.0][IMP] improve logging

sbidoul:
Better use of logging so tools like sentry deduplicate the message.

### DIFF
--- a/base_delivery_carrier_label/models/stock_move_line.py
+++ b/base_delivery_carrier_label/models/stock_move_line.py
@@ -33,7 +33,7 @@ class StockMoveLine(models.Model):
             # if not defined we assume it's in kg
             if product.uom_id.id not in allowed:
                 _logger.warning(
-                    "Type conversion not implemented for product %s" % product.id
+                    "Type conversion not implemented for product %s", product.id
                 )
                 cant_calc_total = True
             # reserved_qty may be 0 if you don't set move line


sbidoul:
Red because of the ctt express integration is broken?

Why don't we use `vcr` to test such integration? At least tests are stable, and it's easy to update the cassettes when the upstream api evolves.

pedrobaeza:
We want to test the real thing. You will have it green very soon. Just retry in half an hour.

sbidoul:
> We want to test the real thing

With vcr you test the real thing. You just record it in your test so the test suite does not hit the network.

And any time you want you can re-record with the real thing. But at least you do it in a controlled manner and don't disturb unrelated contributions.

pedrobaeza:
But it's a frozen version of the reality. As said, we want to know if there's an API change or a problem in the service.

rvalyi:
For the record we had this questionning in l10n-brazil. A trade off we found is to use a special decorator so the test against the external server is performed only once a week and the other days a mocked version is tested.

The only big drawback of this approach is the test coverage percent changes a bit depending on the scenario, so PRs can be red because they have 0.1% test coverage...

sbidoul:
Then you need one repo per integration, with scheduled pipelines. Because as it is now you discover changes randomly when there are contributions on unrelated modules, and that is just annoying poor contributors like me :)

pedrobaeza:
And for me it's also very annoying to deal with all the PIP/wheel/etc stack random failing, but I have learned to live with it...

pedrobaeza:
They haven't still fixed it, so that means right now the service is not working for any customer. I have already told him, so I'll launch the merge command as soon as they fix it.

PR #755, opened 2024-01-05
[16.0][MIG] delivery_purchase: Migration to version 16.0

carolinafernandez-tecnativa:
- Standard migration to v16
- Tests are failed because this PR, depends on: https://github.com/OCA/delivery-carrier/pull/781

@Tecnativa
TT46563

@pedrobaeza @victoralmau


carolinafernandez-tecnativa:
Hi @glitchov
I have checked and the improvement added in this commit is not compatible with delivery_purchase.

https://github.com/OCA/delivery-carrier/commit/9c37fc2ea1af705e4a4d06067b64690b31d8ef79

The problem are with test test_put_in_pack_from_pick_with_wizard and test_put_in_pack_from_pick_without_wizard.
I added in setUpClass this code in order to fix this error message: "odoo.exceptions.UserError: There is no matching vendor price to generate the purchase order for product Test Product (no vendor defined, minimum quantity not reached, dates not valid, ...). Go on the product form and complete the list of vendors."

        seller = cls.env['product.supplierinfo'].create({
            'partner_id': cls.partner.id,
            'price': 12.0,
        })
        cls.product = cls.env["product.product"].create(
            {"name": "Test Product", "type": "product",
            'seller_ids': [(6, 0, [seller.id])],}
        )

After this, in both test when is trying to get pick_move = ship_move.move_orig_ids[0], move_orig_ids got a list index out of range,could you please check it? Thanks!!

@pedrobaeza do yo have any idea where this issue came from?

pedrobaeza:
Is the same problem without having this module?

carolinafernandez-tecnativa:
> Is the same problem without having this module?

Check now, if I comment both test it is working, so the problem is this improvement added on stock_picking_delivery_link https://github.com/OCA/delivery-carrier/commit/9c37fc2ea1af705e4a4d06067b64690b31d8ef79

pedrobaeza:
No, that's not the proof. Your module may be impacting on the other. You have to make a dummy PR without noticeable changes to see if CI is red. If it's green, then the module being added here is impacting the other.

victoralmau:
Can you continue with this migration?

pedrobaeza:
@victoralmau @carolinafernandez-tecnativa the other patch seems still not enough

victoralmau (approved):

pedrobaeza (approved):
/ocabot merge nobump

PR #756, opened 2024-01-05
16.0 temp check ci

carolinafernandez-tecnativa:


pedrobaeza:
You are including the commit history of `delivery_purchase` here, so it's not a valid test for checking if current branch is red.

PR #761, opened 2024-01-08
[FIX] delivery_package_type_number_parcels: make number of parcels computed

sbejaoui:
avoid related for this field as we don't want to change the pakcage type value when it changes in the package

cc/ @lmignon

### DIFF
--- a/delivery_package_type_number_parcels/models/stock_quant_package.py
+++ b/delivery_package_type_number_parcels/models/stock_quant_package.py
@@ -1,13 +1,17 @@
 # Copyright 2023 ACSONE SA/NV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class StockQuantPackage(models.Model):
     _inherit = "stock.quant.package"
 
     number_of_parcels = fields.Integer(
-        related="package_type_id.number_of_parcels",
-        store=True,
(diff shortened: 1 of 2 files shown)


delivery_package_type_number_parcels/models/stock_quant_package.py
```diff
@@ -1,13 +1,17 @@
...
+from odoo import api, fields, models
 
 
 class StockQuantPackage(models.Model):
     _inherit = "stock.quant.package"
 
     number_of_parcels = fields.Integer(
-        related="package_type_id.number_of_parcels",
-        store=True,
+        compute="_compute_number_of_parcels", store=True, readonly=True
```
lmignon:
```suggestion
        compute="_compute_number_of_parcels", store=True, readonly=False
```

lmignon (changes requested):
@sbejaoui IMO, the compute should be `readonly=False`

delivery_package_type_number_parcels/models/stock_quant_package.py
sbejaoui: done

rousseldenis (approved):
LGTM

lmignon (approved):

sbejaoui:
ready to merge

rousseldenis:
@sbejaoui Could you just add the module name in commit message ? Thanks

sbejaoui:
> @sbejaoui Could you just add the module name in commit message ? Thanks

done

PR #765, opened 2024-01-25
[MIG] delivery_automatic_package: Migration to 15.0

ghost:
Downgrade from 16.0


ghost:
I'm getting errors in the "delivery_cttexpress" does anybody have a hint what might be the problem. locally I also get the same errors even if I don't install "delivery_atuomatic_package" -> The module I'd like to migrate ...

PR #767, opened 2024-02-07
[14.0][IMP] delivery_carrier_return_barcode_pattern: Exclude inactive carriers

mt-software-de:
cc @TDu @jbaudoux

### DIFF
--- a/delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
+++ b/delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
@@ -57,5 +57,7 @@ class DeliveryCarrier(models.Model):
         if self.ids:
             pattern_query += "AND id IN %s"
             ids = tuple(self.ids)
+        else:
+            pattern_query += "AND active"
         pattern_query += ";"
         return (pattern_query, [ids])


delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
```diff
@@ -57,5 +57,7 @@ def _get_pattern_query(self):
         if self.ids:
             pattern_query += "AND id IN %s"
             ids = tuple(self.ids)
+        else:
+            pattern_query += "AND active"
```
mmequignon, line 61: shouldn't this be a default?
What if self contains inactive carriers?

mmequignon (commented):
LG, tiny question

delivery_carrier_return_barcode_pattern/models/delivery_carrier.py
mt-software-de: Hmm i think i someone requests it for an inactive carrier it should also work.
So i think it should not default.

TDu: From my understanding, it would exclude all inactive carrier only when searching for specific carrier ids.
So it makes sense.

TDu (approved):

mt-software-de:
ping @mmequignon

jbaudoux (approved):

mt-software-de:
> @jbaudoux your merge command was aborted due to failed check(s), which you can inspect on [this commit of 14.0-ocabot-merge-pr-767-by-jbaudoux-bump-patch](https://github.com/OCA/delivery-carrier/commits/38f72b599beef9a4067c8d96757e5beb488b5d76).
> 
> After fixing the problem, you can re-issue a merge command. Please refrain from merging manually as it will most probably make the target branch red.

@jbaudoux could you retry the merge
there was an error happening with an external api
https://github.com/OCA/delivery-carrier/actions/runs/8339620595/job/22821953901

PR #774, opened 2024-02-27
[16.0][IMP] delivery_driver: Flag driver in partner

EmilioPascual:
- When a driver is assigned in a shipping method, the partner is marked as a driver.
- Unmarking partners as drivers is not allowed if it is associated with a sending method.
- In pickings and move lines add domain in driver. Only partners marked as driver.
- When selecting a partner driver, only the name of the partner is displayed.
- If driver is changed in picking, also it's changed in origin picking (if you activated multi-steps routes)

MT-5232 @moduon

@rafaelbn @yajo @shide @gelojr please review.

PR #775, opened 2024-02-28
[IMP] delivery_driver_stock_picking_batch: Show only name in drivers

EmilioPascual:
- Assign the driver's user in batch if batch isn't wave.
- Show only name in drivers

PR Series: https://github.com/OCA/delivery-carrier/pull/774 (must be merged before this one)

MT-5232 @moduon

@shide @rafaelbn please review

### DIFF
--- a/delivery_driver_stock_picking_batch/views/stock_picking_batch.xml
+++ b/delivery_driver_stock_picking_batch/views/stock_picking_batch.xml
@@ -8,7 +8,12 @@
         <field name="inherit_id" ref="stock_picking_batch.view_picking_tree_batch" />
         <field name="arch" type="xml">
             <xpath expr="//field[@name='scheduled_date']" position="after">
-                <field name="driver_id" optional="hide" />
+                <field
+                    name="driver_id"
+                    optional="hide"
+                    context="{'show_driver': True}"
+                    options='{"always_reload": True}'
+                />
             </xpath>
         </field>
     </record>
@@ -18,7 +23,12 @@
         <field name="inherit_id" ref="stock_picking_batch.stock_picking_batch_form" />
         <field name="arch" type="xml">
             <xpath expr="//field[@name='scheduled_date']" position="after">
-                <field name="driver_ids" widget="many2many_tags" />
+                <field
+                    name="driver_ids"
+                    widget="many2many_tags"
+                    context="{'show_driver': True}"
+                    options='{"always_reload": True}'
+                />
             </xpath>
         </field>
     </record>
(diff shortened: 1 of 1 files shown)


delivery_driver_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -15,8 +15,19 @@ class StockPickingBatch(models.Model):
...
+        store=True,
+    )
 
     @api.depends("picking_ids.driver_id")
     def _compute_driver_ids(self):
         for batch in self:
             batch.driver_ids = batch.picking_ids.mapped("driver_id")
+
+    @api.depends("driver_ids")
+    def _compute_user_id(self):
```
rousseldenis: I'm not sure at all this is correct as this modifies the user_id values for all flows even without drivers.

IMHO, this should be readonly=False too.

delivery_driver_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -15,8 +15,19 @@ class StockPickingBatch(models.Model):
...
+        for batch in self.filtered(lambda b: not b.is_wave):
+            users = batch.mapped("driver_ids.user_ids")
+            if users:
+                batch.user_id = users[0]
```
rousseldenis: So, the user is the first driver if several mentionned on all pickings of the batch ??

rafaelbn:
Example:

- We have our own 16 trucks for delivery
- We have our clients that sometime wants more and sometimes less, so every client has its delivery method.
- But if a Delivery method has few kilos, the we join in the same batch different delivery method.
- One (1) single drivers make that in 1 truck. 
- Automatically assigned is the first but this could be changed later.

😄

delivery_driver_stock_picking_batch/models/stock_picking_batch.py
rafaelbn: @rousseldenis do you know other module that computes this? If driver is not set or this module is not installed this don't impact

delivery_driver_stock_picking_batch/models/stock_picking_batch.py
rafaelbn: Only one driver can drive a batch but as commented here: https://github.com/OCA/delivery-carrier/pull/775#issuecomment-1992038756 some times you want to join for get the maximun weight per truck so even initially multi-drivers should work, in this cases only one of them will make the delivery

EmilioPascual:
@rafaelbn although you are right, I think the functionality may be controversial and not suitable for all workflows. I have decided to move the functionality to our repository.

EmilioPascual:
Failed tests are for another module: https://github.com/OCA/delivery-carrier/actions/runs/8263936917/job/22606437935?pr=775#step:8:335

rafaelbn (approved):
Ready to merge in my side 👍🏼 

> @rafaelbn although you are right, I think the functionality may be controversial and not suitable for all workflows. I have decided to move the functionality to our repository.

Perfect.

Shide (approved):

PR #777, opened 2024-03-05
[15.0][IMP] delivery_package_number: Ask only once for the number of packages in wizard.

sergio-teruel:
Since the wizard was created to ask user if want a number of package is enouht, not need to ask again in backorder wizard or inmediate transfer wizard.

cc @Tecnativa TT48170

ping @CarlosRoca13 @chienandalu


chienandalu (approved):
:+1: I just wonder it this could suppose a behavior change for those relaying on those wizards to set the package number.

CarlosRoca13 (approved):

sergio-teruel:
Hi @chienandalu Maybe... Do you know any module??

pedrobaeza:
One of them was added in #705 

@angelgarciadelachica @HaraldPanten which was the reason for adding it (double or not)?

chienandalu:
> Hi @chienandalu Maybe... Do you know any module??

Not really a module. When you don't set to force the number of packages on the operation type you could always set the packages number on the wizards...

Anyway, I think it's cleaner this way.

angelgarciadelachica:
Hi @sergio-teruel 
I don't really see where the duplicity is... The wizard pops up to make sure that the user enters the number of packages when there is a carrier. There are two possible cases:
- If you validate the whole picking and there is a carrier selected-> It asks how many packages are in that picking.
- If you want to create a backorder and there is a carrier selected-> It asks how many packages are in that backorder.

With this PR you eliminate this step of confirmation of number of packages completely? I don't think it's a good idea to remove it, as many users are used to enter the number of packages from these wizards.

On the other hand, I have seen that you also eliminate the creation of the labels from the assistant, why is this?

THX!

sergio-teruel:
Hi @angelgarciadelachica 

I will try to explain the reason for this.

There is a case that this module doesn't cover... Try to configure manually all the units in the picking.... It doesn't launch any wizard so the user can't configure the field number of packages...

To solve this... we create the new wizard to pick up if needed so I think we are keeping similar code in three wizards so the best option is to have only one to report the package numbers....

I attach a gif to explain when the two wizards don't get raised...

Peek 16-03-2024 07-01

I attach other gif to display the picking validation process 

Peek 16-03-2024 07-46

sergio-teruel:
Hi... 

I attach other gif with multi pick validation

Peek 16-03-2024 08-04

HaraldPanten:
@sergio-teruel @pedrobaeza I think you have a wrong contributor in Tecnativa's list of contributors in this module. Shouldn't that be fixed/removed?

@angelgarciadelachica Could you check Sergio's comments?

THX.

pedrobaeza:
> I think you have a wrong contributor in Tecnativa's list of contributors in this module. Shouldn't that be fixed/removed?

Yes, indeed it's incorrect. @sergio-teruel can you remove it?

sergio-teruel:
Done!!

angelgarciadelachica:
Hi @sergio-teruel!

Thank you very much for the explanation, it is perfectly understood. 
LGTM 👍🏻

THX!

PR #599, opened 2023-01-16
[16.0][MIG] delivery_carrier_label_gls

glitchov:
Migration from a 10.0

PR #778, opened 2024-03-07
[14.0] [MIG] delivery_carrier_label_gls

paradoxxxzero:
Backport to 14.0 from the 16.0 migration PR: #599 

Also fix a problem with the carrier tracking url being wrongly coerced to a non json parseable string.


delivery_carrier_label_gls/migrations/16.0.1.0.0/post-migrate.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2023 ACSONE SA/NV
```
sebastienbeau, line 1: Rename version ? 16.0.1.0.0 => 14.0.0.0

PR #779, opened 2024-03-07
 [17.0][MIG] delivery_carrier_info

AndreuOForgeFlow:
Standard Migration

@ForgeFlow


delivery_carrier_info/migrations/16.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,18 @@
...
+            """
+            SELECT id, description
+            FROM delivery_carrier
+            WHERE description IS NOT NULL AND carrier_description IS NULL
+        """
+        )
+        for carrier_id, description in env.cr.fetchall():
+            env["delivery.carrier"].browse(carrier_id).write(
+                {"carrier_description": description}
+            )
```
GuillemCForgeFlow:
```suggestion
```
When migrating to the new version, we need to remove the migration files from the previous one. We just need to add some, if they are needed for the migration to v17, in this case.

GuillemCForgeFlow (changes requested):

GuillemCForgeFlow (approved):
LGTM! Code and functional review 👍🏽

rousseldenis (approved):
Code review.

I know this has been discussed but I'm wondering if this deserves a base module to add a group in forms and then add 'code' and so on...

@florian-dacosta

PR #780, opened 2024-03-12
[Fix] delivery_auto_refresh manually delete delivery line

santostelmo:
# Context
When deleting manually delivery line we get the following:
delivery_line_delete_error

### DIFF
--- a/delivery_auto_refresh/tests/test_delivery_auto_refresh.py
+++ b/delivery_auto_refresh/tests/test_delivery_auto_refresh.py
@@ -325,3 +325,11 @@ class TestDeliveryAutoRefresh(common.TransactionCase):
         order = order_form.save()
         delivery_line = order.order_line.filtered("is_delivery")
         self.assertFalse(delivery_line.exists())
+
+    def test_auto_refresh_so_and_manually_unlink_delivery_line(self):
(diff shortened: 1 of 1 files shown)


delivery_auto_refresh/models/sale_order.py
```diff
@@ -72,6 +73,12 @@ def _auto_refresh_delivery(self):
                     {"recompute_delivery_price": False}
                 )
 
+    def _order__delete_delivery_lines_after_commit(self, line_ids):
+        self.with_context(auto_refresh_delivery=True)._remove_delivery_line()
+        # _logger.debug("%s deleted after commit", recs)
+        # Needs an explicit commit as it runs after the original commit is done
```
rousseldenis: Where ?

santostelmo: Sorry I was with the PR under update.

Now its okay. I expect the unit test to fail. And for the moment I am not sure how to fix that

santostelmo:
@rousseldenis the unit test self describes the issue that we are facing

santostelmo:
Closed this PR. After a second thought  I think we are not suppose to delete delivery line. 
In the other hand we should not face this error if we try todo it

PR #781, opened 2024-03-13
[16.0][FIX] stock_picking_delivery_link: Test compatibility with purchase

victoralmau:
Related to https://github.com/OCA/delivery-carrier/pull/755 (https://github.com/OCA/delivery-carrier/actions/runs/8269746616/job/22625732737)

Avoid test errors if the purchase module is installed

Please @pedrobaeza and @carolinafernandez-tecnativa can you review it?

@Tecnativa TT46563

### DIFF
--- a/stock_picking_delivery_link/tests/test_delivery_link.py
+++ b/stock_picking_delivery_link/tests/test_delivery_link.py
@@ -26,6 +26,16 @@ class TestStockPickingDeliveryLink(StockPickingDeliveryLinkCommonCase):
                 "product_id": test_carrier_product.id,
             }
         )
+        # We need to know if purchase module is installed
+        cls.purchase_installed = False
+        if "purchased_product_qty" in cls.env["product.product"]._fields:
+            cls.purchase_installed = True
+            cls.vendor = cls.env["res.partner"].create({"name": "Test vendor"})
+            cls.product.write(
+                {
+                    "seller_ids": [(0, 0, {"partner_id": cls.vendor.id})],
+                }
+            )
 
     def test_ship_data_from_pick(self):
         move1 = self._create_move(
@@ -123,6 +133,12 @@ class TestStockPickingDeliveryLink(StockPickingDeliveryLinkCommonCase):
         )
         ship_move._assign_picking()
         ship_move._action_confirm()
+        # Confirm purchase order
+        if self.purchase_installed:
+            purchase = self.env["purchase.order"].search(
+                [("partner_id", "=", self.vendor.id)]
+            )
+            purchase.button_confirm()
         pick_move = ship_move.move_orig_ids[0]
         pick_picking = pick_move.picking_id
         ship_picking = ship_move.picking_id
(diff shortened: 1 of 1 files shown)


rousseldenis:
@victoralmau I don't like putting some code that depends on module that is not in dependencies...

victoralmau:
> @victoralmau I don't like putting some code that depends on module that is not in dependencies...

Ok, I understand, any suggestions to avoid tests failing if purchase is installed?

carolinafernandez-tecnativa:
I have checked and it looks good to me, what do you think @pedrobaeza ?

rousseldenis:
> > @victoralmau I don't like putting some code that depends on module that is not in dependencies...
> 
> Ok, I understand, any suggestions to avoid tests failing if purchase is installed?

@victoralmau Don't check purchase module. Fill in simply the seller_ids field as this is implemented in product.

victoralmau:
> > > @victoralmau I don't like putting some code that depends on module that is not in dependencies...
> > 
> > 
> > Ok, I understand, any suggestions to avoid tests failing if purchase is installed?
> 
> @victoralmau Don't check purchase module. Fill in simply the seller_ids field as this is implemented in product.



> > > @victoralmau I don't like putting some code that depends on module that is not in dependencies...
> > 
> > 
> > Ok, I understand, any suggestions to avoid tests failing if purchase is installed?
> 
> @victoralmau Don't check purchase module. Fill in simply the seller_ids field as this is implemented in product.

Even if we do that we will still have to confirm the purchase and the tests will fail. Please try the following use case:

- Install `stock_picking_delivery_link` + `purchase`.
- Run the tests of `stock_picking_delivery_link`

victoralmau:
Any other proposal to solve the tests of this module? This blocks https://github.com/OCA/delivery-carrier/pull/755

victoralmau:
Any other **better** proposals to solve this problem? This blocks several migrations.

carolinafernandez-tecnativa:
Can we procedure to merge? So we can unblock several migrations. Thanks!

carolinafernandez-tecnativa:
Can we move forward to merge? @pedrobaeza @rousseldenis

carolinafernandez-tecnativa:
@pedrobaeza What do you think?

stock_picking_delivery_link/tests/test_delivery_link.py
```diff
@@ -26,6 +26,16 @@ def setUpClass(cls):
                 "product_id": test_carrier_product.id,
             }
         )
+        # We need to know if purchase module is installed
+        cls.purchase_installed = False
```
etobella, line 30: Maybe it is cleaner to use 
```suggestion
        cls.purchase_installed = cls.env.ref("base.module_purchase").state == "installed"
```

pedrobaeza:
@rousseldenis I have been 2 hours trying to get another solution in #755, but I see no other solution than this one.

The problem is that the tests `test_put_in_pack_from_pick_with_wizard*` of `stock_picking_delivery_link` works by chance: they create the stock.move with procure_method = "make_to_order", launching a procurement that is resolved with a lucky fallback to the rule "WH: Vendors → Stock", which creates an extra move for resolving `move_orig_ids`, but that's not possible when `purchase_stock` gets installed (or at least I'm not able to force it).

Can you propose an alternative or to not block this?

carolinafernandez-tecnativa:
Hi, @rousseldenis have you checked what @pedrobaeza commented before?

carolinafernandez-tecnativa:
Hi, any news? we need to unlock several migrations

pedrobaeza (approved):
Let's merge this as the less bad solution. Later, other options can be proposed:

/ocabot merge patch

PR #720, opened 2023-10-12
[14.0][FIX] delivery_send_to_shipper_at_operation

jbaudoux:
When duplicating an already send picking, allow to register it at the shipper

Includes: https://github.com/OCA/delivery-carrier/pull/734

PR #782, opened 2024-03-14
[14.0][FIX] delivery_send_to_shipper_at_operation: Keep delivery_notification_sent set on backorder

mt-software-de:
If a backorder is created out of a transfer which is marked as delivery_notification_sent = True.
The backorder should also be marked as delivery_notification_sent. 

@jbaudoux @TDu 

Depends on:
Fixing tests
https://github.com/OCA/delivery-carrier/pull/786

### DIFF
--- a/delivery_send_to_shipper_at_operation/models/stock_picking.py
+++ b/delivery_send_to_shipper_at_operation/models/stock_picking.py
@@ -1,5 +1,6 @@
 # Copyright 2021 Camptocamp SA
 # Copyright 2023 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+# Copyright 2024 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
 from lxml import etree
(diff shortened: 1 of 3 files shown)


jbaudoux (commented):
Why can't we simply revert this https://github.com/OCA/delivery-carrier/pull/720 ?

mt-software-de:
> Why can't we simply revert this #720 ?

No it is fine to not copy the value, because move_dest_ids/move_orig_ids will also not copied from a move. 
In a backorder case the move_orig_ids will be the same as in the new backorder

jbaudoux:
@mt-software-de I should have added a unittest for my fix https://github.com/OCA/delivery-carrier/pull/720. If you do a partial pack, then you don't want to keep `delivery_notification_sent` on the pack backorder.
So, it's a bit more complex...

mt-software-de:
> @mt-software-de I should have added a unittest for my fix #720. If you do a partial pack, then you don't want to keep `delivery_notification_sent` on the pack backorder. So, it's a bit more complex...

For the pack operation backorder it is fine, to not set the `delivery_notification_sent`
therefore we need to remove
https://github.com/OCA/delivery-carrier/blob/1ec54201e1b6b21761dfc7e831d6a4a3a4990450/delivery_send_to_shipper_at_operation/models/stock_picking.py#L54

mt-software-de:
@jbaudoux could you please add the label "no stale". This is waiting for https://github.com/OCA/delivery-carrier/pull/786 or https://github.com/OCA/delivery-carrier/pull/854.

mt-software-de:
> @jbaudoux could you please add the label "no stale". This is waiting for #786 or #854.

Since there is no real progress, i added the context value `set_default_package` from `base_delivery_carrier_label`, this ensures the tests are working. This could be undone when the mentioned PR's are done. 

@jbaudoux @i-vyshnevska

jbaudoux (approved):

PR #783, opened 2024-03-14
[FIX] delivery_driver: wrong model name

yajo:
Regression from https://github.com/OCA/delivery-carrier/pull/774:

```

  File "/opt/odoo/auto/addons/delivery_driver/migrations/16.0.1.0.1/post-migration.py", line 15, in migrate
    env["carrier.driver"]
  File "/opt/odoo/custom/src/odoo/odoo/api.py", line 549, in __getitem__
    return self.registry[model_name](self, (), ())
  File "/opt/odoo/custom/src/odoo/odoo/modules/registry.py", line 190, in __getitem__
    return self.models[model_name]
KeyError: 'carrier.driver'
Error: 'carrier.driver'
```

@moduon MT-5232

### DIFF
--- a/delivery_driver/migrations/16.0.1.0.1/post-migration.py
+++ b/delivery_driver/migrations/16.0.1.0.1/post-migration.py
@@ -12,7 +12,7 @@ def migrate(cr, version):
     with api.Environment.manage():
         env = api.Environment(cr, SUPERUSER_ID, {})
         partners = (
-            env["carrier.driver"]
+            env["delivery.carrier"]
             .search([("driver_id", "!=", False)])
             .mapped("driver_id")
         )


rousseldenis (approved):

rousseldenis:
Tests are failing because of that sh... CTTExpress module

yajo:
It seems one of those cases where a unit test has been confused with a monitoring tool. Now what? This issue is pretty critical...

pedrobaeza:
The test in CTT is not intended for monitoring, but for assuring that the API doesn't change and the module stops to work. Mocking that will lead to fail in that case but green CIs. Anyway, it's true that the demo environment against the tests is done is a bit unstable, but it's usually question of minutes.

You can meanwhile add the PR in your deployment, as it's not critical, as it doesn't affect any new installation nor anyone not using the module.

rousseldenis:
> It seems one of those cases where a unit test has been confused with a monitoring tool. Now what? This issue is pretty critical...

In ten years, we'll laugh about that around a drink :smiley: 

But, as already said, that one deserves a separate repo to not bother other people testing flows.

yajo:
I think we need a guideline about flaky tests, so here's my proposal: https://github.com/OCA/odoo-community.org/pull/158

PR #784, opened 2024-03-14
[16.0][FIX] delivery_driver: constraint in multi-edit

EmilioPascual:
Fix in constraint when editing is_driver in several partner

@yajo @Shide @rafaelbn please review

MT-5232 @moduon

### DIFF
--- a/delivery_driver/models/res_partner.py
+++ b/delivery_driver/models/res_partner.py
@@ -16,7 +16,9 @@ class ResPartner(models.Model):
         drivers_in_carrier = self.env["delivery.carrier"].search(
             [("driver_id", "in", self.ids)]
         )
-        if drivers_in_carrier and not self.is_driver:
+        if drivers_in_carrier and drivers_in_carrier.filtered_domain(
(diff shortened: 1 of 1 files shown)


yajo (approved):

delivery_driver/models/res_partner.py
```diff
@@ -16,7 +16,9 @@ def _check_is_driver(self):
         drivers_in_carrier = self.env["delivery.carrier"].search(
             [("driver_id", "in", self.ids)]
         )
-        if drivers_in_carrier and not self.is_driver:
+        if drivers_in_carrier and drivers_in_carrier.driver_id.filtered_domain(
+            [("is_driver", "=", False)]
```
Shide: Will fail if drivers_in_carriers are more than one.

```suggestion
        if drivers_in_carrier and drivers_in_carrier.filtered_domain(
            [("driver_id.is_driver", "=", False)]
```

Shide (changes requested):

Shide (approved):

PR #787, opened 2024-03-19
[16.0][IMP] delivery_deliverea: add exchange attribute in serviceAttributes 

imoralesc:
Add exchange service attribute in _get_service_attributes method. 

Selección_245

### DIFF
--- a/delivery_deliverea/models/delivery_carrier.py
+++ b/delivery_deliverea/models/delivery_carrier.py
@@ -107,6 +107,12 @@ class DeliveryCarrier(models.Model):
         comodel_name="carrier.deliverea.service",
         domain="[('deliverea_distribution_center_id', '=', deliverea_distribution_center_id)]",
     )
+    deliverea_exchange = fields.Boolean(
+        string="Exchange",
+        help="Mark expedition as exchange (when delivering the package,"
+        " another package has to be picked up from the final client and will be sent"
+        " back to the original sender. Availability depends on carrier and service)",
+    )
 
     def deliverea_get_distribution_centers(self):
         deliverea_request = DelivereaRequest(self)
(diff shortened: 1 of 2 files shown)


OscarIndias (approved):
LGTM

imoralesc:
Hi @pedrobaeza, could you check this PR?

pedrobaeza:
Please use a proper commit message:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message

imoralesc:
> Please use a proper commit message:
> 
> https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message

Done :wink: @pedrobaeza

pedrobaeza:
`[UPD]` is not a valid tag. You have `[IMP]` or `[FIX]`.

imoralesc:
> `[UPD]` is not a valid tag. You have `[IMP]` or `[FIX]`.

Done @pedrobaeza

pedrobaeza (commented):
Is this something reasonable to be put at carrier level? Shouldn't it be at picking level?

imoralesc:
> Is this something reasonable to be put at carrier level? Shouldn't it be at picking level?

Hi @pedrobaeza, it must be at carrier level because it is a shipment where the carrier delivers one parcel and collects another at the time of delivery.

pedrobaeza:
shipment = picking, so that's why I say it. If not, you are force to have 2 carriers for the same shipper.

imoralesc:
@pedrobaeza it is mandatory to configure it as two carriers, they are completely different operations.

The customer uses that type of transport for a specific case, and also this attribute depends on the carrier. Not all carriers are compatible with this attribute, in other carriers it is a simple service.

pedrobaeza:
> The customer uses that type of transport for a specific case

Yes, and for that case, you can mark the check instead of selecting the other carrier.

> and also this attribute depends on the carrier.

Not sure about this affirmation. In the image screenshot, the text says "Mark expedition as exchange", so it seems an attribute per expedition, not per carrier or service.

> Not all carriers are compatible with this attribute

Correct, but this can work the same as others attribute that declares if they are capable or not.

PR #795, opened 2024-04-16
[15.0][FIX] delivery_package_number: download label

chienandalu:
After #777 the only wizard for package labels printing was this. But it's method isn't return the printing action to the web client and thus the label isn't printed.

cc @Tecnativa TT48759

please review @sergio-teruel as I think this could impact your threaded printing developements


chienandalu:
@pedrobaeza unstale, please

@sergio-teruel we'll have to do something about this...

sergio-teruel (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #791, opened 2024-03-27
[14.0] [FIX] delivery_auto_refresh: manually delete delivery line

renda-dev:
supersedes https://github.com/OCA/delivery-carrier/pull/780 for version 14.0

PR #790, opened 2024-03-27
[14.0] [FIX] delivery_auto_refresh: dynamic carrier line

renda-dev:
- Re-proposing the fix that has been made in https://github.com/OCA/delivery-carrier/pull/695 and removed in https://github.com/OCA/delivery-carrier/pull/656
- Fixed a test that should have failed when this feature was removed
- Proposing me and my team as maintainers for this module

PR #797, opened 2024-04-16
[16.0][FIX] delivery_auto_refresh: dynamic carrier line

toita86:
fw port of
* https://github.com/OCA/delivery-carrier/pull/790
* https://github.com/OCA/delivery-carrier/pull/791


pedrobaeza (approved):

jbaudoux (changes requested):
Thanks for the forward port.

In my opinion, there are many issues in this module with how context variables are given and managed. The added code is increasing the complexity of the module instead of solving the source of the issue.

I have opened another PR https://github.com/OCA/delivery-carrier/pull/799 that fixes among other the way code is executed. I have taken over the tests change and addition from here.

PR #323, opened 2020-12-11
[REF] delivery_carrier_label_postlogistics becomes delivery_postlogistics in 13.0

phuctranfxvn:
Description:
- New module `delivery_postlogistics` as a refactoring of module `delivery_carrier_label_postlogistics` to get it closer to the native `delivery` module

PR #798, opened 2024-04-16
[16.0][FIX] delivery_postlogistics: splitting package label by package

imlopes:
### Problem : 

In a picking, if we create 2 or more packages, after generating the postlogistic labels, all tracking numbers are concatenated on the first package.

image

image

It seems that this behavior was added here :
https://github.com/OCA/delivery-carrier/commit/5504c9f26dae2739abb68314bab7ffb2756f332d

More precisely here : https://github.com/OCA/delivery-carrier/blob/5504c9f26dae2739abb68314bab7ffb2756f332d/delivery_postlogistics/postlogistics/web_service.py#L511C1-L518C14

and here : 
https://github.com/OCA/delivery-carrier/pull/323/files#diff-d3c06dad2d58ebdeb558386909ed5b2ad9ed22d84b8476e5fb2658ae4308d61f

I didn't identify why the code was changed before versions, but now we are always taking the first package when preparing the label values that will be used on ``write_tracking_number_label`` method.

### DIFF
--- a/delivery_postlogistics/postlogistics/web_service.py
+++ b/delivery_postlogistics/postlogistics/web_service.py
@@ -568,7 +568,7 @@ class PostlogisticsWebService(object):
             res["success"] = True
             res["value"].append(
                 {
-                    "item_id": item_list[0]["itemID"],
+                    "item_id": item["itemID"],
                     "binary": binary,
                     "tracking_number": response_dict["item"]["identCode"],
                     "file_type": file_type,


rousseldenis:
@imlopes Thanks for this. Could you maybe add a test to avoid regression ?

ivantodorovich (approved):
LG

I agree it'd be nice to have unit tests

sebalix (approved):
LG, indeed a test is welcome

PR #799, opened 2024-04-17
[16.0][IMP-FIX] delivery_auto_refresh - performance & compatibility & cleanup

jbaudoux:
In summary:
* Cleanup code (use a single context variable to prevent re-execution, make dedicated methods for each purpose).
* Increase performance (there was many recursive execution of the same delete/add line - code is now executed once and only what needs to be modified). 
* Make the module compatible with sale_order_carrier_auto_assign. ~~Move carrier assignment to that module so that this module concentrate on cost delivery line update (i.e. refresh as from its name). See https://github.com/OCA/sale-workflow/pull/3081~~ Marked as migration v17.0 improvement
* Make module more robust. Implement SO line create/write

Depends on:
* ~~https://github.com/OCA/sale-workflow/pull/3081~~

Settings:
image

In details (one commit for each) - in reverse order:
* config parameters moved to company. "sale_" prefix added to the parameters
* add tests from https://github.com/OCA/delivery-carrier/pull/790 & https://github.com/OCA/delivery-carrier/pull/791
* improve help & readme
* refresh on SO line create/write
* set carrier: ~~use https://github.com/OCA/sale-workflow/pull/3081~~ integrated here and marked to move on v17.0
* refresh the minimum: Create or delete line only if necessary. Update existing line if necessary
* iterate over list, not set
* fix write & discount
* Don't pass discount variable in context
* fix create in batch (api.model_create_multi instead of api.model)
* docstring: Don't modify standard method docstring

Compatibility checks/fixes with other modules triggering refresh on SO create/write:
- [x] https://github.com/OCA/sale-workflow/tree/16.0/sale_product_packaging_container_deposit
    https://github.com/OCA/product-attribute/pull/1583
- [x] https://github.com/OCA/sale-workflow/tree/16.0/sale_order_carrier_auto_assign
   Solved
- [x] https://github.com/OCA/sale-promotion/pull/208
    Looks good
- [x] https://github.com/OCA/sale-workflow/tree/16.0/sale_product_multi_add
    https://github.com/OCA/sale-workflow/pull/3084

cc @pedrobaeza @santostelmo @yvaucher @simahawk @francesco-ooops @renda-dev

### DIFF
--- a/delivery_auto_refresh/__manifest__.py
+++ b/delivery_auto_refresh/__manifest__.py
@@ -11,6 +11,8 @@
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "depends": ["delivery", "sale_order_carrier_auto_assign"],
+    "depends": ["delivery"],
+    # Migration Note 17.0: Add dependency to sale_order_carrier_auto_assign
+    # "depends": ["delivery", "sale_order_carrier_auto_assign"],
     "data": ["views/sale_order_views.xml", "views/res_config_settings_views.xml"],
 }
--- a/delivery_auto_refresh/__manifest__.py
+++ b/delivery_auto_refresh/__manifest__.py
@@ -4,13 +4,13 @@
 {
     "name": "Auto-refresh delivery",
     "summary": "Auto-refresh delivery price in sales orders",
-    "version": "16.0.1.0.1",
+    "version": "16.0.2.0.0",
     "category": "Delivery",
     "website": "https://github.com/OCA/delivery-carrier",
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "depends": ["delivery"],
+    "depends": ["delivery", "sale_order_carrier_auto_assign"],
     "data": ["views/sale_order_views.xml", "views/res_config_settings_views.xml"],
 }
--- a/delivery_auto_refresh/migrations/16.0.2.0.0/post-migration.py
+++ b/delivery_auto_refresh/migrations/16.0.2.0.0/post-migration.py
@@ -8,7 +8,7 @@ def _migrate_setting_to_company(env):
     if env["ir.config_parameter"].get_param(
         "delivery_auto_refresh.set_default_carrier"
     ):
-        env["res.company"].search([]).carrier_auto_assign_on_create = True
+        env["res.company"].search([]).sale_auto_assign_carrier_on_create = True
 
 
 def migrate(cr, version):
--- a/delivery_auto_refresh/migrations/16.0.2.0.0/post-migration.py
+++ b/delivery_auto_refresh/migrations/16.0.2.0.0/post-migration.py
@@ -1,0 +1,16 @@
+# Copyright 2024 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import SUPERUSER_ID, api
+
+
+def _migrate_setting_to_company(env):
+    if env["ir.config_parameter"].get_param(
+        "delivery_auto_refresh.set_default_carrier"
+    ):
+        env["res.company"].search([]).carrier_auto_assign_on_create = True
+
+
+def migrate(cr, version):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    _migrate_setting_to_company(env)
--- a/delivery_auto_refresh/migrations/16.0.2.0.0/post-migration.py
+++ b/delivery_auto_refresh/migrations/16.0.2.0.0/post-migration.py
@@ -10,6 +10,21 @@ def _migrate_setting_to_company(env):
     ):
         env["res.company"].search([]).sale_auto_assign_carrier_on_create = True
 
+    if env["ir.config_parameter"].get_param(
+        "delivery_auto_refresh.auto_add_delivery_line"
+    ):
+        env["res.company"].search([]).sale_auto_add_delivery_line = True
+
+    if env["ir.config_parameter"].get_param(
+        "delivery_auto_refresh.refresh_after_picking"
+    ):
+        env["res.company"].search([]).sale_refresh_delivery_after_picking = True
+
+    if env["ir.config_parameter"].get_param(
+        "delivery_auto_refresh.auto_void_delivery_line"
+    ):
+        env["res.company"].search([]).sale_auto_void_delivery_line = True
+
 
 def migrate(cr, version):
     env = api.Environment(cr, SUPERUSER_ID, {})
--- a/delivery_auto_refresh/models/res_config_settings.py
+++ b/delivery_auto_refresh/models/res_config_settings.py
@@ -14,15 +14,15 @@ class ResConfigSettings(models.TransientModel):
     )
     # End migration note
 
-    auto_add_delivery_line = fields.Boolean(
-        "Refresh shipping cost line automatically",
-        config_parameter="delivery_auto_refresh.auto_add_delivery_line",
+    sale_auto_add_delivery_line = fields.Boolean(
+        related="company_id.sale_auto_add_delivery_line",
+        readonly=False,
     )
-    refresh_after_picking = fields.Boolean(
-        "Refresh after picking automatically",
-        config_parameter="delivery_auto_refresh.refresh_after_picking",
+    sale_refresh_delivery_after_picking = fields.Boolean(
+        related="company_id.sale_refresh_delivery_after_picking",
+        readonly=False,
     )
-    auto_void_delivery_line = fields.Boolean(
-        "Void delivery lines automatically",
-        config_parameter="delivery_auto_refresh.auto_void_delivery_line",
+    sale_auto_void_delivery_line = fields.Boolean(
+        related="company_id.sale_auto_void_delivery_line",
+        readonly=False,
     )
--- a/delivery_auto_refresh/models/res_config_settings.py
+++ b/delivery_auto_refresh/models/res_config_settings.py
@@ -7,16 +7,12 @@ from odoo import fields, models
 class ResConfigSettings(models.TransientModel):
     _inherit = "res.config.settings"
 
-    set_default_carrier = fields.Boolean(
-        "Set Default Carrier Automatically",
-        config_parameter="delivery_auto_refresh.set_default_carrier",
-    )
     auto_add_delivery_line = fields.Boolean(
-        "Add Delivery Line Automatically",
+        "Refresh shipping cost line automatically",
         config_parameter="delivery_auto_refresh.auto_add_delivery_line",
     )
     refresh_after_picking = fields.Boolean(
-        "Refresh After Picking Automatically",
+        "Refresh after picking automatically",
         config_parameter="delivery_auto_refresh.refresh_after_picking",
     )
     auto_void_delivery_line = fields.Boolean(
--- a/delivery_auto_refresh/models/res_config_settings.py
+++ b/delivery_auto_refresh/models/res_config_settings.py
@@ -7,6 +7,13 @@ from odoo import fields, models
 class ResConfigSettings(models.TransientModel):
     _inherit = "res.config.settings"
 
+    # Migration Note 17.0: move this to module sale_order_carrier_auto_assign
+    sale_auto_assign_carrier_on_create = fields.Boolean(
+        related="company_id.sale_auto_assign_carrier_on_create",
+        readonly=False,
+    )
+    # End migration note
+
     auto_add_delivery_line = fields.Boolean(
         "Refresh shipping cost line automatically",
         config_parameter="delivery_auto_refresh.auto_add_delivery_line",
--- a/delivery_auto_refresh/models/sale_order.py
+++ b/delivery_auto_refresh/models/sale_order.py
@@ -57,18 +57,32 @@ class SaleOrder(models.Model):
             .get_param("delivery_auto_refresh.auto_add_delivery_line")
         )
 
+    def _get_delivery_discount(self):
+        self.ensure_one()
+        delivery_lines = self.order_line.filtered("is_delivery")
+        return delivery_lines[-1:].discount
+
     def _auto_refresh_delivery(self):
         self.ensure_one()
+        if self.env.context.get("auto_refresh_delivery"):
+            return
+        if not self._get_param_auto_add_delivery_line():
+            return
+
+        delivery_discount = self._get_delivery_discount()
+
         # Make sure that if you have removed the carrier, the line is gone
         if self.state in {"draft", "sent"}:
             # Context added to avoid the recursive calls and save the new
             # value of carrier_id
             self.with_context(auto_refresh_delivery=True)._remove_delivery_line()
-        if self._get_param_auto_add_delivery_line() and self.carrier_id:
+        if self.carrier_id:
             if self.state in {"draft", "sent"}:
                 price_unit = self.carrier_id.rate_shipment(self)["price"]
                 if not self.is_all_service:
-                    self._create_delivery_line(self.carrier_id, price_unit)
+                    sol = self._create_delivery_line(self.carrier_id, price_unit)
+                    if delivery_discount and sol:
+                        sol.discount = delivery_discount
                 self.with_context(auto_refresh_delivery=True).write(
                     {"recompute_delivery_price": False}
                 )
@@ -82,25 +96,13 @@ class SaleOrder(models.Model):
         return orders
 
     def write(self, vals):
-        # Create or refresh delivery line after saving
-        res = super().write(vals)
-        if self._get_param_auto_add_delivery_line() and not self.env.context.get(
-            "auto_refresh_delivery"
-        ):
-            for order in self:
-                delivery_line = order.order_line.filtered("is_delivery")
-                order.with_context(
-                    delivery_discount=delivery_line[-1:].discount,
-                )._auto_refresh_delivery()
+        # Prevent to refresh delivery in the call to super
+        res = super(SaleOrder, self.with_context(auto_refresh_delivery=True)).write(
+            vals
+        )
+        for order in self:
+            order._auto_refresh_delivery()
         return res
-
-    def _create_delivery_line(self, carrier, price_unit):
-        # Allow users to keep discounts to delivery lines. Unit price will be recomputed anyway
-        sol = super()._create_delivery_line(carrier, price_unit)
-        discount = self.env.context.get("delivery_discount")
-        if discount and sol:
-            sol.discount = discount
-        return sol
 
     def set_delivery_line(self, carrier, amount):
         if self._get_param_auto_add_delivery_line() and self.state in {"draft", "sent"}:
--- a/delivery_auto_refresh/models/sale_order.py
+++ b/delivery_auto_refresh/models/sale_order.py
@@ -72,12 +72,12 @@ class SaleOrder(models.Model):
         delivery_discount = self._get_delivery_discount()
 
         # Make sure that if you have removed the carrier, the line is gone
-        if self.state in {"draft", "sent"}:
+        if self.state in ("draft", "sent"):
             # Context added to avoid the recursive calls and save the new
             # value of carrier_id
             self.with_context(auto_refresh_delivery=True)._remove_delivery_line()
         if self.carrier_id:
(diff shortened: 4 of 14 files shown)


delivery_auto_refresh/__manifest__.py
```diff
@@ -4,13 +4,13 @@
...
-    "version": "16.0.1.0.1",
+    "version": "16.0.2.0.0",
     "category": "Delivery",
     "website": "https://github.com/OCA/delivery-carrier",
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "depends": ["delivery"],
+    "depends": ["delivery", "sale_order_carrier_auto_assign"],
```
pedrobaeza: You can't add a new dependency in this module in stable, but also because those using this module since the beginning don't want to have a new one.

pedrobaeza (changes requested):

delivery_auto_refresh/__manifest__.py
jbaudoux: Luckily it is in Beta stage and not Stable/Production
Don't you find it better to have the features implemented in the proper corresponding modules?

pedrobaeza: It's not really beta. It's that nobody has changed its maturity level (because nobody is worrying about that). I prefer to have it as is, and do the change on the next version.

jbaudoux: @pedrobaeza I removed the dependency. I marked the code to drop in v17.0. I placed this in a separate commit.

I still have to change the config parameters to make them configurable per company

pedrobaeza: OK, thanks for the efforts. Please include migration script for converting them smoothly out of the system parameters.

delivery_auto_refresh/migrations/16.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2024 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import SUPERUSER_ID, api
+
+
+def _migrate_setting_to_company(env):
+    if env["ir.config_parameter"].get_param(
+        "delivery_auto_refresh.set_default_carrier"
```
simahawk, line 9: don't we get rid of the parameter then?

delivery_auto_refresh/tests/test_delivery_auto_refresh.py
```diff
@@ -310,8 +294,8 @@ def test_auto_refresh_so_and_unlink_line(self):
         self.assertFalse(delivery_line.exists())
 
     def test_auto_add_delivery_line_add_service(self):
-        """Delivery line should not be created because
-        there are only service products in SO"""
+        """No delivery line when service only"""
+        self.env["ir.config_parameter"].sudo().set_param(self.auto_add_delivery_line, 1)
```
simahawk: don't we use the flag on the company?

delivery_auto_refresh/models/res_config_settings.py
```diff
@@ -7,16 +7,19 @@
...
+    carrier_auto_assign_on_create = fields.Boolean(
+        related="company_id.carrier_auto_assign_on_create",
+        readonly=False,
     )
+    # End migration note
+
     auto_add_delivery_line = fields.Boolean(
-        "Add Delivery Line Automatically",
+        "Refresh shipping cost line automatically",
         config_parameter="delivery_auto_refresh.auto_add_delivery_line",
```
simahawk: why not moving these to company flags too?

delivery_auto_refresh/models/sale_order.py
```diff
@@ -56,63 +85,109 @@ def _get_param_auto_add_delivery_line(self):
...
+            self._remove_delivery_line()
+        elif self.is_all_service:
+            self._remove_delivery_line()
+        else:
+            price_unit = self.carrier_id.rate_shipment(self)["price"]
+            delivery_lines = self.order_line.filtered("is_delivery")
+            if not delivery_lines:
+                self._create_delivery_line(self.carrier_id, price_unit)
+            elif len(delivery_lines) > 1:
+                delivery_discount = self._get_delivery_discount()
```
simahawk: you have already filtered the lines above. I would make this method accept the lines instead. Or trash the method unless you need it as a hook.

delivery_auto_refresh/models/sale_order.py
```diff
@@ -56,63 +85,109 @@ def _get_param_auto_add_delivery_line(self):
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        # Prevent to refresh delivery in the call to super
+        orders = (
+            super(SaleOrder, self.with_context(auto_refresh_delivery=True))
+            .create(vals_list)
+            .with_context(auto_refresh_delivery=False)
+        )
+        for order in orders:
```
simahawk, line 154:
```suggestion
        )
        for order in orders.with_context(auto_refresh_delivery=False):
```
I'd find this more clear. Matter of taste likely....

delivery_auto_refresh/__manifest__.py
jbaudoux: for sure :)
It's done

delivery_auto_refresh/migrations/16.0.2.0.0/post-migration.py
jbaudoux: is it really better to delete them afterwards? I don't know...

delivery_auto_refresh/tests/test_delivery_auto_refresh.py
jbaudoux: Was on it's way. It's now done

delivery_auto_refresh/models/res_config_settings.py
jbaudoux: Was on it's way. It's now done

delivery_auto_refresh/models/sale_order.py
jbaudoux: I prefer my way as it is linked to the call to create where I set context before and remove after...

delivery_auto_refresh/models/sale_order.py
jbaudoux: Indeed. I removed the double filtering

jbaudoux:
Config parameters moved to company. `sale_` prefix added to the parameters

jbaudoux:
Ready for final review

delivery_auto_refresh/models/sale_order.py
```diff
@@ -8,115 +9,185 @@
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    # Migration note: This field is not used anymore and can be dropped in later versions
     available_carrier_ids = fields.Many2many(
         comodel_name="delivery.carrier",
         compute="_compute_available_carrier_ids",
     )
 
-    @api.onchange("partner_id", "partner_shipping_id")
```
pedrobaeza, line 16: Why changing this logic if it was correct?

delivery_auto_refresh/models/sale_order.py
```diff
@@ -8,115 +9,185 @@
...
-        """Create or refresh delivery line on create."""
-        order = super().create(vals)
-        order._auto_refresh_delivery()
-        return order
+        if self.env.context.get("auto_refresh_delivery"):
+            return
+        if not self._is_auto_add_delivery_line():
+            return
+        if self.state not in ("draft", "sent"):
+            return
```
pedrobaeza: Why not compacting them for less lines and better readability?

```suggestion
        if (
            self.env.context.get("auto_refresh_delivery")
            or not self._is_auto_add_delivery_line()
            or self.state not in ("draft", "sent")
        ):
            return
```

delivery_auto_refresh/models/sale_order.py
```diff
@@ -8,115 +9,185 @@
...
+
+        self = self.with_context(auto_refresh_delivery=True)
+
+        if not self.carrier_id:
+            self._set_delivery_carrier()
+
+        if not self.carrier_id:
+            self._remove_delivery_line()
+        elif self.is_all_service:
+            self._remove_delivery_line()
```
pedrobaeza:
```suggestion
        if not self.carrier_id or self.is_all_service:
            self._remove_delivery_line()
```

delivery_auto_refresh/models/sale_order.py
```diff
@@ -8,115 +9,185 @@
...
+        # Prevent to refresh delivery in the call to super
+        orders = (
+            super(SaleOrder, self.with_context(auto_refresh_delivery=True))
+            .create(vals_list)
+            .with_context(auto_refresh_delivery=False)
+        )
+        for order in orders:
+            # Migration Note 17.0: move this to module sale_order_carrier_auto_assign
+            if not order.carrier_id and order._is_auto_set_carrier_on_create():
+                order._set_delivery_carrier()
```
pedrobaeza, line 157: Why doing this instead of converting the field into a computed writable field that makes the code more declarative?

delivery_auto_refresh/models/sale_order_line.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2024 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def _get_protected_fields(self):
```
pedrobaeza: As this is a child model (coming from a o2m), we prefer to not separate it in a different file. Please respect this preference and reduce the diff.

delivery_auto_refresh/models/sale_order_line.py
```diff
@@ -0,0 +1,32 @@
...
+        # Avoid locked orders validation error when voiding the delivery line
+        fields = super()._get_protected_fields()
+        if self.env.context.get("delivery_auto_refresh_override_locked"):
+            return [x for x in fields if x not in ["product_uom_qty", "price_unit"]]
+        return fields
+
+    @api.model_create_multi
+    def create(self, vals_list):
```
pedrobaeza: Why all the create/write overrides? It seems a great overhead.

delivery_auto_refresh/models/sale_order.py
jbaudoux: Because it's not only setting the carrier but also creating the SO delivery line

delivery_auto_refresh/models/sale_order_line.py
jbaudoux: For implementing the missing feature from the roadmap :)
`* This is only working from user interface, as there's no way of making compatible the auto-refresh intercepting create/write methods from sale order lines.`

delivery_auto_refresh/models/sale_order_line.py
jbaudoux: I prefer one file per model, but as you are the author of this one, I'll respect this and do the change

delivery_auto_refresh/models/sale_order.py
jbaudoux: * I made it generic for onchange & create
* I call the wizard instead of re-implementing a feature of the wizard in this module (`_compute_available_carrier_ids`) to keep compatibility with any module changing the wizard behavior.
* I changed the logic to keep the carrier that is already set instead of overriding with wizard default value.
* I also update the delivery line if the partner is changed while there are already SO lines

delivery_auto_refresh/migrations/16.0.2.0.0/post-migration.py
simahawk: if you don't use them anymore and if you are taking care of their migration I think it makes sense...

cyrilmanuel:
hi @pedrobaeza  could you review this PR please ?

delivery_auto_refresh/models/sale_order.py
```diff
@@ -130,7 +196,7 @@ def _is_delivery_line_voidable(self):
         # nothing to be done either. If there are more than one delivery lines
         # we won't be doing anything as well.
         if (
-            self.state not in {"done", "sale"}
+            self.state not in ("done", "sale")
```
pedrobaeza, line 199: The set was correct for faster operation. Why changing it?

delivery_auto_refresh/models/sale_order.py
```diff
@@ -145,9 +211,25 @@ class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
     def _get_protected_fields(self):
-        """Avoid locked orders validation error when voiding the
-        delivery line"""
+        # Avoid locked orders validation error when voiding the delivery line
```
pedrobaeza, line 214: Why changing legit docstring to comments? The docstring is specifying what the method override is doing, so please don't change it.

delivery_auto_refresh/models/res_config_settings.py
```diff
@@ -7,19 +7,22 @@
 class ResConfigSettings(models.TransientModel):
     _inherit = "res.config.settings"
 
-    set_default_carrier = fields.Boolean(
```
pedrobaeza, line 10: No need to rename these fields for making the diff bigger. I understand that is for avoiding name clashing, and I agree the initial name was not very correct, but not sure the convenience of increasing the diff. You tell me if it's only this one.

pedrobaeza (commented):
Thanks for your patience, Jacques. Apart from the comments inline (which there can be several occurrences, but I have written only once):

As this code was originally mine, I would appreciate that you follow my coding style not adding empty lines inside a method (it's also something recommended - but not enforced - in Python style guides).

There are docstring changed to comments.

delivery_auto_refresh/models/sale_order.py
simahawk: I may answer: the usage of `set` or `list` is kind of wrong in all the cases where you have a fixed set of values and you don't need to purge duplicates or manipulate values afterwards.
Plus, have a look at the memory size:
```
>>> sys.getsizeof(("done", "sale"))
56
>>> sys.getsizeof({"done", "sale"})
216
```
In fact, sounds like something that the linter could prevent :)

Yet, such changes might be applied afterwards to reduce diff size in the future. You have a point here :wink:

pedrobaeza: @yajo do you have at hand the times tests where sets beats inclusion comparison against lists? This is something of performance, not memory size.

jbaudoux: @pedrobaeza is right that sets are better optimized for inclusion comparison. But this is only relevant if the set is big enough.

```
>>> def in_test(iterable):
...     for i in range(10):
...         if i in iterable:
...             pass
... 
>>> from timeit import timeit
>>> timeit(
...     "in_test(iterable)",
...     setup="from __main__ import in_test; iterable = set(range(5))",
...     number=10000)
0.030677038012072444
>>> timeit(
...     "in_test(iterable)",
...     setup="from __main__ import in_test; iterable = list(range(5))",
...     number=10000)
0.014917187043465674
>>> timeit(
...     "in_test(iterable)",
...     setup="from __main__ import in_test; iterable = tuple(range(5))",
...     number=10000)
0.009921401971951127
```

delivery_auto_refresh/models/sale_order.py
jbaudoux: The docstring is provided by the base method. You were not extending the docstring but replacing it with different information that looks for me a comment of this specific extension.

With your way of doing:
```
>>> help(env["sale.order.line"]._get_protected_fields)
Help on method _get_protected_fields in module odoo.addons.delivery_auto_refresh.models.sale_order:

_get_protected_fields() method of odoo.api.sale.order.line instance
    Avoid locked orders validation error when voiding the delivery line
```

After the fix recovering proper docstring:
```
>>> help(env["sale.order.line"]._get_protected_fields)
Help on method _get_protected_fields in module odoo.addons.delivery_auto_refresh.models.sale_order:

_get_protected_fields() method of odoo.api.sale.order.line instance
    Give the fields that should not be modified on a locked SO.
    
    :returns: list of field names
    :rtype: list
```

I hope this clarifies it

pedrobaeza: Wow, that changes my way of thinking in all of this, but it's also a pity as I feel that a comment doesn't cover the feature I try to put: to explain why the whole method is being overridden. Do you know any other way to put this?

delivery_auto_refresh/models/res_config_settings.py
jbaudoux: I need this module to be multi company aware. So I had to move all the config from global config parameters to company object. As I was doing this move, it was the occasion to improve the config names as I'm creating new fields on company.

For `res.config.settings`, it's a transient model... Better align the names with the one on company. The diff is only 6 lines in the tests and in it's related commit. Worth it in my opinion.

delivery_auto_refresh/models/sale_order.py
jbaudoux: We probably wait that odoo drops `_inherit` for proper python inheritance before trying to do better.

delivery_auto_refresh/models/res_config_settings.py
pedrobaeza: OK, fair enough.

pedrobaeza (approved):
Thanks again for your efforts.

/ocabot merge nobump

jbaudoux:
> Thanks again for your efforts.

@pedrobaeza Thanks for your time on this review :)

rafaelbn:
This PR affects MT-5997 @moduon and https://github.com/OCA/delivery-carrier/pull/815 which is being reviewed by @yajo 

@Gelojr @fcvalgar please deep functional review 😄 ❤️

rafaelbn:
Hello @jbaudoux ,

We don't want to use `sale_order_carrier_auto_assign` module. It's mandatory?

We wouldn't be forced yo use it 😢 

Let me know please! 😄 

Thank you!

jbaudoux:
> Hello @jbaudoux ,
> 
> We don't want to use `sale_order_carrier_auto_assign` module. It's mandatory?
> 
> We wouldn't be forced yo use it 😢. 
> 
> Let me know please! 😄
> 
> Thank you!

There is no new dependency in v16.

rafaelbn:
OK, sorry. I miss some comments 🙏🏼 . Thank you! 😄

yajo:
It seems v16 settings layout broke:

image

delivery_auto_refresh/models/sale_order.py
yajo: Hi there! Your test is doomed:

```py
>>> from timeit import timeit
>>> def in_test(iterable):
...     return 1000 in iterable
>>> timeit("in_test(iterable)", setup="from __main__ import in_test; iterable = set(range(5))")
0.06247029999940423
>>> timeit("in_test(iterable)", setup="from __main__ import in_test; iterable = list(range(5))")
0.11310941600095248
>>> timeit("in_test(iterable)", setup="from __main__ import in_test; iterable = tuple(range(5))")
0.1125952110014623
```

Using `set` for an `in` construct is always faster. In a `list`/`tuple` of 5 elements, it's still 50% faster.

For this use case, the tuple contains 2 elements. Let's see:

```py
>>> timeit("in_test(iterable)", setup="from __main__ import in_test; iterable = set(range(2))")
0.07049787299911259
>>> timeit("in_test(iterable)", setup="from __main__ import in_test; iterable = list(range(2))")
0.07827233200077899
>>> timeit("in_test(iterable)", setup="from __main__ import in_test; iterable = tuple(range(2))")
0.07647827899927506
```

Using `set` is still faster, although barely noticeable. This means that, if you see a pattern such as `a in (0, 1)`, it's probably not worth it to optimize it. But if you see one like `a in {0, 1}`, it's definitely not worth it un-optimizing it to use a tuple.

It's not a matter of % of improvement, but of algorithmic complexity. Using a set will be of O*1 complexity (constant), while using a list or tuple will be O*n complexity (linear). Thus, using a set is the default rule of thumb here.

It's not worth reverting this change for such a small improvement now that it's merged, but at least I hope you learned something today 😄

rousseldenis:
> It seems v16 settings layout broke:

@jbaudoux

toita86:
@rousseldenis I'm not sure if it can help, but this is my attempt to fix the layout e6214da220132e6289d01741fd08d00229ee2efb

PR #800, opened 2024-04-19
[16.0][MIG] delivery_translatable

David-Luis-Mora:


rousseldenis:
@David-Luis-Mora In fact, this module is no more necessary : https://github.com/odoo/odoo/blob/16.0/addons/delivery/models/delivery_carrier.py#L40

PR #803, opened 2024-04-29
[16.0][FIX] delivery_deliverea: Generate a deliverea return when sending to shipper

bencoronel:
Changed send_to_shipper method to generate the return.

The Print Label button was generating the return instead of the Generate Return button and the Label wasn't being saved.

Now the Generate Return actually generates the return.


OscarIndias (approved):
LGTM!

PR #734, opened 2023-12-11
[14.0][FIX] delivery_cttexpress: Fix failing test

jbaudoux:
Test is failing @chienandalu 

https://github.com/OCA/delivery-carrier/actions/runs/7168825842/job/19517839511?pr=720

PR #805, opened 2024-05-02
[14.0][FIX] delivery_schenker: Use display_name in name1 field of address

RuudFreshfilter:
Not always is the `partner.name` filled. If this is the case then the Schenker API returns an error.

### DIFF
--- a/delivery_schenker/models/delivery_carrier.py
+++ b/delivery_schenker/models/delivery_carrier.py
@@ -266,7 +266,7 @@ class DeliveryCarrier(models.Model):
         """
         vals = {
             "type": address_type,
-            "name1": partner.display_name,
+            "name1": partner.name or partner.commercial_partner_id.name,
             "locationType": location_type,  # POSTAL or PHYSICAL
             "personType": person_type,  # PERSON OR COMPANY
             "street": partner.street,
--- a/delivery_schenker/models/delivery_carrier.py
+++ b/delivery_schenker/models/delivery_carrier.py
@@ -266,7 +266,7 @@ class DeliveryCarrier(models.Model):
         """
         vals = {
             "type": address_type,
-            "name1": partner.name,
+            "name1": partner.display_name,
             "locationType": location_type,  # POSTAL or PHYSICAL
             "personType": person_type,  # PERSON OR COMPANY
             "street": partner.street,


thomaspaulb (approved):

thomaspaulb:
Failing test should [repair itself](https://github.com/OCA/delivery-carrier/pull/734)

thomaspaulb:
@OCA/logistics-maintainers Could this be re-opened?

jbaudoux (changes requested):
Can you better explain your use case ?
Shouldn't you fail if the partner has no name ?
I don't like the idea of taking the `display_name` as this could contain internal information to ease the selection of a right partner (like code, country...) that should not be sent to schenker

cc @mt-software-de

thomaspaulb:
@RuudFreshfilter Do you remember for which partner this happened? I think it was for Delivery contacts that have no name by itself but just rely on the name of their parent

delivery_schenker/models/delivery_carrier.py
```diff
@@ -266,7 +266,7 @@ def _prepare_schenker_address(
         """
         vals = {
             "type": address_type,
-            "name1": partner.name,
+            "name1": partner.display_name,
```
NL66278: I understand this can happen when the address belongs to another partner (likely the company) and itself has no name. Then maybe better user `partner.name or partner.parent_id.name`. This will prevent the possibility of sending internal info, and when there is no parent, or the parent has no name, using display_name will not help anyway.

RuudFreshfilter:
> @RuudFreshfilter Do you remember for which partner this happened? I think it was for Delivery contacts that have no name by itself but just rely on the name of their parent

This was the issue. it was for delivery and invoice addresses.

delivery_schenker/models/delivery_carrier.py
thomaspaulb: @RuudFreshfilter How about using `partner.name or partner.commercial_partner_id.name` then?

RuudFreshfilter: Yeah that sounds like a good solution. I applied that.

thomaspaulb:
@jbaudoux @NL66278 OK?

delivery_schenker/models/delivery_carrier.py
mt-software-de: Thanks for your contribution. 
But the commercial partner name should not be used, because this belongs maybe not to the actual address. 
You should better use `partner.parent_name`. Like it is done in the contact widget of odoo. https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/odoo/addons/base/views/ir_qweb_widget_templates.xml#L8

NL66278 (approved):
:+1: LGTM

delivery_schenker/models/delivery_carrier.py
```diff
@@ -266,7 +266,7 @@ def _prepare_schenker_address(
         """
         vals = {
             "type": address_type,
-            "name1": partner.name,
+            "name1": partner.name or partner.commercial_partner_id.name,
```
mt-software-de, line 269:
```suggestion
            "name1": partner.name or partner.parent_name
```

Thanks for your contribution.
But the commercial partner name should not be used, because this belongs maybe not to the actual address.
You should better use partner.parent_name. Like it is done in the contact widget of odoo. https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/odoo/addons/base/views/ir_qweb_widget_templates.xml#L8

mt-software-de (changes requested):

delivery_schenker/models/delivery_carrier.py
mt-software-de: cc @jbaudoux

PR #824, opened 2024-05-28
[16.0][MIG] delivery_carrier_pricelist: migration to version 16.0

AnizR:
Migration of module `delivery_carrier_pricelist` to 16.0

Few changes regarding implementation:
 - computation of price using pricelist has changed since version 14.0
 - retrieval of view has also changed since version 14.0
 
 No changes in the behavior of the module.


gurneyalex (approved):

delivery_carrier_pricelist/models/delivery_carrier.py
```diff
@@ -36,12 +35,14 @@ class DeliveryCarrier(models.Model):
...
             # Force computation from pricelist based on invoicing policy
-            # using a temporary record.
             # Required since core's `rate_shipment` relies on `self.delivery_type`
             # to lookup for the right handler.
-            tmp_rec = self.new(self.copy_data()[0])
-            tmp_rec.delivery_type = "pricelist"
-            result = super(DeliveryCarrier, tmp_rec).rate_shipment(order)
+            # using a 'temp record' with new() won't work since it has NewId
+            # and rate_shipment() compares those
+            tmp_type = self.delivery_type
```
rousseldenis, line 42: Mmmmh, I would say this should be protected.

@AnizR 

```python
    with self.env.norecompute():
        (....)
```

AnizR: You wan't to avoid computation of fileds that's right?

I think that `norecompute` is depreceated.
It should be like this, isn't it? 
```python
with self.env.protecting(<all_fields of self>):
    ...
```

PR #826, opened 2024-06-04
[17.0][MIG] delivery_roulier: Migration to 17.0

yankinmax:
Depends on:
- [x] https://github.com/OCA/delivery-carrier/pull/828
- [x] https://github.com/OCA/delivery-carrier/pull/827


delivery_roulier/tests/test_delivery_roulier.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+        with patch("roulier.roulier.get") as mock_roulier:
+            mock_roulier.return_value = roulier_ret
+            self.picking.send_to_shipper()
+            roulier_args = mock_roulier.mock_calls[0][1]
+            self.assertEqual("get_label", roulier_args[1])
+            roulier_payload = roulier_args[2]
+            self.assertEqual(len(roulier_payload["parcels"]), 1)
+            # TODO: understand where ``weight`` is changed
+            self.assertEqual(roulier_payload["parcels"][0].get("weight"), 1.2)
```
yankinmax: I would appreciate if anyone can help with this

delivery_roulier/tests/test_delivery_roulier.py
```diff
@@ -111,6 +108,7 @@ def test_roulier(self):
             self.assertEqual("get_label", roulier_args[1])
             roulier_payload = roulier_args[2]
             self.assertEqual(len(roulier_payload["parcels"]), 1)
+            # TODO: understand where ``weight`` is changed
```
vvrossem: I'm afraid I cannot help you on this (I'm not familiar with this repo), but maybe @florian-dacosta could :crossed_fingers: ?

vvrossem: Can this TODO be removed now?

vvrossem (approved):

ivantodorovich (approved):

delivery_roulier/models/stock_move_line.py
```diff
@@ -39,3 +39,18 @@ def get_unit_price_for_customs(self):
...
+        if "result_package_id" in vals:
+            for move_line in self.filtered(lambda s: not s.result_package_id):
+                quants_by_product = self.env[
+                    "stock.quant"
+                ]._get_quants_by_products_locations(
+                    move_line.product_id, move_line.location_id
+                )
+                if not quants_by_product or len(quants_by_product) > 1:
+                    continue
+                vals["quant_id"] = quants_by_product.get(move_line.product_id.id).id
```
Camille0907: I don't understand here : we are in a for loop, and assigning a value to `vals["quant_id"]`
But, if during the loop, you come a second time on this line, you will overwrite the value previously set, no ?

Camille0907 (approved):
Migration technically LGTM

delivery_roulier/models/stock_move_line.py
```diff
@@ -39,3 +39,18 @@ def get_unit_price_for_customs(self):
     def get_sale_order_line(self):
         self.ensure_one()
         return self.move_id.sale_line_id
+
+    def write(self, vals):
```
rousseldenis: My big question: Why is this specific to roulier ??

yankinmax: It's not specific to roulier.
Somehow after `stock.picking`::`_put_in_pack` call is finished `quant_id` is not assigned to a `stock.quant.package`.
But, from what I understand during tests debug `quant_id` is expected to be assigned.
So, I suppose there is a refactor of delivery between v16 and v17, which should be handled.
Any better ideas?
Thanks for review

delivery_roulier/models/stock_move_line.py
yankinmax: yep, you're right, it's wrong, my bad

delivery_roulier/models/stock_picking.py
```diff
@@ -290,3 +290,17 @@ def open_website_url(self):
         action["domain"] = [("id", "in", packages.ids)]
         action["context"] = {"picking_id": self.id}
         return action
+
+    def _put_in_pack(self, move_line_ids):
```
yankinmax: I've splitted correctly the migration commit to `pre-commit` one.
Here is the latest fixup to update `quant_ids` on the package with all quants from move lines in this package.
@rousseldenis @Camille0907 
thanks for the reviews

yankinmax:
@rousseldenis is this migration ok now?

delivery_roulier/models/stock_picking.py
florian-dacosta: Hello,
I don't get why we need this, not why it is in this module. Is this necessary for this module to work ? Why ?

yankinmax: Hello @florian-dacosta 
From my POV, when the package is created and quants are put in pack, it should be linked to a package.
Probably it was a case in previous versions, but not in v17.
As a result here:
```
def _roulier_get_parcel(self, picking):
        self.ensure_one()
        weight = self.shipping_weight or self.weight
        parcel = {"weight": weight, "reference": self.name}
        return parcel
```
`shipping_weight` or `weight` are missed and roulier response is wrong.
If you can check the code and suggest better solution it would be great.
BTW, I'm not an expert using this module. But, I need it to be migrated for my project.
Thanks in advance.

florian-dacosta: Well, I believe it has always been this way. When you click on "put in pack" button, the pack is set on the operations (stock.move.line) but if you check the pack, it is still empty (no quant inside). This is not specific to the v17.
I understand it may be an issue, if you want to generate the label, before validating the picking, which is not the way odoo works.

The label generation is triggered at the end of "stock.picking._action_done". At this stage, the quant are already linked to the pack, so the weight is ok.

I guess it would be nice to be able to generate label before the validation of the shipping, but : 
- This is not a problem specific to delivery_roulier, I think it could be an issue for any module managing the carrier label. If I should pick a module with something like that, I'd probably create a new dedicated module, which you could choose to install or not, along with delivery_roulier (or any other module managing shipping labels).
- The way to do it does not seem ok to me. Indeed, the button put_in_pack is optional, you can also set the result_package_id of the stock.move.line manually (in this case the weight won't be ok either)
After you set a result package id, you can remove it manually... You can delete the stock.move.line, you can change the pack, you can do anything. Then, the package on some quants will be wrong, which is an issue.
Also, you have a source package and a result package on stock.move.line. When a quant is already in a pack, when you transfer it, the source package is filled. If you write the pack on the quant with the "put in pack" button, then the source package should probably be set, not sure about the impacts.

Anyway, I have no solution to propose around this subject now, but this one should definitely not go here and should probably be well tested as well, with the different cases.

florian-dacosta: Well, I believe it has always been this way. When you click on "put in pack" button, the pack is set on the operations (stock.move.line) but if you check the pack, it is still empty (no quant inside). This is not specific to the v17.
I understand it may be an issue, if you want to generate the label, before validating the picking, which is not the way odoo works.

The label generation is triggered at the end of "stock.picking._action_done". At this stage, the quant are already linked to the pack, so the weight is ok.

I guess it would be nice to be able to generate label before the validation of the shipping, but : 
- This is not a problem specific to delivery_roulier, I think it could be an issue for any module managing the carrier label. If I should pick a module with something like that, I'd probably create a new dedicated module, which you could choose to install or not, along with delivery_roulier (or any other module managing shipping labels).
- The way to do it does not seem ok to me. Indeed, the button put_in_pack is optional, you can also set the result_package_id of the stock.move.line manually (in this case the weight won't be ok either)
After you set a result package id, you can remove it manually... You can delete the stock.move.line, you can change the pack, you can do anything. Then, the package on some quants will be wrong, which is an issue.
Also, you have a source package and a result package on stock.move.line. When a quant is already in a pack, when you transfer it, the source package is filled. If you write the pack on the quant with the "put in pack" button, then the source package should probably be set, not sure about the impacts.

Anyway, I have no solution to propose around this subject now, but this one should definitely not go here and should probably be well tested as well, with the different cases.

florian-dacosta (changes requested):

delivery_roulier/models/stock_picking.py
florian-dacosta: Well, I believe it has always been this way. When you click on "put in pack" button, the pack is set on the operations (stock.move.line) but if you check the pack, it is still empty (no quant inside). This is not specific to the v17.
I understand it may be an issue, if you want to generate the label, before validating the picking, which is not the way odoo works.

The label generation is triggered at the end of "stock.picking._action_done". At this stage, the quant are already linked to the pack, so the weight is ok.

I guess it would be nice to be able to generate label before the validation of the shipping, but : 
- This is not a problem specific to delivery_roulier, I think it could be an issue for any module managing the carrier label. If I should pick a module with something like that, I'd probably create a new dedicated module, which you could choose to install or not, along with delivery_roulier (or any other module managing shipping labels).
- The way to do it does not seem ok to me. Indeed, the button put_in_pack is optional, you can also set the result_package_id of the stock.move.line manually (in this case the weight won't be ok either)
After you set a result package id, you can remove it manually... You can delete the stock.move.line, you can change the pack, you can do anything. Then, the package on some quants will be wrong, which is an issue.
Also, you have a source package and a result package on stock.move.line. When a quant is already in a pack, when you transfer it, the source package is filled. If you write the pack on the quant with the "put in pack" button, then the source package should probably be set, not sure about the impacts.

Anyway, I have no solution to propose around this subject now, but this one should definitely not go here and should probably be well tested as well, with the different cases.

florian-dacosta (changes requested):

yankinmax:
I'll remove the part I've proposed.
I don't have any other better solution for now.

florian-dacosta:
> I'll remove the part I've proposed. I don't have any other better solution for now.

Thanks, I'll approve then.
Can you also have a look here https://github.com/OCA/delivery-carrier/pull/851 and make a cherry pick to include the v16 fix here please?

yankinmax:
> > I'll remove the part I've proposed. I don't have any other better solution for now.
> 
> Thanks, I'll approve then. Can you also have a look here #851 and make a cherry pick to include the v16 fix here please?

of course, thanks

yankinmax:
@florian-dacosta I've checked your fix in v17. Still the same issue:
```
_____________________________________________________________________________________ DeliveryRoulierCase.test_roulier _____________________________________________________________________________________

self = <odoo.addons.delivery_roulier.tests.test_delivery_roulier.DeliveryRoulierCase testMethod=test_roulier>

    def test_roulier(self):
        roulier.get_carriers_action_available = MagicMock(
            return_value={"test": ["get_label"]}
        )
        with patch("roulier.roulier.get") as mock_roulier:
            mock_roulier.return_value = roulier_ret
            self.picking.send_to_shipper()
            roulier_args = mock_roulier.mock_calls[0][1]
            self.assertEqual("get_label", roulier_args[1])
            roulier_payload = roulier_args[2]
            self.assertEqual(len(roulier_payload["parcels"]), 1)
>           self.assertEqual(roulier_payload["parcels"][0].get("weight"), 1.2)
E           AssertionError: 0.0 != 1.2
```
so weight is not properly updated in `_roulier_get_parcel`

florian-dacosta:
> @florian-dacosta I've checked your fix in v17. Still the same issue:
> 
> ```
> _____________________________________________________________________________________ DeliveryRoulierCase.test_roulier _____________________________________________________________________________________
> 
> self = <odoo.addons.delivery_roulier.tests.test_delivery_roulier.DeliveryRoulierCase testMethod=test_roulier>
> 
>     def test_roulier(self):
>         roulier.get_carriers_action_available = MagicMock(
>             return_value={"test": ["get_label"]}
>         )
>         with patch("roulier.roulier.get") as mock_roulier:
>             mock_roulier.return_value = roulier_ret
>             self.picking.send_to_shipper()
>             roulier_args = mock_roulier.mock_calls[0][1]
>             self.assertEqual("get_label", roulier_args[1])
>             roulier_payload = roulier_args[2]
>             self.assertEqual(len(roulier_payload["parcels"]), 1)
> >           self.assertEqual(roulier_payload["parcels"][0].get("weight"), 1.2)
> E           AssertionError: 0.0 != 1.2
> ```
> 
> so weight is not properly updated in `_roulier_get_parcel`

The fix was not related to the weight issue, it was about picking tracking link.
Where do you see this error ? The test on this PR seem fine, am I missing something ?

yankinmax:
ah sorry I forgot to update, wait a sec

yankinmax:
@florian-dacosta 
For this error my fix was applied with `put_in_pack` override:
https://github.com/OCA/delivery-carrier/actions/runs/9856183670/job/27212631968?pr=826#step:8:74

florian-dacosta:
Oh, I understand. After checking the test both on v16 and v17, here is the change that did break the delivery_roulier test : 
https://github.com/OCA/delivery-carrier/pull/828/commits/99e3d580ad8a7bd66b554cafa09aaf5adb318561#diff-8e3509b96799ba2c6bfb09b79df02a224f2ed73f5b529a0e2fa210f105fd64d2R25
Condition was changed from `if not pack.quant_ids` to `if pack.quant_ids`
I don't understand why this was changed, but this is the reason of the test failing now.

yankinmax:
> Oh, I understand. After checking the test both on v16 and v17, here is the change that did break the delivery_roulier test : [99e3d58#diff-8e3509b96799ba2c6bfb09b79df02a224f2ed73f5b529a0e2fa210f105fd64d2R25](https://github.com/OCA/delivery-carrier/commit/99e3d580ad8a7bd66b554cafa09aaf5adb318561#diff-8e3509b96799ba2c6bfb09b79df02a224f2ed73f5b529a0e2fa210f105fd64d2R25) Condition was changed from `if not pack.quant_ids` to `if pack.quant_ids` I don't understand why this was changed, but this is the reason of the test failing now.

I'll prepare a fix

yankinmax:
@florian-dacosta is this what we need:
- [ ] https://github.com/OCA/delivery-carrier/pull/856
I feel like I'm missing something

florian-dacosta:
> @florian-dacosta is this what we need:
> 
>     * [ ]  [[FIX] base_delivery_carrier_label: compute the weight by the product weight #856](https://github.com/OCA/delivery-carrier/pull/856)
>       I feel like I'm missing something

I think it is.
In odoo the native `_compute_weight` method compute weight of a pack from the `stock.move.line` if the context "picking_id" is set (I have no idea when/where it is set) else it is based on the linked quants.
The override in base_delivery_carrier_label tries to compute the weight in case we do not have quant and we do not have the context either (so native method does not work). 
That is not very great... But that is what we have for now, it does seem to do the job!

yankinmax:
@florian-dacosta your test written in `base_delivery_carrier_label` is correct, no doubts. But, then we should find an issue and fix that module.

yankinmax:
@florian-dacosta I'm feeling like we can finish the job finally)
Can you approve?

delivery_roulier/__manifest__.py
```diff
@@ -20,6 +20,7 @@
     "external_dependencies": {
         "python": [
             "roulier",  # '>0.2.0'
+            "odoo_test_helper",
```
florian-dacosta: I wonder if it is necessary ? Should it not be in a test-requirements file or something ?

rousseldenis: @yankinmax @florian-dacosta It's subjective. Either, you add it manually to test-requirements.txt, either you put here (and it will be added automatically to requirements.txt).

There is no guideline on that point particularly

yankinmax: I would leave it here only for one important reason.
When we aggregate PR's as pending merges on projects we are merging them one by one.
And it can happen there is a `test-requirements.txt` file on a branch.
Then we have a conflict and automated process fails.
You need to complete the process manually and sometimes it leads to mistakes.
So, if its possible to have it directly on `requirements.txt` it would be great
@rousseldenis @florian-dacosta

yankinmax:
BTW, if we could merge this one today we can also merge one more so it can be a very fast line day :smile: 
- [ ] https://github.com/OCA/delivery-carrier/pull/829

delivery_roulier/__manifest__.py
florian-dacosta: I though there were some guidlines about this, because one could use the module in production without having the test libs installed (which would lead to some warning beeing in the manifest file no?)
Anyway, I don't really care on my side, so ok for me!

florian-dacosta (approved):

florian-dacosta:
@rousseldenis I guess your comment is now obsolete, but if you could confirm it changing your review it would be great!

yankinmax:
@rousseldenis can you please update your review and in case you approve trigger the merge?

delivery_roulier/__manifest__.py
simahawk: No, please, do not add a test dependency here. We must propose a specific solution w/ a specific key.

delivery_roulier/models/delivery_carrier.py
```diff
@@ -0,0 +1,60 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+
+from odoo import models
+
+_logger = logging.getLogger(__name__)
+try:
```
simahawk: I don't think we need this kind of blocks anymore, am  I wrong?

delivery_roulier/tests/test_delivery_roulier.py
```diff
@@ -0,0 +1,117 @@
...
+
+        # The fake class is imported here !! After the backup_registry
+        from .models import FakeDeliveryCarrier
+
+        cls.loader.update_registry((FakeDeliveryCarrier,))
+        cls.real_get_carriers_action_available = roulier.get_carriers_action_available
+
+    def setUp(self):
+        super().setUp()
+        delivery_product = self.env["product.product"].create(
```
simahawk: Why re-creating this static data for every test? Also can you disable tracking or use BaseCommon as base class?

delivery_roulier/__manifest__.py
yankinmax: I don't understand what you mean @simahawk 
It's not a part of migration.
If we should move this dependency to `test-requirements.txt` it's feasible quickly.

requirements.txt
```diff
@@ -0,0 +1,3 @@
+# generated from manifests external_dependencies
+odoo_test_helper
```
simahawk: to be removed, it must go to test-requirements.txt

simahawk (changes requested):

delivery_roulier/__manifest__.py
simahawk: It's not a dependency. It's a test dependency. Just remove it from here and add it manually to test-req. That's it :wink:

yankinmax:
@simahawk updated with fixup

simahawk (approved):
@yankinmax cool, tnx. Don't forget to squash :wink:

PR #831, opened 2024-06-06
[14.0][ADD] delivery_price_collection_cost: new module

HekkiMelody:
New module to track collection costs separately from the rest of the shipping cost, with a different product and price.

PR also contains a small bridge module between itself and `delivery_price_product_domain`, as well as a small change to `delivery_price_rule_untaxed` necessary to make the bridge module work.

Todo:

- [x] Add tests to the bridge module


rousseldenis:
@aleuffre Isn't that covered by landed_costs ? https://www.odoo.com/documentation/17.0/fr/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_valuation/integrating_landed_costs.html

francesco-ooops:
@rousseldenis looks complicated. Does that allow to set landed costs per sale order, and relate them to a shipping method?

Use case here is "cash on delivery", where carrier collects SO amount, and takes a fixed fee (fee varies according to shipping rules)

Eg: if order value < 300, delivery costs = 8 (collection fee) + 3% order value
if order value >= 300, delivery costs = 16 (collection fee) + 3% order value

francesco-ooops:
Actually to expand the use case: "free delivery" promotion can be applied to a specific SO, but collection costs should be excluded from the promotion.

francesco-ooops (approved):
Functional ok!

ilyasProgrammer (approved):
LGTM

francesco-ooops:
@pedrobaeza all ready for nobump :)

PR #832, opened 2024-06-10
[14.0][FIX] delivery_ups_oca: update delivery state

manuelregidor:
Before this fix, the delivery state was not properly updated, as the first UPS state was taken instead of the last one.

### DIFF
--- a/delivery_ups_oca/models/ups_request.py
+++ b/delivery_ups_oca/models/ups_request.py
@@ -388,8 +388,10 @@ class UpsRequest(object):
                             activity.get("status").get("description"),
                         )
                     )
+                if shipment["package"][0]["activity"]:
                     delivery_state = static_states.get(
-                        activity["status"]["type"], "incidence"
+                        shipment["package"][0]["activity"][0]["status"]["type"],
+                        "incidence",
                     )
             else:
                 for warning in shipment.get("warnings"):


manuelregidor:
@ValentinVinagre

ValentinVinagre (approved):
LGTM 👍🏻

manuelregidor:
@LoisRForgeFlow @rousseldenis @Cedric-Pigeon Could you have a look, please? Thank you.

HaraldPanten (approved):

HaraldPanten:
@pedrobaeza @etobella This is a critical FIX for this module. Could you give us a hand in order to move it forward?

THX!

pedrobaeza:
AFAIK, the idea is to have all the activities added in one shot. This way, you are only getting one update, not the in between. If the activity order is reversed, then do the `for` reversed, but don't remove such loop.

manuelregidor:
@pedrobaeza  What this FIX does is just getting the most recent state update obtained from the carrier, which is in the first position of the list of the states updates list (shipment["package"][0]["activity"][0]). In the previous version of the module, the element of the states updates list that was taken was the last element, which is not correct, as the last element of the list is the oldest state. This is the value that we need to update the delivery_state field.

However, this method also updates the field tracking_state_history with the full list of states updates, placing the most recent states on top, and this part has remained unchaged.

pedrobaeza (commented):
For reducing the diff, please keep the for as it was, and after it, then assign `delivery_state`.

manuelregidor:
@pedrobaeza Done

pedrobaeza (approved):
/ocabot merge patch

PR #838, opened 2024-06-21
[17.0][ADD] delivery_sendcloud

ByteMeAsap:
delivery_sendcloud module provides sendcloud shipping integration with Odoo

This module mostly implements what's described in https://docs.sendcloud.sc/api/v2/shipping/

Full documentation for developers is in https://docs.sendcloud.sc/.

This module works for the Community Edition as well as the Enterprise Edition.


ByteMeAsap:
@pedrobaeza , can we please get the pre-commit issue checked? it is giving an error as Parsing error: Unexpected token = for the backend.esm.js file, but it seems to be fine.

Also, have this module reviewed

pedrobaeza:
I can't say too much about JS, as I'm not an expert, but it seems putting a static variable there is not allowed. Anyway, using such variables should be avoided.

About reviewing, I don't know this API nor I'm interested in it for now, so you should look for other reviewers. You can review other PRs and ask in exchange that they review yours.

PR #842, opened 2024-06-26
[14.0][FIX] delivery_purchase : fix module install on demo data

jbaudoux:
While testing this module, I encountered some issues during installation and testing with demo data:
* initialization of PO delivery price is mixing POs in the loop
* if you have installed a carrier delivery_xxx module that does not implement purchase price, then PO confirmation was crashing

cc @pedrobaeza


pedrobaeza (approved):

pedrobaeza:
You can squash both commits, as they belong to the same concept

jbaudoux:
> You can squash both commits, as they belong to the same concept

@pedrobaeza done

hparfr (approved):
/ocabot merge minor

pedrobaeza:
Please fw-port the fix

PR #845, opened 2024-07-02
[14.0][FIX]delivery_state: stock.picking delivery_state field invalid option

manuelregidor:
Before this fix, there was an invalid option for field delivery_type in model stock.picking in use. The used option was base_one_rule, which does not exist. The right option is base_on_rule (without the "e").

T-6288


pedrobaeza (approved):

manuelregidor:
@HaraldPanten @ValentinVinagre @luis-ron

chienandalu (approved):

HaraldPanten (approved):

PR #846, opened 2024-07-02
[ADD] stock_picking_delivery_package_type_domain

rousseldenis:
Needs (based on): 

- https://github.com/OCA/delivery-carrier/pull/847


hparfr (approved):
Code review, lgtm.

rousseldenis:
@jbaudoux @lmignon @sbejaoui

jbaudoux (commented):
Before v16, it was not possible to propagate the carrier on the chain of moves. In v16, as you can propagate the carrier to the pack operation, can't you just propagate it instead of that new module?
I think `stock_picking_delivery_link` is less useful now unless you want to know the carrier when doing the pick operation.

jbaudoux (approved):
Ah, the new one is not depending on the other :)

PR #847, opened 2024-07-03
[FIX] stock_picking_delivery_link: Don't override context value if no ship carrier

rousseldenis:
If shipping steps is set to ship only, the ship_carrier_id on stock picking is not set. So, don't override context value in that case.


rousseldenis:
@glitchov @victoralmau @TheClaud99

pedrobaeza (approved):

victoralmau (approved):

PR #849, opened 2024-07-04
[14.0][FIX] delivery_auto_refresh

toita86:
backport of https://github.com/OCA/delivery-carrier/pull/799


francesco-ooops:
@jbaudoux would you mind reviewing this backporting? thanks!

PicchiSeba (approved):
Code Review:

That's a lot of work. Kudos. The changes to adapt to `14.0` look fine to me.

jbaudoux (approved):
LGTM

jbaudoux:
cc @pedrobaeza

delivery_auto_refresh/migrations/16.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2024 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
```
pedrobaeza: This file is incorrect for a 14.0 version.

pedrobaeza (commented):
Can you please squash a bit the commit history?

rousseldenis (approved):

toita86:
> Can you please squash a bit the commit history?

@pedrobaeza 
Of course, do you have any suggestion? I'm not to sure on how to do it.

francesco-ooops:
Or rather, which commits to squash together

rousseldenis:
I would say commits should remain as it is. They reflect development flow and are the same on the other main branch (even if they should have been a little bit enhanced in their messages).

pedrobaeza:
No, they don't add value when merged. The commit "delivery_auto_refresh: fix create in batch" is not adding value to have it separated. The code itself speaks by itself. And they are not following commit message guidelines:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

francesco-ooops:
@pedrobaeza can you please tell us in detail which ones should be squashed in your opinion, so we can proceed?

rousseldenis:
@jbaudoux

rousseldenis:
@pedrobaeza I would say you had merged it yousrself like this : https://github.com/OCA/delivery-carrier/pull/799#pullrequestreview-2118683894

pedrobaeza:
IMO, you can pack together improvements which code speaks by itself, summarizing in the commit message all of them. At the end, what we want when we do a blame, is to identify the reason of the change. But if there are several commits touching the same lines of code, at the end, the blame operation is a hell.

The same way, a commit that is not telling the reason behind, doesn't have value too much value. Take this commit:

https://github.com/OCA/delivery-carrier/pull/849/commits/4785c77d8d28b207cd74085e03ef2336e610ea32

With the commit message `delivery_auto_refresh: method copied from 16.0`. Which value brings to have it separated if it's not bringing any useful information, and even more, it's not being called in any place in the same commit!!! It may be interesting to keep it as a separated commit if at least where it's used is packed in the same commit and there's an explanation of why this is done.

I know that it seems comfortable to do continuously `git commit` to "save the current work", but at the end, that doesn't help when reviewing code (as our development cycle may not respond the feature cycle), and also for grouping changes in a logical way.

For reference, what I do to save the work is usually `git commit --amend`. If working in several features at the same time, I put several commits, and later changes are committed with `--fixup <sha>` for merging them before pushing with the proper original commits.

pedrobaeza:
> @pedrobaeza I would say you had merged it yousrself like this : [#799 (review)](https://github.com/OCA/delivery-carrier/pull/799#pullrequestreview-2118683894)

Yes, as said in other thread, I don't fight all the possible battles. I have to resign in some of them. In this case, it was my fault to block the PR a lot of time and I already asked for a lot of changes, so I didn't want to be even more picky about that.

In this case, it's also not the most important thing, as this is an old branch, but the number of commits have increased (from 11 to 16), and it can be an opportunity to converge in commit strategies.

francesco-ooops:
@pedrobaeza in the case of commit `delivery_auto_refresh: method copied from 16.0` it was isolated in a specific commit to preserve odoo SA authorship. Should we also keep commit name/description, even if it's not the full content of the commit?

pedrobaeza:
Good that you have arisen this, as I wasn't really aware (didn't check this PR in content yet). Starting with, you can't do that, as this module is AGPL, and the original code is LGPL. More in the formal aspect, you are putting that method, but not indicating anything in the commit message saying that this is copied from Odoo code, and not putting in the same commit the changes on the README `Part of this module is a backport...`. That's what I refer about packing things in a logical way in the commits, and also to document things properly for helping others (or ourselves in the future) to understand why things are there.

About the code itself, I don't think the prepare method is bringing too much value, and even if so, you can just put it without that attribution concern, as at the end, it's not a specific secret technique to do the code, and the lines are obvious even and will be the same if you do it by hand, and for sure you need later adjustments for version 14.

delivery_auto_refresh/models/res_company.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2024 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    # Migration Note 17.0: move this to module sale_order_carrier_auto_assign
```
PicchiSeba: remove

delivery_auto_refresh/models/res_company.py
```diff
@@ -0,0 +1,24 @@
...
+    sale_auto_assign_carrier_on_create = fields.Boolean(
+        "Set default shipping method automatically"
+    )
+    # End migration note
```
PicchiSeba: remove

delivery_auto_refresh/models/res_config_settings.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2023 Tecnativa - Pilar Vargas Moreno
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    # Migration Note 17.0: move this to module sale_order_carrier_auto_assign
```
PicchiSeba: remove

delivery_auto_refresh/models/res_config_settings.py
```diff
@@ -0,0 +1,28 @@
...
+    sale_auto_assign_carrier_on_create = fields.Boolean(
+        related="company_id.sale_auto_assign_carrier_on_create",
+        readonly=False,
+    )
+    # End migration note
```
PicchiSeba: remove

delivery_auto_refresh/models/sale_order.py
```diff
@@ -1,126 +1,238 @@
...
-from odoo import api, fields, models
-
-from ..utils import get_bool_param
+from odoo import _, api, fields, models
 
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    # Migration note: This field is not used anymore and can be dropped in later versions
```
PicchiSeba: remove

delivery_auto_refresh/models/sale_order.py
```diff
@@ -1,126 +1,238 @@
...
-
     @api.depends("partner_shipping_id")
     def _compute_available_carrier_ids(self):
         """We want to apply the same carriers filter in the header as in the wizard"""
         for sale in self:
             wizard = self.env["choose.delivery.carrier"].new({"order_id": sale.id})
             sale.available_carrier_ids = wizard.available_carrier_ids._origin
 
-    def _get_param_auto_add_delivery_line(self):
+    # End migration note
```
PicchiSeba: remove

delivery_auto_refresh/models/sale_order.py
```diff
@@ -1,126 +1,238 @@
...
+
+    # Migration Note 17.0: move this section to module sale_order_carrier_auto_assign
```
PicchiSeba: remove

delivery_auto_refresh/models/sale_order.py
```diff
@@ -1,126 +1,238 @@
...
+        if self.company_id.sale_auto_assign_carrier_on_create:
+            self._set_delivery_carrier(set_delivery_line=False)
+
+    def _is_auto_set_carrier_on_create(self):
+        self.ensure_one()
+        if self.state not in ("draft", "sent"):
+            return False
+        return self.company_id.sale_auto_assign_carrier_on_create
+
+    # End migration note
```
PicchiSeba: remove

delivery_auto_refresh/models/sale_order.py
```diff
@@ -1,126 +1,238 @@
...
+    @api.model_create_multi
+    def create(self, vals_list):
+        # Prevent to refresh delivery in the call to super
+        orders = (
+            super(SaleOrder, self.with_context(auto_refresh_delivery=True))
+            .create(vals_list)
+            .with_context(auto_refresh_delivery=False)
+        )
+        for order in orders:
+            # Migration Note 17.0: move this to module sale_order_carrier_auto_assign
```
PicchiSeba: remove

delivery_auto_refresh/models/sale_order.py
```diff
@@ -1,126 +1,238 @@
...
+            if not order.carrier_id and order._is_auto_set_carrier_on_create():
+                order._set_delivery_carrier()
+            # End migration note
```
PicchiSeba: remove

delivery_auto_refresh/tests/test_sale_order_carrier_auto_assign.py
```diff
@@ -0,0 +1,88 @@
+# Copyright 2024 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+# MIGRATION NOTE for 17.0: Move this to module sale_order_carrier_auto_assign
```
PicchiSeba: remove

delivery_auto_refresh/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,65 @@
...
+        <field name="name">res.config.settings.view.form.inherit.sale</field>
+        <field name="model">res.config.settings</field>
+        <field name="priority" eval="9" />
+        <field name="inherit_id" ref="sale.res_config_settings_view_form" />
+        <!-- Migration note 17.0: depend on sale_order_carrier_auto_assign
+        <field
+            name="inherit_id"
+            ref="sale_order_carrier_auto_assign.res_config_settings_view_form_sale"
+        />
+        -->
```
PicchiSeba: remove

delivery_auto_refresh/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,65 @@
...
+        <field name="arch" type="xml">
+            <xpath expr="//div[@name='shipping_setting_container']" position="inside">
+                <div class="col-12 col-lg-6 o_setting_box" id="sale_auto_refresh">
+                <!-- Migration note 17.0: depend on sale_order_carrier_auto_assign
+                <xpath expr="//div[@id='carrier_auto_assign']" position="inside">
+                -->
+                    <!-- Migration note 17.0: move this to module sale_order_carrier_auto_assign -->
```
PicchiSeba: remove

delivery_auto_refresh/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,65 @@
...
+                    <div class="o_setting_left_pane">
+                        <field name="sale_auto_assign_carrier_on_create" />
+                    </div>
+                    <div class="o_setting_right_pane">
+                        <label for="sale_auto_assign_carrier_on_create" />
+                        <div class="text-muted">
+                            On the sales quotation, add the shipping method on creation.
+                        </div>
+                    </div>
+                    <!-- End Migratio note -->
```
PicchiSeba: remove

delivery_auto_refresh/__manifest__.py
```diff
@@ -13,5 +13,11 @@
     "application": False,
     "installable": True,
     "depends": ["delivery"],
-    "data": ["data/ir_config_parameter.xml", "views/sale_order_views.xml"],
+    # Migration Note 17.0: Add dependency to sale_order_carrier_auto_assign
+    # "depends": ["delivery", "sale_order_carrier_auto_assign"],
```
PicchiSeba: remove

PicchiSeba (changes requested):
Code review: Remove all references of the `17.0` migration. This should not be a concern to us because this is a backport and this matter should be resolved upstream

toita86:
@pedrobaeza I'd appreciate any advice or suggestions on how to handle this more efficiently, or if there's a better approach to resolve this circular dependency.

I'm currently facing a circular dependency issue between two PRs:

https://github.com/OCA/delivery-carrier/pull/849: delivery_auto_refresh
https://github.com/OCA/sale-promotion/pull/224: sale_coupon_delivery_auto_refresh

The issue arises because `sale_coupon_delivery_auto_refresh` has an unreleased dependency on `delivery_auto_refresh`, and this creates a circular reference between the two modules. The fixes introduced in both PRs need each other to avoid breaking the logic.

To resolve this, my idea is to:
- Remove the unreleased dependency from `sale_coupon_delivery_auto_refresh` PR temporarily, allowing us to merge the `delivery_auto_refresh` PR first.

- After merging `delivery_auto_refresh`, we can merge `sale_coupon_delivery_auto_refresh`, which includes the fix for the dependency issue between the two modules.
    
Does this approach make sense, or would you recommend a different solution? I’d appreciate your input!

pedrobaeza:
Yes, remove here the temporary dependency, as I don't see too much sense in adding it while the target module is not being used here.

toita86:
> Yes, remove here the temporary dependency, as I don't see too much sense in adding it while the target module is not being used here.

Thanks for the suggestion! I'll go ahead and remove the temporary dependency.

PR #860, opened 2024-07-15
[14.0][IMP] delivery_purchase: Add Original picking (delivery) (delivery_picking_orig_id) field to purchase order lines

victoralmau:
Add Original picking (delivery) (`delivery_picking_orig_id`) field to purchase order lines

If there are several pickings for an order and several have carriers, several purchase order lines (`is_delivery`) will be created but we need to know the origin (picking) that created them.

Please @pedrobaeza and @carolinafernandez-tecnativa can you review it?

@Tecnativa TT50146

### DIFF
--- a/delivery_purchase/models/purchase_order.py
+++ b/delivery_purchase/models/purchase_order.py
@@ -73,6 +73,9 @@ class PurchaseOrderLine(models.Model):
     _inherit = "purchase.order.line"
 
     is_delivery = fields.Boolean(string="Is a Delivery", default=False)
+    delivery_picking_orig_id = fields.Many2one(
+        comodel_name="stock.picking", string="Origin picking (delivery)"
+    )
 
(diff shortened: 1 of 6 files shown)


delivery_purchase/models/purchase_order.py
```diff
@@ -73,6 +73,9 @@ class PurchaseOrderLine(models.Model):
     _inherit = "purchase.order.line"
 
     is_delivery = fields.Boolean(string="Is a Delivery", default=False)
+    delivery_picking_orig_id = fields.Many2one(
```
carolinafernandez-tecnativa, line 76: Translation is missing for this field.

carolinafernandez-tecnativa (commented):
Rest of review, LGTM :)

pedrobaeza (approved):
/ocabot merge minor

PR #861, opened 2024-07-17
[FIX] base_delibery_carrier_label test helper

gurneyalex:
The migration to 17 of the test helper carrier_label_case was missing the classmethod decorator on the setUpClass method

### DIFF
--- a/base_delivery_carrier_label/tests/carrier_label_case.py
+++ b/base_delivery_carrier_label/tests/carrier_label_case.py
@@ -9,6 +9,7 @@ class CarrierLabelCase(TransactionCase):
     """Base class for carrier label tests. Inherit and override _get_carrier
     to return the carrier you want to test"""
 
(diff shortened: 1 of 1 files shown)


TDu (approved):

cyrilmanuel (approved):

base_delivery_carrier_label/tests/carrier_label_case.py
```diff
@@ -9,6 +9,7 @@ class CarrierLabelCase(TransactionCase):
     """Base class for carrier label tests. Inherit and override _get_carrier
     to return the carrier you want to test"""
 
+    @classmethod
     def setUpClass(self):
         super().setUpClass(self)
```
simahawk, line 14: it's still broken, at least the naming: should be `cls` and you should not pass it :wink:

PR #830, opened 2024-06-05
[17.0][MIG] delivery_roulier_chronopost_fr: Migration to 17.0

yankinmax:
Depends on:
- [x] https://github.com/OCA/delivery-carrier/pull/828
- [x] https://github.com/OCA/delivery-carrier/pull/827
- [x] https://github.com/OCA/delivery-carrier/pull/826
- [x] https://github.com/OCA/delivery-carrier/pull/829

PR #862, opened 2024-07-17
[FIX] base_delivery_carrier_label test decorators

gurneyalex:
Previous fix was incomplete: the class method should only call other class methods

### DIFF
--- a/base_delivery_carrier_label/tests/carrier_label_case.py
+++ b/base_delivery_carrier_label/tests/carrier_label_case.py
@@ -10,75 +10,81 @@ class CarrierLabelCase(TransactionCase):
     to return the carrier you want to test"""
 
     @classmethod
-    def setUpClass(self):
-        super().setUpClass(self)
-        self._create_order_picking()
+    def setUpClass(cls):
+        super().setUpClass()
+        cls._create_order_picking()
 
     @property
     def transfer_in_setup(self):
         return True
 
-    def _create_order_picking(self):
+    @classmethod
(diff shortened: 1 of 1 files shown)


rousseldenis (approved):

base_delivery_carrier_label/tests/carrier_label_case.py
```diff
@@ -10,75 +10,81 @@ class CarrierLabelCase(TransactionCase):
     to return the carrier you want to test"""
 
     @classmethod
-    def setUpClass(self):
-        super().setUpClass(self)
-        self._create_order_picking()
+    def setUpClass(cls):
+        super().setUpClass(cls)
```
yankinmax: why not `super().setUpClass()`?

yankinmax:
@gurneyalex 
I've used your fix and tried to rewrite tests for `delivery_roulier_chronopost_fr`
As I've written here https://github.com/OCA/delivery-carrier/pull/830#issuecomment-2233782397
finally I think we shouldn't inherit from this test.

yankinmax (approved):

gurneyalex:
> @gurneyalex I've used your fix and tried to rewrite tests for `delivery_roulier_chronopost_fr` As I've written here [#830 (comment)](https://github.com/OCA/delivery-carrier/pull/830#issuecomment-2233782397) finally I think we shouldn't inherit from this test.

I will still get this merged, to have some correct code in the module.

PR #863, opened 2024-07-19
[14.0][ADD] delivery_purchase_label

TDu:
Print delivery labels for dropshipping purchase order. Useful when then vendor that will process the order does not have the capabilities to print the transporter labels needed for the delivery.

### DIFF
--- a/delivery_purchase_label/__manifest__.py
+++ b/delivery_purchase_label/__manifest__.py
@@ -1,0 +1,19 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Delivery Purchase Label",
+    "summary": "Allows printing carrier delivery labels for a dropshipping vendor.",
+    "version": "14.0.1.0.0",
+    "category": "Delivery",
+    "website": "https://github.com/OCA/delivery-carrier",
+    "author": "Camptocamp, BCIM, Odoo Community Association (OCA)",
+    "maintainers": ["TDu", "jbaudoux"],
+    "installable": True,
+    "license": "AGPL-3",
+    "depends": ["delivery", "purchase"],
+    "data": [
+        "data/stock_picking_type.xml",
+        "views/delivery_carrier.xml",
+        "views/purchase_order_views.xml",
+    ],
+}
--- a/delivery_purchase_label/data/stock_picking_type.xml
+++ b/delivery_purchase_label/data/stock_picking_type.xml
@@ -1,0 +1,9 @@
+<odoo noupdate="1">
+    <record id="picking_type_send_label" model="stock.picking.type">
+        <field name="name">Dropship Carrier Label</field>
+        <field name="code">internal</field>
+        <field name="sequence_code">LBL</field>
+        <field name="default_location_dest_id" ref="stock.stock_location_suppliers" />
+        <field name="default_location_src_id" ref="stock.stock_location_suppliers" />
+    </record>
+</odoo>
--- a/delivery_purchase_label/models/delivery_carrier.py
+++ b/delivery_purchase_label/models/delivery_carrier.py
@@ -1,0 +1,12 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import fields, models
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = "delivery.carrier"
+
+    purchase_label_picking_type = fields.Many2one(
+        string="Operation Type For Purchase Label", comodel_name="stock.picking.type"
+    )
--- a/delivery_purchase_label/models/mail_compose_message.py
+++ b/delivery_purchase_label/models/mail_compose_message.py
@@ -1,0 +1,37 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import api, models
+
+
+class MailComposer(models.TransientModel):
+    _inherit = "mail.compose.message"
+
+    @api.model
+    def generate_email_for_composer(self, template_id, res_ids, fields):
+        res = super().generate_email_for_composer(template_id, res_ids, fields)
+        if self.model != "purchase.order":
+            return res
+        if (
+            template_id
+            in self.env["purchase.order"]._mail_templates_to_not_attach_labels()
+        ):
+            return
+        purchase_orders = self.env["purchase.order"].browse(res_ids)
+        for order in purchase_orders:
+            # Add the labels generated when sending the order by email
+            label_picking = order.delivery_label_picking_id
+            if not label_picking:
+                continue
+            attachments = self.env["ir.attachment"].search(
+                [
+                    ("res_model", "=", label_picking._name),
+                    ("res_id", "=", label_picking.id),
+                ]
+            )
+            if attachments:
+                attachment_ids = (
+                    res[order.id].get("attachment_ids", []) + attachments.ids
+                )
+                res[order.id]["attachment_ids"] = attachment_ids
+        return res
--- a/delivery_purchase_label/models/purchase_order.py
+++ b/delivery_purchase_label/models/purchase_order.py
@@ -1,0 +1,150 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import SUPERUSER_ID, _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    vendor_label_carrier_id = fields.Many2one(
+        "delivery.carrier",
+        "Vendor Label Carrier",
+        domain=[("purchase_label_picking_type", "!=", False)],
+    )
+    delivery_label_picking_id = fields.Many2one(
+        "stock.picking",
+        "Delivery Label Picking",
+        store=True,
+        readonly=True,
+    )
+
+    @api.model
+    def _states_to_generate_delivery_label(self):
+        """Labels will be (re)generated only if the PO is in one of these states."""
+        return ["draft", "sent"]
+
+    @api.model
+    def _mail_templates_to_not_attach_labels(self):
+        return self.env.ref("purchase.email_template_edi_purchase_reminder").ids
+
+    def action_rfq_send(self):
+        self.ensure_one()
+        self._generate_purchase_delivery_label()
+        return super().action_rfq_send()
+
+    def button_cancel(self):
+        self._cancel_purchase_delivery_label_picking()
+        return super().button_cancel()
+
+    def _is_valid_for_vendor_labels(self):
+        self.ensure_one()
+        if self.state not in self._states_to_generate_delivery_label():
+            return False
+        if not self.dest_address_id:
+            return False
+        if not any(
+            product.type in ["product", "consu"]
+            for product in self.order_line.product_id
+        ):
+            return False
+        if not self.vendor_label_carrier_id.purchase_label_picking_type:
+            return False
+        return True
+
+    def _is_picking_label_uptodate(self):
+        """Check if the picking label needs to be regenerated."""
+        self.ensure_one()
+        if not self.delivery_label_picking_id:
+            return False
+        if self.delivery_label_picking_id.state != "done":
+            return False
+        pick = self.delivery_label_picking_id
+        moves_new = []
+        for line in self.order_line:
+            for value in line._prepare_stock_moves(self.delivery_label_picking_id):
+                moves_new.append(
+                    {key: value[key] for key in ("product_id", "product_uom_qty")}
+                )
+        move_prev = [
+            {"product_id": move.product_id.id, "product_uom_qty": move.product_uom_qty}
+            for move in pick.move_lines
+        ]
+        if len(move_prev) == len(moves_new):
+            if [
+                move_value for move_value in move_prev if move_value not in moves_new
+            ] == []:
+                return True
+        return False
+
+    def _generate_purchase_delivery_label(self):
+        """Create a transfer to generate the carrier labels."""
+        self.ensure_one()
+        if not self._is_valid_for_vendor_labels():
+            return
+        if not self.partner_id.property_stock_supplier.id:
+            raise UserError(
+                _(
+                    "You must set a Vendor Location for this partner %s",
+                    self.partner_id.name,
+                )
+            )
+        if self._is_picking_label_uptodate():
+            return
+
+        order = self.with_company(self.company_id)
+        picking = order._create_purchase_delivery_label_picking(carrier)
+
+        if order.delivery_label_picking_id:
+            order._cancel_purchase_delivery_label_picking()
+
+        order.delivery_label_picking_id = picking
+        picking.message_post_with_view(
+            "mail.message_origin_link",
+            values={"self": picking, "origin": self},
+            subtype_id=self.env.ref("mail.mt_note").id,
+        )
+
+    def _create_purchase_delivery_label_picking(self, carrier):
+        self.ensure_one()
+        values = self._get_purchase_delivery_label_picking_value(carrier)
+        picking = self.env["stock.picking"].with_user(SUPERUSER_ID).create(values)
+        moves = self.order_line._create_stock_moves(picking)
+        moves.location_id = picking.location_id
+        moves.location_dest_id = picking.location_dest_id
+        # Remove the link on the sale and purchase
+        # To not impact the delivered quantity on them
+        picking.sale_id = False
+        moves.sale_line_id = False
+        moves.purchase_line_id = False
+        picking.action_assign()
+        for move in picking.move_lines:
+            move.quantity_done = move.product_uom_qty
+        picking._action_done()
+        return picking
+
+    def _cancel_purchase_delivery_label_picking(self):
+        delivery_label_pickings = self.delivery_label_picking_id
+        for picking in delivery_label_pickings:
+            picking.cancel_shipment()
+        # Using wirte to by pass internal checks, not a problem
+        # because this is a fake move (Vendor to Vendor)
+        delivery_label_pickings.move_line_ids.write({"state": "cancel"})
+        delivery_label_pickings.move_lines.write({"state": "cancel"})
+        delivery_label_pickings.write({"state": "cancel"})
+
+    def _get_purchase_delivery_label_picking_value(self, carrier):
+        return {
+            "picking_type_id": carrier.purchase_label_picking_type.id,
+            "partner_id": self.dest_address_id.id,
+            "user_id": False,
+            "date": fields.Datetime.now(),
+            "origin": " - ".join(
+                [origin for origin in [self.origin, self.name] if origin]
+            ),
+            "location_dest_id": self.partner_id.property_stock_supplier.id,
+            "location_id": self.partner_id.property_stock_supplier.id,
+            "company_id": self.company_id.id,
+            "carrier_id": carrier.id,
+        }
--- a/delivery_purchase_label/models/purchase_order.py
+++ b/delivery_purchase_label/models/purchase_order.py
@@ -6,6 +6,10 @@ from odoo import SUPERUSER_ID, _, api, fields, models
 from odoo.exceptions import UserError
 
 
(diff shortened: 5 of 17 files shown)


delivery_purchase_label/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,8 @@
+On the partner form a new field `Purchase Delivery Method` allows to select
+the carrier that will be used by the supplier to send the goods.
+
+In "Inventory > Configuration > Delivery > Shipping Methods", a new field
+`Operation Type for Sending Labels`
+allows to select the opeartion type to use for the stock transfer that will
```
henrybackman:
```suggestion
allows to select the operation type to use for the stock transfer that will
```

delivery_purchase_label/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,15 @@
...
+This is done by creating and processing a stock transfer with the source
+and destination location set as the vendor location. And the carrier set.
+
+The transporter being used is the one set on the `Purchase Delivery Method` of the
+vendor in the "Sales & Purchase" tab.
+
+A new operation type is created by the module and used for the transfer.
+All attachment (should be only the labels) on the transfer are then attached
+to the email send to the vendor (rfq or po).
+When the purhcase order is still in draft the label are regenerated everytime
```
henrybackman:
```suggestion
When the purchase order is still in draft the labels are regenerated everytime
```

delivery_purchase_label/models/purchase_order.py
```diff
@@ -0,0 +1,93 @@
...
+        """Labels will be (re)generated only if the PO is in one of these states."""
+        return ["draft", "sent"]
+
+    @api.model
+    def _mail_templates_to_not_attach_labels(self):
+        return self.env.ref("purchase.email_template_edi_purchase_reminder").ids
+
+    def action_rfq_send(self):
+        self.ensure_one()
+        if self._states:
```
henrybackman: It's not obvious to me why the check on `_states` here. What does it mean?

delivery_purchase_label/models/purchase_order.py
```diff
@@ -0,0 +1,93 @@
...
+        self.ensure_one()
+        if not self.dest_address_id:
+            return False
+        return True
+
+    def _generate_purchase_delivery_label(self):
+        """Create a transfer to generate the carrier labels."""
+        self.ensure_one()
+        if self.state not in self._states_to_generate_delivery_label():
+            return
```
henrybackman: Could this and the following `product.type` check be in `_is_valid_for_vendor_labels()`?

delivery_purchase_label/models/purchase_order.py
```diff
@@ -0,0 +1,93 @@
...
+        if not self._is_valid_for_vendor_labels():
+            return
+        if not any(
+            product.type in ["product", "consu"]
+            for product in self.order_line.product_id
+        ):
+            return
+        # Find the carrier and that will be used
```
henrybackman:
```suggestion
        # Find the carrier that will be used
```

henrybackman (commented):
A couple of small comments

delivery_purchase_label/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Delivery Purchase Label",
+    "summary": "Allows printing carrier delivery labels for a dropshipping vendor.",
+    "version": "14.0.1.0.0",
+    "category": "Delivery",
+    "website": "https://github.com/OCA/delivery-carrier",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
```
jbaudoux:
```suggestion
    "author": "Camptocamp, BCIM, Odoo Community Association (OCA)",
```

delivery_purchase_label/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,15 @@
+Print delivery labels for dropshipping purchase order.
+Useful when then vendor that will process the order does not have the
```
vuwnevska:
```suggestion
Useful when the vendor that will process the order does not have the
```

delivery_purchase_label/models/purchase_order.py
TDu: Good catch this was a left over from refactoring

delivery_purchase_label/models/purchase_order.py
TDu: yes it could, changed.

delivery_purchase_label/models/mail_compose_message.py
```diff
@@ -0,0 +1,39 @@
...
+        res = super().generate_email_for_composer(template_id, res_ids, fields)
+        if self.model != "purchase.order":
+            return res
+        if (
+            template_id
+            in self.env["purchase.order"]._mail_templates_to_not_attach_labels()
+        ):
+            return
+        if isinstance(res_ids, int):
+            res_ids = [res_ids]
```
ivantodorovich:
```suggestion
```

This doesn't seem necessary

given `self.env["purchase.order"].browse(1) == self.env["purchase.order"].browse([1])`

ivantodorovich (approved):
Thanks !

It seems there are quite some lines not covered by unit tests. It might be a good idea to add a few more cases to them :)

LGTM

delivery_purchase_label/models/purchase_order.py
```diff
@@ -0,0 +1,93 @@
...
+            subtype_id=self.env.ref("mail.mt_note").id,
+        )
+
+    def _get_purchase_delivery_label_picking_value(self, carrier):
+        return {
+            "picking_type_id": carrier.purchase_label_picking_type.id,
+            "partner_id": self.dest_address_id.id,
+            "user_id": False,
+            "date": fields.Datetime.now(),
+            "origin": self.name,
```
jbaudoux: or maybe the SO number ? To clarify
```suggestion
            "origin": self.origin,
```

delivery_purchase_label/models/purchase_order.py
```diff
@@ -0,0 +1,93 @@
...
+        picking.action_assign()
+        for move in moves:
+            move.quantity_done = move.product_uom_qty
+        picking._action_done()
+        order.delivery_label_picking_id = picking
+        picking.message_post_with_view(
+            "mail.message_origin_link",
+            values={"self": picking, "origin": self},
+            subtype_id=self.env.ref("mail.mt_note").id,
+        )
```
mt-software-de: This will create a new transfer on every rfq send. 
We should check if something is created or if the create delivery transfer is still equal to the lines in the purchase.order, otherwise cancel the delivery transfer a create a new one.

delivery_purchase_label/models/purchase_order.py
```diff
@@ -0,0 +1,93 @@
...
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    delivery_label_picking_id = fields.Many2one(
+        "stock.picking",
+        "Delivery Label Picking",
+        store=True,
+        readonly=True,
+    )
```
mt-software-de, line 26: Could you please add this field to the view.

mt-software-de (changes requested):

delivery_purchase_label/models/purchase_order.py
```diff
@@ -0,0 +1,93 @@
...
+        values = order._get_purchase_delivery_label_picking_value(carrier)
+        picking = self.env["stock.picking"].with_user(SUPERUSER_ID).create(values)
+        moves = order.order_line._create_stock_moves(picking)
+        # Remove the link on the sale and purchase
+        # To not impact the delivered quantity on them
+        picking.sale_id = False
+        moves.sale_line_id = False
+        moves.purchase_line_id = False
+        picking.action_assign()
+        for move in moves:
```
mt-software-de: The moves can change on action_assign
```suggestion
        picking.action_assign()
        moves = picking.move_lines
        for move in moves:
```

delivery_purchase_label/data/stock_picking_type.xml
```diff
@@ -0,0 +1,9 @@
+<odoo noupdate="1">
+    <record id="picking_type_send_label" model="stock.picking.type">
+        <field name="name">Dropship Carrier Label</field>
+        <field name="code">internal</field>
+        <field name="sequence_code">LBL</field>
+        <field name="default_location_dest_id" ref="stock.stock_location_suppliers" />
+        <field name="default_location_src_id" ref="stock.stock_location_suppliers" />
```
mt-software-de, line 7: Why is this suppliers? And not customers?
```suggestion
        <field name="default_location_src_id" ref="stock.stock_location_suppliers" />
```

delivery_purchase_label/data/stock_picking_type.xml
```diff
@@ -0,0 +1,9 @@
+<odoo noupdate="1">
+    <record id="picking_type_send_label" model="stock.picking.type">
+        <field name="name">Dropship Carrier Label</field>
+        <field name="code">internal</field>
+        <field name="sequence_code">LBL</field>
+        <field name="default_location_dest_id" ref="stock.stock_location_suppliers" />
```
mt-software-de, line 6: Why is there the suppliers location used?
And not customers?
And why is the code internal and not outgoing?
What are the reasons for that? Stock valuation?

jbaudoux: It's on purpose. It's the supplier packing operation, so an internal operation done by the supplier. Goods are not yet sent.

mt-software-de: Yes and it is good for valuation layer, because if it would be supplier -> customer is would be recognized as dropship and will create a valuation for it.

delivery_purchase_label/models/purchase_order.py
mt-software-de, line 1: On cancel the created label picking should be as well cancelled.

delivery_purchase_label/models/delivery_carrier.py
```diff
@@ -0,0 +1,12 @@
...
+
+from odoo import fields, models
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = "delivery.carrier"
+
+    purchase_label_picking_type = fields.Many2one(
+        string="Operation Type For Purchase Label", comodel_name="stock.picking.type"
+    )
```
mt-software-de, line 12: Is this really needed to be set on the carrier?
Or is it enough to just simple configure the carrier on the partner?
And the for creating the purchase delivery label - just use the operation type from data xml?

This would result in less code and it would ensure that there is just one operation type for this matter. 

```suggestion
```

delivery_purchase_label/models/purchase_order.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def _generate_purchase_delivery_label(self):
+        """Create a transfer to generate the carrier labels."""
+        self.ensure_one()
+        if not self._is_valid_for_vendor_labels():
+            return
+        # Find the carrier that will be used
+        carrier = self.partner_id.purchase_delivery_carrier_id
+        if not carrier.purchase_label_picking_type:
+            return
```
mt-software-de: Is it needed to check the `purchase_label_picking_type` or can we assume that if the `purchase_delivery_carrier_id` is set on the partner we are just using the created operation type from the data file. 
```suggestion
        if not carrier:
            return
```

delivery_purchase_label/models/purchase_order.py
```diff
@@ -0,0 +1,95 @@
...
+        order.delivery_label_picking_id = picking
+        picking.message_post_with_view(
+            "mail.message_origin_link",
+            values={"self": picking, "origin": self},
+            subtype_id=self.env.ref("mail.mt_note").id,
+        )
+
+    def _get_purchase_delivery_label_picking_value(self, carrier):
+        return {
+            "picking_type_id": carrier.purchase_label_picking_type.id,
```
mt-software-de, line 142:
```suggestion
            "picking_type_id": self.env.ref("delivery_purchase_label.picking_type_send_label"),
```

mt-software-de (changes requested):

delivery_purchase_label/models/delivery_carrier.py
jbaudoux: The feature needs to be enabled somewhere. On the carrier is the best place and configuring the operation type is also legit

delivery_purchase_label/models/purchase_order.py
jbaudoux: It's better to configure than assume anything

delivery_purchase_label/models/purchase_order.py
jbaudoux: it is good as it is

delivery_purchase_label/models/delivery_carrier.py
mt-software-de: yes the feature needs to be enabled, but it is already enabled on the partner where the carrier is set.

mt-software-de: The background to this question is.
How do you find out which transfers are "purchase deliver label" transfers. 

The simplest way would be to have just one operation type for all of those transfers. This would mean no configuration on carrier just on the supplier. 

With the current implementation there could be a lot types used for that. 
Which makes it a lot harder to find the right transfers. 

A other possible way would be to mark those operation types as delivery purchase label type.

delivery_purchase_label/models/purchase_order.py
```diff
@@ -0,0 +1,132 @@
...
+        # Remove the link on the sale and purchase
+        # To not impact the delivered quantity on them
+        picking.sale_id = False
+        moves.sale_line_id = False
+        moves.purchase_line_id = False
+        picking.action_assign()
+        # Moves can change on action assign
+        moves = picking.move_lines
+        for move in moves:
+            move.quantity_done = move.product_uom_qty
```
mt-software-de: Could this please move into an own method.
Including also line 113 `picking._action_done`

delivery_purchase_label/models/purchase_order.py
```diff
@@ -0,0 +1,132 @@
...
+        if order.delivery_label_picking_id:
+            # Canceling the previous picking.
+            pick = order.delivery_label_picking_id
+            # Using wirte to by pass internal checks
+            # Should not be a problem because this is a fake
+            # move (Vendor to Vendor)
+            pick.move_line_ids.write({"state": "cancel"})
+            pick.move_lines.write({"state": "cancel"})
+            pick.write({"state": "cancel"})
```
mt-software-de: Could this also move into an own method. 
To make it hookable and please also call `pick.cancel_shipment`

mt-software-de (changes requested):

delivery_purchase_label/models/purchase_order.py
```diff
@@ -0,0 +1,132 @@
...
+                return True
+        return False
+
+    def _generate_purchase_delivery_label(self):
+        """Create a transfer to generate the carrier labels."""
+        self.ensure_one()
+        if not self._is_valid_for_vendor_labels():
+            return
+        # Find the carrier that will be used
+        carrier = self.partner_id.purchase_delivery_carrier_id
```
jbaudoux: We need a carrier on the PO as initially stated. Use cases showed that the vendor will not always ship with the same carrier. You can keep partner `purchase_delivery_carrier_id` as a fallback value but ideally we shouldn't configure any partner. We are generating drop-shipping PO and the PO carrier should come from the SO delivery carrier.
I would define a field on the PO like `vendor_label_carrier_id`.
This new field should be filled in when the PO is created/updated (done outside this module).
```suggestion
        carrier = self.vendor_label_carrier_id or self.partner_id.purchase_delivery_carrier_id
```

delivery_purchase_label/models/purchase_order.py
```diff
@@ -0,0 +1,132 @@
...
+                    self.partner_id.name,
+                )
+            )
+        if self._is_picking_label_uptodate():
+            return
+        order = self.with_company(self.company_id)
+        # Create and process the transer to send the labels
+        values = order._get_purchase_delivery_label_picking_value(carrier)
+        picking = self.env["stock.picking"].with_user(SUPERUSER_ID).create(values)
+        moves = order.order_line._create_stock_moves(picking)
```
jbaudoux: Low priority remark:
Instead of making a dry run with `_is_picking_label_uptodate`, you should create a savepoint, make the picking with the moves, then compare and rollback to savepoint if the moves content is the same.
Without this, you have no warranty your dry-run is doing the right thing as you are calling only a subset of the complete picking creation

delivery_purchase_label/models/purchase_order.py
jbaudoux: Yes, indeed, make a method for canceling the delivery label picking.
You need to call this also on PO cancellation as mentioned by MT

delivery_purchase_label/models/res_partner.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+    purchase_delivery_carrier_id = fields.Many2one(
```
jbaudoux: As this is used for label and not on the real transfer, I would rename to remove confusion
```suggestion
    purchase_label_carrier_id = fields.Many2one(
```

delivery_purchase_label/models/res_partner.py
```diff
@@ -0,0 +1,14 @@
...
+        "delivery.carrier",
+        string="Purchase Delivery Method",
```
jbaudoux:
```suggestion
        string="Purchase Label Delivery Method",
```

jbaudoux (changes requested):
Some remarks.
Canceling the PO should cancel the generated picking for label. That's a major issue that needs to be tackled.

delivery_purchase_label/models/purchase_order.py
mt-software-de: Yes use self.origin this makes it much easier to generate the label out the purchae delivery label record. 
It represents then more or less a copy of the initial purchase order.

jbaudoux: Don't we wanted both on label? If we concatenate the two doesn't it solve the label?
"%origin - %name" ?

mt-software-de: ping @TDu

delivery_purchase_label/models/purchase_order.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def _generate_purchase_delivery_label(self):
+        """Create a transfer to generate the carrier labels."""
+        self.ensure_one()
+        if not self._is_valid_for_vendor_labels():
+            return
+        # Find the carrier that will be used
+        carrier = self.vendor_label_carrier_id
+        if not carrier.purchase_label_picking_type:
+            return
```
jbaudoux: Shouldn't this be part of `_is_valid_for_vendor_labels` ?

delivery_purchase_label/models/purchase_order.py
jbaudoux: seems tackled now

delivery_purchase_label/models/purchase_order.py
mt-software-de: This is exactly what is causing problems now. Without the dry-run you are comparing the ordered products and qties against a maybe different transferred products. The moves are not the same as the ordered for example for products with a kit bom.

TDu: Knowing that the `mrp` module will change the moves when the procurement is being run. This is probably too complex to reproduce here.
Would checking  that the purchase lines have not changed (using the `write_date`) since the current delivery label picking has been generated be a good enough solution ?

TDu: Finally using a savepoint

delivery_purchase_label/models/purchase_order.py
mt-software-de: yes

mt-software-de (approved):
LG. Just the minor remark of @jbaudoux https://github.com/OCA/delivery-carrier/pull/863#discussion_r1773139156

delivery_purchase_label/models/purchase_order.py
TDu: Done

jbaudoux (approved):
Can you squash fixups ?

PR #886, opened 2024-09-23
[14.0][FIX] commercialName is mandatory for dropoff site so we can add it every time

bguillot:
commercialName is not an mandatory field in basic case but mandatoiry for dropoff site delivery.

I propose to send it every time to avoid other delivery types issues instead of adding it in only in laposte_dropoff_site module


florian-dacosta (approved):

hparfr (approved):

rousseldenis:
@bguillot For next time, don't forget to mention module name in your commit message. Thanks

PR #887, opened 2024-09-23
[MIG][IMP] delivery_multi_destination: Introduce 'based on destination' delivery type

carmenbianca:
Backported from #871 
Internal task T12787

### DIFF
--- a/delivery_multi_destination/wizards/choose_delivery_carrier.py
+++ b/delivery_multi_destination/wizards/choose_delivery_carrier.py
@@ -1,0 +1,18 @@
+# SPDX-FileCopyrightText: 2024 Coop IT Easy SC
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+from odoo import api, models
+
+
+class ChooseDeliveryCarrier(models.TransientModel):
+    _inherit = "choose.delivery.carrier"
+
+    @api.onchange("carrier_id")
+    def _onchange_carrier_id(self):
+        result = super()._onchange_carrier_id()
+        if self.delivery_type == "base_on_destination":
+            vals = self._get_shipment_rate()
+            if vals.get("error_message"):
+                return {"error": vals["error_message"]}
+        return result
--- a/delivery_multi_destination/wizards/choose_delivery_carrier.py
+++ b/delivery_multi_destination/wizards/choose_delivery_carrier.py
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
-from odoo import api, models
+from odoo import _, api, models
 
 
 class ChooseDeliveryCarrier(models.TransientModel):
@@ -13,6 +13,15 @@ class ChooseDeliveryCarrier(models.TransientModel):
         result = super()._onchange_carrier_id()
         if self.delivery_type == "base_on_destination":
             vals = self._get_shipment_rate()
+            # although ._onchange_carrier_id() in the delivery module can
+            # return a dict with an "error" key and a string value, this is
+            # not handled by the caller (see BaseModel._onchange_eval()),
+            # which only handles dicts with a "warning" key and a dict as a
+            # value.
             if vals.get("error_message"):
-                return {"error": vals["error_message"]}
+                return {
+                    "warning": {"title": _("Error"), "message": vals["error_message"]}
+                }
+            if vals.get("warning_message"):
+                return {"warning": {"message": vals["warning_message"]}}
         return result
--- a/delivery_multi_destination/__init__.py
+++ b/delivery_multi_destination/__init__.py
@@ -1,3 +1,4 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from . import models
+from . import wizards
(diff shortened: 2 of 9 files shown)


polchampion:
@carmenbianca tested functionally [on the runboat](http://oca-delivery-carrier-16-0-pr887-88a7398d1f76.runboat.odoo-community.org/shop/payment)

- I created some multiple destination delivery methods with non fixed children
- On the delivery selection page, the available carrier rate is computed right away
- The other carrier rates are displayed loading instead of being hidden
- I get the error below on the delivery selection page:

screenshot-oca-delivery-carrier-16-0-pr887-88a7398d1f76 runboat odoo-community org-2024 10 29-16_28_06


```UncaughtClientError > TypeError

Uncaught Javascript Error > browser.navigator.clipboard is undefined

TypeError: browser.navigator.clipboard is undefined
    onClickClipboard@http://oca-delivery-carrier-16-0-pr887-88a7398d1f76.runboat.odoo-community.org/web/assets/641-afd6f82/1/web.assets_frontend_lazy.min.js:6760:20
    mainEventHandler@http://oca-delivery-carrier-16-0-pr887-88a7398d1f76.runboat.odoo-community.org/web/assets/641-afd6f82/1/web.assets_frontend_lazy.min.js:2063:77
    listener@http://oca-delivery-carrier-16-0-pr887-88a7398d1f76.runboat.odoo-community.org/web/assets/641-afd6f82/1/web.assets_frontend_lazy.min.js:1233:15```

**Comment by huguesdk** [comment]
@polchampion could you try again? it seems that the error you reported is unrelated to the changes. how to reproduce it?

**Comment by polchampion** [comment]
@huguesdk Can't reproduce the above error, but I tested anew and couldn't make it work.
Tested on [this test db](https://odoo16.main.test.srv.coopiteasy.be/web/login?db=rotor-v16-test).

- I created a multi-destination delivery method called "xtra multi", with two non-fixed children. The price calculation is based on rules and depends on the product weight.
- Odoo can neither calculate the price [in the backend](https://odoo16.main.test.srv.coopiteasy.be/web?db=rotor-v16-test#id=43&cids=1&menu_id=236&action=386&model=sale.order&view_type=form) nor on the website:

screenshot-odoo16 main test srv coopiteasy be-2025 02 25-10_50_40

screenshot-odoo16 main test srv coopiteasy be-2025 02 25-10_50_47

 
screenshot-odoo16 main test srv coopiteasy be-2025 02 25-10_51_44


I'm perhaps wrong on the multi-destination usage, but I'm otherwise not sure how to best test the wanted behaviour. I believe the test should work in any case.

**Comment by huguesdk** [comment]
`delivery_multi_destination` is only about defining different carriers depending on the shipping destination. i think that the rules only allow to compute a different price for the same carrier.

**Review by polchampion — APPROVED** [review]
Functionally tested and LGTM

**Comment by huguesdk** [comment]
@OCA/logistics-maintainers could someone review or merge this? i could review it myself, but i made some changes myself, so interest is conflicting here.

File: delivery_multi_destination/wizards/choose_delivery_carrier.py [hunk 1]
```diff
@@ -0,0 +1,18 @@
...
+class ChooseDeliveryCarrier(models.TransientModel):
+    _inherit = "choose.delivery.carrier"
+
+    @api.onchange("carrier_id")
+    def _onchange_carrier_id(self):
+        result = super()._onchange_carrier_id()
+        if self.delivery_type == "base_on_destination":
+            vals = self._get_shipment_rate()
+            if vals.get("error_message"):
+                return {"error": vals["error_message"]}
```
hparfr: what  does {"error":} do and where ?

huguesdk: thanks for the review! this reproduces [the behavior of the parent model](https://github.com/OCA/OCB/blob/16.0/addons/delivery/wizard/choose_delivery_carrier.py#L34) (in the `delivery` module), but after some digging, i found that an `onchange` method can return a dictionary with a `warning` key (as explained in [the doc](https://www.odoo.com/documentation/16.0/developer/reference/backend/orm.html#odoo.api.onchange)), but that nothing will happen if the dictionary contains an `error` key (as can be seen in [the code handling the return value](https://github.com/OCA/OCB/blob/16.0/odoo/models.py#L6357)). good catch!

i also noticed that the dictionary returned by `._get_shipment_rate()` [can contain a `warning_message` key](https://github.com/OCA/OCB/blob/16.0/addons/delivery/models/delivery_carrier.py#L180). what would you suggest? something like this?

```python
            if vals.get("error_message"):
                return {
                    "warning": {"title": _("Error"), "message": vals["error_message"]}
                }
            if vals.get("warning_message"):
                return {"warning": {"message": vals["warning_message"]}}
```

hparfr: Yes sounds good.

hparfr (approved):

hparfr:
@huguesdk please rebase

huguesdk:
@hparfr i started rebasing to make it mergeable, but was interrupted. it’s done now.

hparfr (commented):
/ocabot merge nobump

huguesdk:
@hparfr many thanks!

PR #888, opened 2024-10-01
[16.0][IMP] delivery_deliverea: send incoterms

Samuel-res:
Add function to improve the incoterms in the information we sent to Deliverea


dalonsofl:
My bad PR, continue here -> https://github.com/OCA/delivery-carrier/pull/918

PR #898, opened 2024-10-29
[FIX] delivery_postlogistics: raise exception when no token received

ajaniszewska-dev:
Sometimes due to API error it may happen we don't receive anything back when requesting token. In this case we want to show meaningful message.

### DIFF
--- a/delivery_postlogistics/postlogistics/web_service.py
+++ b/delivery_postlogistics/postlogistics/web_service.py
@@ -9,11 +9,13 @@ import threading
 import urllib.parse
 from datetime import datetime, timedelta
 from io import BytesIO
+from json import JSONDecodeError
 
 import requests
 from PIL import Image
 
 from odoo import _, exceptions
+from odoo.exceptions import UserError
 
 _logger = logging.getLogger(__name__)
 
@@ -450,7 +452,22 @@ class PostlogisticsWebService(object):
             },
             timeout=60,
         )
-        return response.json()
+
+        try:
+            response.raise_for_status()
+            json_response = response.json()
+        except (
+            JSONDecodeError,
+            requests.exceptions.HTTPError,
+        ) as error:
+            raise UserError(
+                _(
+                    "Postlogistics service is not accessible at the moment. Error code: %s. "
+                    "Please try again later." % (response.status_code or "None")
+                )
+            ) from error
+
+        return json_response
 
     @classmethod
     def get_access_token(cls, picking_carrier):
--- a/delivery_postlogistics/tests/test_postlogistics.py
+++ b/delivery_postlogistics/tests/test_postlogistics.py
@@ -4,6 +4,8 @@ from os.path import dirname, join
 
 from vcr import VCR
 
+from odoo.exceptions import UserError
+
 from .common import TestPostlogisticsCommon
 
 recorder = VCR(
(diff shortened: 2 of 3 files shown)


grindtildeath (approved):

twalter-c2c (approved):
LGTM

delivery_postlogistics/postlogistics/web_service.py
```diff
@@ -450,7 +450,20 @@ def _request_access_token(cls, delivery_carrier):
...
+        response.raise_for_status()
+
+        try:
+            json_response = response.json()
+        except requests.exceptions.JSONDecodeError as exc:
+            if not json_response:
+                raise exceptions.UserError(
+                    _("API response without content - no access token received.")
+                ) from exc
+            raise
```
mmequignon:
```suggestion
        try:
            json_response = response.json()
            response.raise_for_status()
        except requests.exceptions.JSONDecodeError as exc:
            raise exceptions.UserError(
                _("API response without content - no access token received.")
            ) from exc
        except requests.exceptions.HTTPError as exc:
            raise exceptions.UserError(
                _("Raise the message you think is better here")
            ) from exc
```
This is supposed to raise an exception when response code isn't 200 see the [documentation](https://3.python-requests.org/user/quickstart/#response-status-codes)

Shouldn't this be in the try/except clause?

mmequignon (changes requested):

delivery_postlogistics/postlogistics/web_service.py
grindtildeath: > This is supposed to raise an exception when response code isn't 200 see the [documentation](https://3.python-requests.org/user/quickstart/#response-status-codes)

No, it will not raise if code is a 2xx. Empty content is a 204 and `raise_for_status` will not raise anything in that case.

Also `raise_for_status` needs to happen before `response.json`, because if HTTP code is 4xx or 5xx, you want the corresponding error to be raised before attempting to convvert the response to json.

mmequignon: Yeah, 2XX is what I meant.
Anyways, this needed to happen in the `try` block.

mmequignon (approved):

ajaniszewska-dev:
do not merge ATM, seems this exception is not correctly raised (raised when service works fine)

henrybackman (approved):

delivery_postlogistics/tests/test_postlogistics.py
```diff
@@ -108,3 +110,9 @@ def test_postlogistics_rate_shipment(self):
             res = self.carrier.postlogistics_rate_shipment(None)
             self.assertEqual(len(cassette.requests), 2)
         self.assertEqual(res["price"], 1.0)
+
+    def test_postlogistics_get_token_error(self):
+        with recorder.use_cassette("test_token_error") as cassette:
+            with self.assertRaises(UserError):
```
simahawk: It's always better to use `assertRaisesRegex` w/ a pattern matching the error msg.
Too often I've seen an exception raised somewhere else :wink:

simahawk (changes requested):

delivery_postlogistics/tests/test_postlogistics.py
ajaniszewska-dev: thanks @simahawk , updated test.

simahawk (approved):

PR #900, opened 2024-10-31
[18][ADD/MIG] delivery_carrier_option (extracted from base_delivery_carrier_label module)

florian-dacosta:
This one should have been done a long time ago IMO
It is a clear independent feature from base_delivery_carrier_label and all carriers don't necessarly needs it.


delivery_carrier_option/models/stock_picking.py
```diff
@@ -0,0 +1,48 @@
...
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    option_ids = fields.Many2many(
+        comodel_name="delivery.carrier.option",
+        string="Options",
+        compute="_compute_option_ids",
+        store=True,
+        readonly=False,
+    )
```
StephaneMangin, line 21: I wonder if reusing an overriden "paper.format" should avoid to create this model. And if a new configuration is needed, a new format should be created. This could decoupled the postlogictic's logic to a wider reusable concept (for other implementations like QuickPAC, etc)

It is not mandatory to adresse this for now. But could be a nice future to be as close as possible regarding the existing features in standard (IMHO the paper format is very very close to this concept, and i presume close enough to reuse)

StephaneMangin (approved):

delivery_carrier_option/models/delivery_carrier_template_option.py
```diff
@@ -0,0 +1,20 @@
...
+from odoo import fields, models
+
+
+class DeliveryCarrierTemplateOption(models.Model):
+    """Available options for a carrier (partner)"""
+
+    _name = "delivery.carrier.template.option"
+    _description = "Delivery carrier template option"
+
+    partner_id = fields.Many2one(comodel_name="res.partner", string="Partner Carrier")
```
StephaneMangin, line 24:
```suggestion
    type = fields.Selection(
        selection=[
            ("basic", "Basic Service"),
            ("additional", "Additional Service"),
            ("delivery", "Delivery Instructions"),
            ("partner_option", "Partner Option"),
        ],
        default="basic",
        string="Option type"
    )
    partner_id = fields.Many2one(comodel_name="res.partner", string="Partner Carrier")
```

We could directly include this field from postlogistics (partial elements). Because this information is generic enough to put it there IMHO.

delivery_carrier_option/views/delivery_carrier_template_option.xml
```diff
@@ -0,0 +1,35 @@
...
+        <field name="model">delivery.carrier.template.option</field>
+        <field name="arch" type="xml">
+            <form string="delivery_carrier_option">
+                <group>
+                    <group>
+                        <field name="name" />
+                    </group>
+                    <group>
+                        <field name="partner_id" />
+                        <field name="code" />
```
StephaneMangin:
```suggestion
                        <field name="type" />
                        <field name="code" />
```

delivery_carrier_option/views/delivery_carrier_template_option.xml
```diff
@@ -0,0 +1,35 @@
...
+
+    <record id="delivery_carrier_template_option_view_list" model="ir.ui.view">
+        <field
+            name="name"
+        >delivery_base.delivery_carrier_template_option.view_list</field>
+        <field name="model">delivery.carrier.template.option</field>
+        <field name="arch" type="xml">
+            <list>
+                <field name="partner_id" />
+                <field name="code" />
```
StephaneMangin, line 32:
```suggestion
                <field name="type" />
                <field name="code" />
```

StephaneMangin (approved):
Behavior addition proposal (no tests at this stage because no related processing)

StephaneMangin:
cc @florian-dacosta  What do you think ?

leemannd (approved):

delivery_carrier_option/models/delivery_carrier_template_option.py
florian-dacosta: Hello @StephaneMangin Thanks for the suggestion.
The only purpose is informative ? There is no additional logic about this field ?
I don't really need it, but I have nothing against I guess

delivery_carrier_option/models/stock_picking.py
florian-dacosta: I did not understand what you meant, I don't get the relation between the carrier options and a "paper.format" ?

delivery_carrier_option/models/delivery_carrier_option.py
```diff
@@ -0,0 +1,53 @@
...
+        help="By check, user can choose to apply this option "
+        "to each Delivery Order\n using this delivery method",
+    )
+    tmpl_option_id = fields.Many2one(
+        comodel_name="delivery.carrier.template.option",
+        string="Option",
+        required=True,
+        ondelete="cascade",
+    )
+    carrier_id = fields.Many2one(comodel_name="delivery.carrier", string="Carrier")
```
vvrossem:
```suggestion
    carrier_id = fields.Many2one(comodel_name="delivery.carrier")
```

delivery_carrier_option/models/delivery_carrier_template_option.py
```diff
@@ -0,0 +1,30 @@
...
+
+    type = fields.Selection(
+        selection=[
+            ("basic", "Basic Service"),
+            ("additional", "Additional Service"),
+            ("delivery", "Delivery Instructions"),
+            ("partner_option", "Partner Option"),
+        ],
+        default="basic",
+        string="Option type",
```
vvrossem:
```suggestion
        string="Option Type",
```

delivery_carrier_option/models/delivery_carrier_template_option.py
```diff
@@ -0,0 +1,30 @@
...
+    )
+    partner_id = fields.Many2one(comodel_name="res.partner", string="Partner Carrier")
+    name = fields.Char(readonly=True)
+    code = fields.Char(readonly=True)
+    description = fields.Char(
+        readonly=True,
+        help="Allow to define a more complete description " "than in the name field.",
+    )
```
vvrossem:
```suggestion
    name = fields.Char()
    code = fields.Char()
    description = fields.Char(
        help="Allow to define a more complete description " "than in the name field.",
    )
```

Would it makes sense to move the `readonly` attribute to the views? 

> If a field is marked as readonly=True on the model, it won't be possible to import it through the Odoo import tool, so avoid it as possible and define the readonly attribute on the views instead.

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

delivery_carrier_option/tests/test_options.py
```diff
@@ -0,0 +1,49 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import TransactionCase
+
+
+class TestPickingOption(TransactionCase):
```
vvrossem:
```suggestion
from odoo.addons.base.tests.common import BaseCommon


class TestPickingOption(BaseCommon):
```

> Use `BaseCommon` as base test class to reduce overhead from tracking and boost the test suite.

delivery_carrier_option/views/delivery_carrier_template_option.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="delivery_carrier_template_option_view_form" model="ir.ui.view">
+        <field name="name">delivery_base.delivery_carrier_option.view_form</field>
+        <field name="model">delivery.carrier.template.option</field>
+        <field name="arch" type="xml">
+            <form string="delivery_carrier_option">
+                <group>
+                    <group>
+                        <field name="name" />
```
vvrossem:
```suggestion
                        <field name="name" readonly="True" />
```

delivery_carrier_option/views/delivery_carrier_template_option.xml
```diff
@@ -0,0 +1,37 @@
...
+                    </group>
+                    <group>
+                        <field name="partner_id" />
+                        <field name="type" />
+                        <field name="code" />
```
vvrossem:
```suggestion
                        <field name="code" readonly="True" />
```

delivery_carrier_option/views/delivery_carrier_template_option.xml
```diff
@@ -0,0 +1,37 @@
...
+                    </group>
+                    <field name="description" />
```
vvrossem:
```suggestion
                    <field name="description" readonly="True" />
```

delivery_carrier_option/models/delivery_carrier_template_option.py
florian-dacosta: I don't know, do you need to import it ?
The template options have to be integrated with the carrier api, so it should be part of  the specific carrier module implementation, and not be created by import right ?

Anyway, if someone using this module needs it, I am ok to change it , but if it is not the case, I would rather not.

delivery_carrier_option/models/stock_picking.py
StephaneMangin: It was an idea to separated concerns.
- Option for label printing reusing standard printing model (A4, PDF, 600dpi, etc)
- Operational delivery services (Priority, With signature, etc) -> the "real" delivery options
But i presume this could maybe increase functional usage and probably implementation too.

Nevermind because i mixed behaviors from postlogistics using labeling configurations...

delivery_carrier_option/models/delivery_carrier_template_option.py
StephaneMangin: These options templating could be imported on the fly. So it is better to restrict on the view instead on the model. That's a very light change, so better to apply it.

delivery_carrier_option/models/delivery_carrier_template_option.py
StephaneMangin: It's enough generalistic to be implemented here. And indeed, it is just informative at this stage. (but not on the delivery_postlogistics module anyway)

florian-dacosta:
Hello @vvrossem I've included all your suggestions
Can you check ?

hparfr (approved):

vvrossem (approved):

PR #902, opened 2024-11-11
[16.0][ADD] delivery_package_sscc

thomaspaulb:
This module allows to generate [SSCC](https://www.gs1.org/standards/id-keys/sscc) codes for packages automatically.

Eg `stock.quant.package.name` changes from PACK000001..000002 etc to an actual valid SSCC code, taken from a range that can be configured.

image

This also makes Odoo's [stock.quant.package.valid_sscc](https://github.com/OCA/OCB/blob/49c64fdff21bff61aa6969ef941b7d117459eeef/addons/stock/models/stock_quant.py#L1208) flag True for any newly created package.


delivery_package_sscc/readme/USAGE.rst
```diff
@@ -0,0 +1,18 @@
+To use this module, you need to:
+
+* Identify your SSCC range. For example: `123456789012xxxxx`.
+* Go to the form view of the `ir.sequence` record for your packages.
```
sebalix: I will make this bullet point less technical, maybe by adding the menu path to follow?
```suggestion
* Go to the form view of the sequence record for your packages.
```

sebalix (commented):
Nice addition, but should not this module lands in https://github.com/OCA/stock-logistics-tracking or https://github.com/OCA/stock-logistics-barcode ?
It's not linked directly to delivery carriers IMO, so the module should be renamed `stock_quant_package_sscc` for instance.

thomaspaulb:
@sebalix You're probably right, I was confused as to where to place it myself. It seems to be linked to delivery in my mind because it's a tracking code for the packages that get sent out. Which of the two would you suggest to put it?

sebalix:
I would go for the tracking one, but let ask others opinions
cc @OCA/logistics-maintainers

jbaudoux (changes requested):
Thanks for the contribution. Looks good.
Move to tracking and rename like sebalix suggested

PR #904, opened 2024-11-18
[14.0] [ADD] delivery_roulier_picking_batch

paradoxxxzero:
This module allows to generate a unique delivery label/tracking for a whole batch of
pickings.

The batch pickings operations will be grouped in a single delivery package if they are
not already in a delivery package. In case of a batch with multiple packages, a label
per package will be created.

This module is only compatible with `delivery_roulier` carriers.


delivery_roulier_picking_batch/models/stock_picking_batch.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def action_done(self):
+        if not self.carrier_id or not (
+            self.carrier_id.integration_level == "rate_and_ship"
+            and self.picking_type_id.code != "incoming"
+        ):
+            return super().action_done()
+
+        self.ensure_one()
+        pickings = self.mapped("picking_ids").filtered(
```
sebastienbeau:
```suggestion
        pickings = self.picking_ids.filtered(
```

delivery_roulier_picking_batch/models/stock_picking.py
```diff
@@ -0,0 +1,123 @@
...
+                    self.env["stock.move.line"]
+                    .search(
+                        [
+                            "|",
+                            ("result_package_id", "=", package.id),
+                            ("package_id", "=", package.id),
+                        ]
+                    )
+                    .mapped("picking_id")
+                )
```
sebastienbeau, line 47: We can make it simplier.
We should just check that all picking of the batch have the same delivery address

paradoxxxzero: I'd rather work at package level since it's them that will be sent and you may want to send 2 packages from different pickings to different addresses.

sebastienbeau: In a functional point of view. When you use the batch delivery carrier you want to prepare all the picking for the same address, if user can mix, it will be a nightmare ;).
I am in favors of forcing to set a partner_id on the stock.picking.batch so it will be more understandable for the customer

PR #907, opened 2024-11-22
[UPD] delivery_deliverea: Update error message displayed

dalonsofl:
### Module

delivery_deliverea

### Describe the situation

Update the error display to only show the error returned by Deliverea instead of showing all errors in the response.

**Affected versions**:

16.0

### DIFF
--- a/delivery_deliverea/models/deliverea_request.py
+++ b/delivery_deliverea/models/deliverea_request.py
@@ -73,28 +73,11 @@ class DelivereaRequest(object):
             if isinstance(error, list):
                 error = error[0]
             return_code = error.get("code")
-            message = error.get("message")
-            detail = error.get("detail", "")
             if return_code:
                 raise UserError(
-                    _("%(name)s: %(rcode)s %(message)s %(detail)s %(ccode)s")
-                    % {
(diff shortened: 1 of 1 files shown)


delivery_deliverea/models/deliverea_request.py
```diff
@@ -73,28 +73,14 @@ def _check_error(self, res):
...
-                        else detail or "",
-                        "ccode": "\n{}: {} {}".format(
-                            _("Carrier Error"),
-                            error.get("carrierCode"),
-                            error.get("carrierMessage"),
-                        )
-                        if error.get("carrierCode")
-                        else "",
-                    }
+                    _("%s")
```
rousseldenis: Do you plan to translate every response ?

Nevertheless, you should use translation function arguments instead.

rousseldenis (changes requested):

delivery_deliverea/models/deliverea_request.py
dalonsofl: You are right! Besides, the messages in response are already translated into Spanish (in my case), so it is not necessary to transalate one by one each message.

Thanks!

PR #914, opened 2024-12-05
[FIX] [16.0] delivery_auto_refresh: Allow override carrier from wizard when changing partner

### LINKED ISSUE #873: [16] delivery_auto_refresh doesn't update if customer is changed
## Module
delivery_auto_refresh

## Describe the bug
delivery_auto_refresh doesn't update if customer is changed

## To Reproduce
**Affected versions**:
16 At least
1. Create a new SO
2. Add a customer and save. The delivery method will be added if it is set in the customer.
3. Modify the customer, add another customer with a different delivery method, and save. The delivery method will not be updated. 

**Expected behavior**
The delivery method should be updated if the customer is changed.
**Additional context**
https://www.loom.com/share/fc71fd79c4c84ea689bd73858ed6c405?sid=eb7b80ad-0401-4cb2-afa4-6e23709d86e0
@moduon MT-6943

Shide:
Always overrides the carrier when partner or delivery partner changes

closes https://github.com/OCA/delivery-carrier/issues/873

MT-6943 @moduon @rafaelbn @Gelojr @fcvalgar @jbaudoux @simahawk please review if you want :)

### DIFF
--- a/delivery_auto_refresh/models/sale_order.py
+++ b/delivery_auto_refresh/models/sale_order.py
@@ -25,7 +25,14 @@ class SaleOrder(models.Model):
     # End migration note
 
     # Migration Note 17.0: move this section to module sale_order_carrier_auto_assign
-    def _set_delivery_carrier(self, set_delivery_line=True):
+    def _set_delivery_carrier(
+        self, set_delivery_line=True, preserve_order_carrier=True
+    ):
+        """Automatically change delivery carrier.
+
+        :param set_delivery_line: It will create or update the delivery line
+        :param preserve_order_carrier: It will respect the carrier set on the order
+        """
         for order in self:
             delivery_wiz_action = order.action_open_delivery_wizard()
             delivery_wiz_context = delivery_wiz_action.get("context", {})
@@ -38,7 +45,7 @@ class SaleOrder(models.Model):
             )
 
             # Do not override carrier
-            if order.carrier_id:
+            if preserve_order_carrier and order.carrier_id:
                 delivery_wiz.carrier_id = order.carrier_id
 
             # If the carrier isn't allowed, we won't default to it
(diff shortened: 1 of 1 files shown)


Shide:
@jbaudoux why you don't want to override carrier when changing partner? I'm missing something, or you do this to cover some case?

delivery_auto_refresh/models/sale_order.py
```diff
@@ -37,10 +37,6 @@ def _set_delivery_carrier(self, set_delivery_line=True):
                 .new({})
             )
 
-            # Do not override carrier
-            if order.carrier_id:
-                delivery_wiz.carrier_id = order.carrier_id
```
jbaudoux: If the order is confirmed, you don't want to change the carrier (here: https://github.com/OCA/sale-workflow/blob/16.0/sale_order_carrier_auto_assign/models/sale_order.py#L24).
I wanted to drop this code as you have the same in the other module.
if you choose on the SO a carrier that is different from the partner default, then I don't want to override it on SO confirmation. Maybe we need an extra parameter then

simahawk: Indeed. You can have a new flag on the company `sale_auto_assign_carrier_override_existing` w/ default False.

jbaudoux: Euh no, not on the company, in the method signature... Nobody wants to loose the carrier set on the SO when you confirm it

Gelojr (approved):
LGTM Nice job @Shide 
https://www.loom.com/share/c4df265876224193aa95df01405e8534?sid=d94f0909-b8cf-4ce5-ba6d-1068a6c675fa

Shide:
@jbaudoux done as I think you prefer it.

Note that sale_order_carrier_auto_assign on v17 has not been migrated following your Migration notes (function has not been renamed) and may need extra work when this change goes to v17 on a migration (I don't think I migrate this module because we are skipping v17 version)

yajo (approved):

simahawk (approved):

jbaudoux (approved):

PR #915, opened 2024-12-06
[18.0][MIG] delivery_free_fee_removal: Migration to 18.0

LucasTran380381:


delivery_free_fee_removal/models/sale_order.py
```diff
@@ -11,9 +11,11 @@ def action_confirm(self):
         res = super().action_confirm()
         # Invoice all the free delivery line on order confirmation
         # Or the order will never be fully invoiced.
-        delivery_lines = self.order_line.filtered(
-            lambda line: line.order_id.state == "sale" and line.is_free_delivery
-        )
-        for line in delivery_lines:
-            line.qty_delivered = line.qty_invoiced = line.product_uom_qty
+        for order in self:
```
rousseldenis: Was the former code problematic ?

rousseldenis (approved):

delivery_free_fee_removal/models/sale_order.py
LucasTran380381: not problem but method `action_confirm` maybe call many sale_orders so I add loop order record

rousseldenis: Yes, but from >= v16, all relation fields can be joined together.

So, you can write `self.order_line.order_id.partner_id` without problems. It will return all the partners as a recordset.

LucasTran380381: So it not necessary loop order record, but I sill use filtered order state direly?

rousseldenis: Yes you can as it removes a carriage return for readability

LucasTran380381: I think former code is fine and more maintainable than my code so I revert this changes

ajaniszewska-dev (approved):

sebalix (approved):

PR #917, opened 2024-12-13
[15.0][IMP] delivery_package_number: Allow force set package number at partner level and picking operation

sergio-teruel:
cc @Tecnativa TT51855

ping @carlosdauden @chienandalu 

Migration script missing

### DIFF
--- a/delivery_package_number/models/stock_picking_type.py
+++ b/delivery_package_number/models/stock_picking_type.py
@@ -6,7 +6,12 @@ from odoo import fields, models
 class StockPickingType(models.Model):
     _inherit = "stock.picking.type"
 
-    force_set_number_of_packages = fields.Boolean()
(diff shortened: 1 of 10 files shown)


delivery_package_number/models/stock_picking_type.py
```diff
@@ -6,7 +6,12 @@
 class StockPickingType(models.Model):
     _inherit = "stock.picking.type"
 
-    force_set_number_of_packages = fields.Boolean()
+    force_set_number_of_packages = fields.Selection(
+        [
+            ("always", "Always"),
+            ("partner", "Partner condition"),
+        ]
+    )
```
chienandalu, line 14: You will need a migration script for this

sergio-teruel: Yes! I know, i wrote in my comment

PR #918, opened 2024-12-16
[16.0][IMP] delivery_deliverea: send incoterms

dalonsofl:
Add function to improve the incoterms in the information we sent to Deliverea

### DIFF
--- a/delivery_deliverea/models/delivery_carrier.py
+++ b/delivery_deliverea/models/delivery_carrier.py
@@ -414,6 +414,7 @@ class DeliveryCarrier(models.Model):
                 or "",
             },
         }
+        self._get_bulky_deliverea(picking, payload)
         self._delete_empty_values(payload)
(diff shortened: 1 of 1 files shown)


delivery_deliverea/models/delivery_carrier.py
```diff
@@ -558,3 +559,17 @@ def deliverea_tracking_state_update(self, picking):
             picking.date_delivered = datetime.strftime(
                 datetime.now(), DEFAULT_SERVER_DATETIME_FORMAT
             )
+
+    def _get_bulky_deliverea(self, picking, payload):
+        # here we can add the zip code and the warehouse
```
hparfr, line 564: is this comment still valid ?

Samuel-res: In the documentation, mention that can be added the warehouse and zip code 
image

hparfr: ok, I've read "it's something we do here" instead of "if we add this feature later, it should be here".

PR #922, opened 2024-12-26
[18.0][MIG] server_environment_delivery: Migration to 18.0

Kimkhoi3010:
### Change in v18:

- Since Odoo 16.0, field `tech_name` has been declared at [server.env.techname.mixin](https://github.com/OCA/server-env/blob/16.0/server_environment/models/server_env_tech_name_mixin.py), so this field is not required anymore in `server_environment_delivery`


ajaniszewska-dev (approved):
seems good

ivantodorovich (approved):

rousseldenis (approved):

PR #924, opened 2024-12-31
[16.0] [ADD] delivery_state_manual

Tisho99:
This module extends the functionality of delivery_state to allow you to manually edit the delivery state of pickings with specific delivery carriers

T-7072


Tisho99:
@ValentinVinagre @HaraldPanten @Jaimermaccione

delivery_state_manual/models/stock_picking.py
```diff
@@ -0,0 +1,26 @@
...
+
+    delivery_state = fields.Selection(
+        readonly=False,
+    )
+    delivery_state_manual = fields.Boolean(related="carrier_id.delivery_state_manual")
+
+    @api.onchange("delivery_state")
+    def _onchange_delivery_state(self):
+        for rec in self:
+            if rec.delivery_state_manual:
```
ValentinVinagre: filter before the for

Jaimermaccione (commented):
@Tisho99 Hi Alberto!

Hi Alberto!

There’s a behavior regarding the setting of the delivery date (date_delivered) that, as I see it, is currently set at the moment in the "shipping_recorded_in_carrier" and "customer_delivered" states. I think it would be more appropriate for "shipping_recorded_in_carrier" to remain hidden and for the date to be set and visible in the "warehouse_delivered" state, which is essentially another type of delivery.

Let me explain: the "shipping_recorded_in_carrier" state is meant to represent the first step, where the goods are handed over to the person responsible for transporting them to the client, whether it’s an agency or an in-house driver. The delivery date should be set when the goods have reached their destination, and I believe the two states I mentioned align better with that timing. 

If we decide to make this change, we should also update the README accordingly. Just something to keep in mind. What do you think?

delivery_state_manual/models/delivery_carrier.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2024 Alberto Martínez <alberto.martinez@sygel.es>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = "delivery.carrier"
+
+    delivery_state_manual = fields.Boolean(string="Manual Delivery State")
```
rousseldenis: Maybe a help here could be great (normal users don't have usually access to README :-))

delivery_state_manual/models/stock_picking.py
```diff
@@ -0,0 +1,25 @@
...
+import datetime
+
+from odoo import api, fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    delivery_state = fields.Selection(
+        readonly=False,
```
rousseldenis, line 13: Same as above

Tisho99:
@Jaimermaccione 

Okay, i see it well

I replaced the "shipping_recorded_in_carrier" state for the "warehouse_delivered" state in the code

delivery_state_manual/models/delivery_carrier.py
Tisho99: Done!

delivery_state_manual/models/stock_picking.py
Tisho99: Done!

Jaimermaccione (approved):
@Tisho99 

LGTM! Tested in local enviroment.

manuelregidor (approved):
Technical review. LGTM

ValentinVinagre (approved):
LGTM 👍🏻

HaraldPanten:
@etobella Do you mind merging this one?

All the comments have been considered.

THX!

HaraldPanten (approved):

PR #854, opened 2024-07-09
[16.0][IMP]delivery_delivery_package_fee delivery_automatic_package delivery_package_number simplify tests

hparfr:
Simplify tests by removing the now useless context key.

Includes:
 - [x]  #853 
 - [x]  #855  (included in 853)

Tests only pass if we have all the commit (this PR).

PR #932, opened 2025-01-15
[18.0][MIG] delivery_package_number: Migration to 18.0

CarlosRoca13:
cc @Tecnativa TT52963

ping @sergio-teruel @chienandalu


rousseldenis (approved):

delivery_package_number/models/stock_picking.py
```diff
@@ -18,11 +18,12 @@ class StockPicking(models.Model):
...
     ask_number_of_packages = fields.Boolean(compute="_compute_ask_number_of_packages")
 
-    @api.depends("package_ids")
+    @api.depends("move_line_ids", "move_line_ids.result_package_id")
     def _compute_number_of_packages(self):
         for picking in self:
-            if picking.package_ids:
-                picking.number_of_packages = len(picking.package_ids)
+            packages = picking.move_line_ids.result_package_id
+            if packages:
```
hparfr, line 25: why this `if` ?

CarlosRoca13: Because we don't want to loose the info set on `number_of_packages` when no packages are set when modifying the move_line_ids

delivery_package_number/models/stock_picking.py
```diff
@@ -0,0 +1,75 @@
...
+    _inherit = "stock.picking"
+
+    number_of_packages = fields.Integer(
+        string="Number of Packages",
+        compute="_compute_number_of_packages",
+        readonly=False,
+        store=True,
+        default=0,
+        copy=False,
+    )
```
jbaudoux, line 18: Any chance to extract this in a separate module `stock_picking_number_package` in `stock-logistics-warehouse` ?

jbaudoux (commented):
Some delivery modules depend on this field `number_of_packages` on the picking but not everyone need this wizard to manually enter a number of parcels.
In my opinion, this module deserves to be split in 2 modules.
I remember some old discussions on this matter. cc @mt-software-de 

And it would be the occasion to make it compatible with https://github.com/OCA/delivery-carrier/pull/973/files#diff-c8a077a1b4a9e53784be8d2bd8afc36a7ef4a45e87e29ba9bd5712f1927e71c7R10
cc @rousseldenis @lmignon

mt-software-de:
> Some delivery modules depend on this field `number_of_packages` on the picking but not everyone need this wizard to manually enter a number of parcels. In my opinion, this module deserves to be split in 2 modules. I remember some old discussions on this matter. cc @mt-software-de
> 
> And it would be the occasion to make it compatible with https://github.com/OCA/delivery-carrier/pull/973/files#diff-c8a077a1b4a9e53784be8d2bd8afc36a7ef4a45e87e29ba9bd5712f1927e71c7R10 cc @rousseldenis @lmignon

Yes but the discussion was on the automatic package creation. Which was a problem when running the unittests.
https://github.com/OCA/delivery-carrier/pull/854

But still i think it would be great to separate it.

sergio-teruel:
Hi, maybe other option is add an option to ask or not the number of packages...

https://github.com/OCA/delivery-carrier/blob/5331e6a846c2f9bbe35613e5d5554619b57666e6/delivery_package_number/models/stock_picking.py#L43-L50

jbaudoux (commented):
Another remark regarding module name.
Several carrier API use the term of package number to identify the "position" of the package in a list of packages. For example, when you have 3 packages and you number them 1/3, 2/3, 3/3, the package number is 1 (2 and 3 resp.) and the package total is 3.

What about renaming this module to stock_delivery_number_packages ?

delivery_package_number/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2020 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Stock Picking Package Number",
```
jbaudoux: Maybe align with module name?

pedrobaeza:
Just a remark: module names shouldn't contain plurals

delivery_package_number/__manifest__.py
CarlosRoca13: Done

CarlosRoca13:
I have added an option to prevent the wizard from popping up in the operation type.

jbaudoux:
> I have added an option to prevent the wizard from popping up in the operation type.

@CarlosRoca13 Thanks. I prefer opt-in than opt-out options. Otherwise, set opt-out default to True

CarlosRoca13:
> > I have added an option to prevent the wizard from popping up in the operation type.
> 
> @CarlosRoca13 Thanks. I prefer opt-in than opt-out options. Otherwise, set opt-out default to True

And what about the customers who already have these modules? Will they lose a feature during the migration, or will it be necessary to include a script to disable the field for existing operation types?  

From my point of view, the field should remain as it is now to avoid giving customers the impression that they are losing a feature and to prevent spending extra time creating a script that would achieve the same result.

What do you think? @pedrobaeza @sergio-teruel

pedrobaeza:
Yes, the default for the option should preserve the existing behavior. Those not wanting them should change that default option on their installations.

mt-software-de:
> > > I have added an option to prevent the wizard from popping up in the operation type.
> > 
> > 
> > @CarlosRoca13 Thanks. I prefer opt-in than opt-out options. Otherwise, set opt-out default to True
> 
> And what about the customers who already have these modules? Will they lose a feature during the migration, or will it be necessary to include a script to disable the field for existing operation types?
> 
> From my point of view, the field should remain as it is now to avoid giving customers the impression that they are losing a feature and to prevent spending extra time creating a script that would achieve the same result.
> 
> What do you think? @pedrobaeza @sergio-teruel

It should be like @jbaudoux mentioned it, if someone using it already you should create a migration for it. 
But in my opinion the best way would be to separate those features, then we wouldn't need a option to enable or disable the wizard.

jbaudoux:
> Just a remark: module names shouldn't contain plurals

`stock_delivery_number_package` then ? Is that helping to remove confusion? FYI, the planned package number module would be named `stock_quant_package_number` in stock-logistics-tracking repo

pedrobaeza:
Well, it's true that splitting into 2 modules is not something wild: `delivery_package_number` and `delivery_package_number_report` or similar. A migration script should force the installation of the second for not having troubles on the migration path.

jbaudoux:
> Well, it's true that splitting into 2 modules is not something wild: `delivery_package_number` and `delivery_package_number_report` or similar. A migration script should force the installation of the second for not having troubles on the migration path.

*(edited)*

- `stock_picking_number_package` in `stock-logistics-warehouse` for the computed field on stock.picking used by other modules
- `(stock_)delivery_number_package` in `delivery-carrier` for the wizard with the option to opt-out
- `stock_picking_report_number_package` in `stock-logistics-reporting` for the report

CarlosRoca13:
@jbaudoux @mt-software-de Sorry for the inconvenience but right now, I don’t have any more time to carry out the separation you mentioned. If splitting it into multiple modules is a crucial requirement, I’d be happy to review your contributions in the coming days. Otherwise, I think this should be merged so we can continue with our planning and consider the separation for future versions.

sergio-teruel (approved):

Issue #896, opened 2024-10-28
[18] Split base_delivery_carrier_label and drop it


florian-dacosta:
Hello,
`base_delivery_carrier_label` was meant to be a base module for shipping carrier label generation specific modules, but over the time it became quite hard to use, mainly because, it does too much stuff to be such a base module. Also, it was created before Odoo did start to manage carrier labels and has a lot less sense today.
I had already started to split it in version 16, extracting the carrier account model : https://github.com/OCA/delivery-carrier/pull/570

I believe we should continue to split the used features so we can get rid of it and maybe avoid its migration to version 18.
Here how I see this : 

1) A `delivery_carrier_option` module, with all features about option you can choose on the picking about the delivery.

2) A `stock_picking_delivery_label_link` (or another name of course) for the `shipping.label` model which allow to identify the shipping labels from other attachments, mainly for further printing.
A module is beeing implemented in v15 : https://github.com/OCA/delivery-carrier/pull/892 but with a totally different approach.
This later module add many2many fields on picking to identify the shipping labels while `base_delivery_carrier_label` adds a `shipping.label` model inheriting attachment, to allow to find the labels.
=> About this one, it would be nice to have the opinion of the users of these feature, if it would be ok to go with the many2many way, that Tecnativa team seems to support or if it is important to keep the current way of base_delivery_carrier_label.
I do prefer the `shipping.label` as for me it offers more possibility (adding field to help with the printing) but I could live with the other  way, and I think it is important to converge and avoid having 2 concurrent modules for such a small feature.

I am not sure who use this but it would be nice to have some opinion before taking action.
@hparfr @sebastienbeau @bguillot @yvaucher @hbrunn 
FYI @pedrobaeza @chienandalu 

3)The manifest wizard =>  In a new dedicated module if needed (I do not use this one and do not know someone still does.)
@ismaelcj 

4) Tracking reference fields at package level + button to check the carrier tracking url on package
As far as I know only `delivery_roulier` uses it, it may not be worth a dedicated module to just add to fields on `stock.quant.package`, unless other modules uses it
=> I will integrate it in `delivery_roulier` for now.

5) Weight features
- A new field weight on `stock.move.line` which is not filled anywhere => I guess we ignore it for now and decide when someone need it, if anyone does..
- The override of `stock.quant.package._compute_weight` computed from `stock.move.line.get_weight`
[...]

PR #935, opened 2025-01-20
[18.0][MIG] base_delivery_carrier_label: Migration to 18.0

imlopes:
Depends on: https://github.com/OCA/delivery-carrier/pull/936


florian-dacosta:
Hello @imlopes 

Please check https://github.com/OCA/delivery-carrier/issues/896
And https://github.com/OCA/delivery-carrier/pull/900

We mainly still have to extract the `shipping.label` functionalities in v18. I plan to work on this soon.

florian-dacosta:
And here is the other about the `shipping.label` model : https://github.com/OCA/delivery-carrier/pull/938

If these 2 modules are not enough to replace your usage of base_delivery_carrier_label, please tell me what more you are using and the context/usecase so we can think forward.

PR #939, opened 2025-01-24
[16][IMP] delivery_package_number: changes on delivery_package_number_document

adasatorres:
Good morning, I suggest this change to allow configuring the report with a different paper format than the default one. For example, if you need to create labels on an A4 sheet, the line break forced by ```<div style="page-break-before:always;" />```  wastes space unnecessarily.

The proposed change forces a page break if the label is cut off.

### DIFF
--- a/delivery_package_number/reports/report_package_number.xml
+++ b/delivery_package_number/reports/report_package_number.xml
@@ -11,8 +11,8 @@
                 t-value="doc.move_ids.browse(move_ids).filtered('sale_line_id')[:1].picking_id"
             />
             <t t-foreach="doc.number_of_packages" t-as="pack_num">
-                <div class="page">
-                    <div class="row">
+                <div class="page" style="page-break-inside:avoid;">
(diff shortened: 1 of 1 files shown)


christian-ramos-tecnativa (approved):
IMO this change make the module more flexible at the same time that keeps the initial behavior.

rrebollo (approved):
Code Review. LGTM!

antoniodavid:
LGTM

lauradiaz22 (approved):
Functional review and this LGTM 👍

lauradiaz22:
Good morning,

During my functional testing, although the addon works perfectly, I found a detail to improve. When extracting the report from the package label, as customer information I get the name of the address (the name of the contact). And I think it is interesting and necessary that along with the name of the contact appears the complete delivery address (city, street, postal code...). Since there are deliverymen who need this information on the boxes as well as on the delivery note.

hparfr (approved):
/ocabot merge minor

PR #965, opened 2025-02-07
[18.0][MIG] stock_picking_report_delivery_cost: Migration to 18.0

BhaveshHeliconia:


stock_picking_report_delivery_cost/models/stock_picking.py
```diff
@@ -0,0 +1,27 @@
...
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    # Needed for fields.Monetary
+    currency_id = fields.Many2one(
+        related="sale_id.currency_id",
+        readonly=True,
+        string="Currency",
+        related_sudo=True,  # for avoiding access problems
+    )
```
victoralmau: This is incorrect (an error occurs) if we try to print a picking that is not linked to a sales order (`sale_id`).

```suggestion
    currency_id = fields.Many2one(
        comodel_name="res.currency",
        string="Currency",
        compute="_compute_currency_id",
        compute_sudo=True,  # for avoiding access problems
    )
    
    @api.depends("sale_id", "sale_id.currency_id", "company_id")
    def _compute_currency_id(self):
        for item in self:
            item.currency_id = item.sale_id.currency_id or item.company_id.currency_id
```

victoralmau (changes requested):

BhaveshHeliconia:
@victoralmau You’re right—I tested it and updated the implementation as you suggested. Thanks for the helpful input!

victoralmau (approved):
Functional review OK

TT54520

victoralmau:
Although everything is correct, add the commit https://github.com/OCA/delivery-carrier/pull/1040/commits/4f9a6372fa8406bf0d9cf29868915867c68c29b5 from 16.0 before the pre-commit commit so that the change history is consistent between versions.

victoralmau (changes requested):
The `_compute_currency_id()` method appears twice because the changes from https://github.com/OCA/delivery-carrier/commit/4f9a6372fa8406bf0d9cf29868915867c68c29b5 are added after the migration commit, and the migration commit also partially adds those changes.

The solution is to add the https://github.com/OCA/delivery-carrier/commit/4f9a6372fa8406bf0d9cf29868915867c68c29b5 commit **before** the pre-commit commit.
ejemplo-github

BhaveshHeliconia:
@victoralmau : Please review and let me know if I’m still missing anything.

stock_picking_report_delivery_cost/tests/test_stock_picking_report_delivery_cost.py
```diff
@@ -52,9 +49,6 @@ def setUpClass(cls):
                 ],
             }
         )
-        cls.warehouse = cls.env["stock.warehouse"].search(
-            [("company_id", "=", cls.env.company.id)], limit=1
-        )
```
victoralmau: You should not delete this.

stock_picking_report_delivery_cost/tests/test_stock_picking_report_delivery_cost.py
```diff
@@ -79,25 +73,6 @@ def test_carrier_price_for_report_after(self):
...
-        with picking_form.move_ids_without_package.new() as line:
-            line.product_id = self.product
-            line.product_uom_qty = 1
-        picking = picking_form.save()
-        picking.carrier_price = 5
-        # report without errors
-        res = self.env["ir.actions.report"]._render(
-            "stock.report_deliveryslip", picking.ids
-        )
-        self.assertRegex(str(res[0]), picking.name)
```
victoralmau: You should not delete this.

stock_picking_report_delivery_cost/models/stock_picking.py
```diff
@@ -8,10 +8,10 @@ class StockPicking(models.Model):
 
     # Needed for fields.Monetary
     currency_id = fields.Many2one(
-        related="sale_id.currency_id",
-        readonly=True,
+        comodel_name="res.currency",
         string="Currency",
-        related_sudo=True,  # for avoiding access problems
+        compute="_compute_currency_id",
+        compute_sudo=True,  # for avoiding access problems
```
victoralmau: These changes are added in the migration commit, but they should be in the “**[FIX] stock_picking_report_delivery_cost: Fix currency_id field + Improve tests**” commit right?

victoralmau (changes requested):

BhaveshHeliconia:
@victoralmau : Now i think all fixed.

victoralmau (approved):

stock_picking_report_delivery_cost/report/report_shipping.xml
```diff
@@ -6,10 +6,10 @@
         <xpath expr="//div[@name='div_sched_date']" position="after">
             <div
                 t-if="o.carrier_price_for_report"
-                class="col-auto"
+                class="col"
                 name="div_carrier_price_for_report"
             >
-                <strong>Delivery cost:</strong>
+                <strong>Delivery cost</strong>
```
pedrobaeza, line 12: This change is making all the current translations invalid. Is it needed? If so, please adapt all the translated PO files to remove the `:` both in the source and the translation.

BhaveshHeliconia:
@pedrobaeza : I’ve updated all the translation files as suggested. Please review and let me know if the changes are correct or if anything needs adjustment.

pedrobaeza (approved):
/ocabot merge nobump

PR #938, opened 2025-01-23
[18][ADD] delivery_carrier_shipping_label : extracted from base_delivery_carrier_label

florian-dacosta:
Part of the issue https://github.com/OCA/delivery-carrier/issues/896
The purpose beeing to get rid of `base_delivery_carrier_label`

@hparfr

PR #936, opened 2025-01-20
[18.0][MIG] delivery_carrier_info: Migration to 18.0

imlopes:

PR #974, opened 2025-02-10
[18.0] [MIG] delivery_postlogistics: Migration to 18.0

StephaneMangin:
Includes: 
 - reusing of "delivery.carrier.template.option" from OCA module "delivery_carrier_option"
 - reusing of "shipping.label" from OCA module "delivery_carrier_shipping_label" instead of custom "postlogistics.shipping.label"
 - Add a default package type to fullfill `postlogistics_send_shipping` requirements (Using ECO default shipping pricing)

Depends on:
 - https://github.com/OCA/delivery-carrier/pull/900
 - https://github.com/OCA/delivery-carrier/pull/936
 - https://github.com/OCA/delivery-carrier/pull/938

Due to some modules overrding the web service and as this implementation doesn't allow to properly chain modules (that could not be chained). An API refactoring has occur to allow to override from the picking instead of the webservice (some prepare methods)


StephaneMangin:
Superseeds https://github.com/OCA/delivery-carrier/pull/940

yankinmax:
hello @StephaneMangin 
There is an issue in migration. You're overriding the selection of the `type` field declared for `delivery.carrier.template.option` model in `delivery_carrier_option` module.
Here is Odoo logs:
```
WARNING testdb odoo.fields: delivery.carrier.template.option.type: selection=[('label_layout', 'Label Layout'), ('output_format', 'Output Format'), ('resolution', 'Output Resolution'), ('basic', 'Basic Service'), ('additional', 'Additional Service'), ('delivery', 'Delivery Instructions'), ('partner_option', 'Partner Option')] overrides existing selection; use selection_add instead
```

delivery_postlogistics/views/stock_package_type_view.xml
```diff
@@ -1,21 +1,21 @@
 <?xml version="1.0" encoding="utf-8" ?>
-<!-- Copyright 2021 Camptocamp SA
-     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
```
jbaudoux: Why removing this?

delivery_postlogistics/views/stock.xml
```diff
@@ -1,36 +1,37 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<!-- Copyright 2019 Camptocamp SA
```
jbaudoux: Why removing this?

delivery_postlogistics/views/delivery.xml
```diff
@@ -1,32 +1,29 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<!-- Copyright 2019 Camptocamp SA
```
jbaudoux: Why removing this?

delivery_postlogistics/tests/fixtures/cassettes/test_token_error.yaml
```diff
@@ -15,7 +15,7 @@ interactions:
       content-type:
         - application/x-www-form-urlencoded
     method: POST
-    uri: https://wedecint.post.ch/WEDECOAuth/token
```
jbaudoux: Why this change?

delivery_postlogistics/tests/fixtures/cassettes/test_store_label.yaml
```diff
@@ -67,7 +67,7 @@ interactions:
         content-type:
           - application/json
       method: POST
-      uri: https://wedecint.post.ch/api/barcode/v1/generateAddressLabel
```
jbaudoux: Why this change?

delivery_postlogistics/tests/fixtures/cassettes/test_store_label.yaml
```diff
@@ -15,7 +15,7 @@ interactions:
         content-type:
           - application/x-www-form-urlencoded
       method: POST
-      uri: https://wedecint.post.ch/WEDECOAuth/token
```
jbaudoux: Why this change?

delivery_postlogistics/tests/fixtures/cassettes/test_missing_language.yaml
```diff
@@ -67,7 +67,7 @@ interactions:
         content-type:
           - application/json
       method: POST
-      uri: https://wedecint.post.ch/api/barcode/v1/generateAddressLabel
```
jbaudoux: Why this change?

delivery_postlogistics/tests/fixtures/cassettes/test_missing_language.yaml
```diff
@@ -15,7 +15,7 @@ interactions:
         content-type:
           - application/x-www-form-urlencoded
       method: POST
-      uri: https://wedecint.post.ch/WEDECOAuth/token
```
jbaudoux: Why this change?

delivery_postlogistics/tests/common.py
```diff
@@ -1,17 +1,17 @@
 # Copyright 2021 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
-from odoo.tests.common import TransactionCase
+from odoo.addons.base.tests.common import BaseCommon
 
 from ..postlogistics.web_service import PostlogisticsWebService
 
-ENDPOINT_URL = "https://wedecint.post.ch/"
```
jbaudoux: Why this change?

delivery_postlogistics/tests/fixtures/cassettes/test_missing_language.yaml
bizzappdev: @jbaudoux 
It has been changed because from core odoo, in test cases, it is restricting the external URL.
Agreed, that these are VCR cassettes, but Odoo is not considering them and is just blocking them. Therefore, the URL has to be internal.
FYI: https://github.com/odoo/odoo/blob/18.0/odoo/tests/common.py#L314

delivery_postlogistics/models/stock_quant_package.py
```diff
@@ -0,0 +1,66 @@
...
+
+class StockQuantPackage(models.Model):
+    _inherit = "stock.quant.package"
+
+    postlogistics_manual_cod_amount = fields.Float(
+        "PostLogistics Cash On Delivery Amount",
+        help="If the cash on delivery amount for this package is different "
+        "than the total of the sales order, write the amount there.",
+    )
+    parcel_tracking = fields.Char()
```
jbaudoux, line 15: This field should come from one of the dependencies. It's not specific to postogistics

test-requirements.txt
```diff
@@ -0,0 +1,4 @@
+vcrpy
+odoo-addon-delivery_carrier_option @ git+https://github.com/OCA/delivery-carrier.git@refs//pull/900/head#subdirectory=delivery_carrier_option
+odoo-addon-delivery_carrier_info @ git+https://github.com/OCA/delivery-carrier.git@refs//pull/936/head#subdirectory=delivery_carrier_info
+odoo-addon-delivery_carrier_shipping_label @ git+https://github.com/OCA/delivery-carrier.git@refs//pull/938/head#subdirectory=delivery_carrier_shipping_label
```
bizzappdev:
```suggestion
odoo-addon-delivery_carrier_option @ git+https://github.com/OCA/delivery-carrier.git@refs/pull/900/head#subdirectory=delivery_carrier_option
odoo-addon-delivery_carrier_info @ git+https://github.com/OCA/delivery-carrier.git@refs/pull/936/head#subdirectory=delivery_carrier_info
odoo-addon-delivery_carrier_shipping_label @ git+https://github.com/OCA/delivery-carrier.git@refs/pull/938/head#subdirectory=delivery_carrier_shipping_label
```

delivery_postlogistics/models/stock_quant_package.py
imlopes: I was extracted these fields from `base_delivery_carrier_label` here: https://github.com/OCA/delivery-carrier/pull/983/files#diff-30585d13fe1214c2c074433e2d5286e50a9e6c510a6f1caf50928ce1dcf43409

But after checking this current PR, I closed my.

delivery_postlogistics/views/stock_package_type_view.xml
StephaneMangin: I almost never see this copyrights on views, only on python code and module manifest. I put it back.

delivery_postlogistics/models/stock_quant_package.py
StephaneMangin: > I was extracted these fields from `base_delivery_carrier_label` here: https://github.com/OCA/delivery-carrier/pull/983/files#diff-30585d13fe1214c2c074433e2d5286e50a9e6c510a6f1caf50928ce1dcf43409
> 
> But after checking this current PR, I closed my.

I think it was a good idea to extract these fields at the end. Can you open back your PR please.

imlopes: Yes, for sure !!
Doing now

delivery_postlogistics/tests/fixtures/cassettes/test_missing_language.yaml
StephaneMangin: See https://github.com/OCA/delivery-carrier/pull/974#discussion_r1957833924

delivery_postlogistics/tests/fixtures/cassettes/test_store_label.yaml
StephaneMangin: See https://github.com/OCA/delivery-carrier/pull/974#discussion_r1957833924

delivery_postlogistics/tests/fixtures/cassettes/test_store_label.yaml
StephaneMangin: See https://github.com/OCA/delivery-carrier/pull/974#discussion_r1957833924

delivery_postlogistics/tests/fixtures/cassettes/test_token_error.yaml
StephaneMangin: See https://github.com/OCA/delivery-carrier/pull/974#discussion_r1957833924

delivery_postlogistics/tests/common.py
StephaneMangin: See https://github.com/OCA/delivery-carrier/pull/974#discussion_r1957833924

StephaneMangin:
> hello @StephaneMangin There is an issue in migration. You're overriding the selection of the `type` field declared for `delivery.carrier.template.option` model in `delivery_carrier_option` module. Here is Odoo logs:
> 
> ```
> WARNING testdb odoo.fields: delivery.carrier.template.option.type: selection=[('label_layout', 'Label Layout'), ('output_format', 'Output Format'), ('resolution', 'Output Resolution'), ('basic', 'Basic Service'), ('additional', 'Additional Service'), ('delivery', 'Delivery Instructions'), ('partner_option', 'Partner Option')] overrides existing selection; use selection_add instead
> ```

Yes it is transitional, because these values are added  in a dependency. Updating...

delivery_postlogistics/models/stock_picking.py
```diff
@@ -0,0 +1,256 @@
...
+            return super().attach_shipping_label(label)
+        self.ensure_one()
+        data = self.get_shipping_label_values(label)
+        context_attachment = self.env.context.copy()
+        # remove default_type setted for stock_picking
+        # as it would try to define default value of attachement
+        if "default_type" in context_attachment:
+            del context_attachment["default_type"]
+        return (
+            self.env["postlogistics.shipping.label"]
```
yankinmax:
```suggestion
            self.env["shipping.label"]
```
do I correctly understand that `self.env["postlogistics.shipping.label"]` is deprecated

StephaneMangin: Yes  it is a model coming from a dependency

delivery_postlogistics/postlogistics/web_service.py
```diff
@@ -0,0 +1,603 @@
...
+        else:
+            attributes["proClima"] = False
+
+        attributes["przl"] = services
+
+        return attributes
+
+    def _get_itemid(self, picking, pack_no):
+        """Allowed characters are alphanumeric plus `+`, `-` and `_`
+        Last `+` separates picking name and package number (if any)
```
jbaudoux: I would give in argument the package instead of the package name.
Prefixing the package name with the picking name could be a parameter.
```suggestion
    def _get_itemid(self, picking, package):
        """ID assigned by customers at address label level.
        
        Allowed characters are alphanumeric plus `+`, `-` and `_`.
        Returns the picking name and package name (if any) separated by `+`. 
```

delivery_postlogistics/postlogistics/web_service.py
```diff
@@ -0,0 +1,603 @@
...
+        if not packages:
+            attributes = self._prepare_attributes(picking)
+            add_item()
+            return item_list
+
+        pack_total = len(packages)
+        for pack in packages:
+            attributes = self._prepare_attributes(
+                picking, pack, pack_counter, pack_total
+            )
```
jbaudoux: Need a hook to collect the pack number and total from the package. By default, it would return pack_counter and pack_total

StephaneMangin: These values directly depend from the packages given in parameter which is mandatory. I don't see the usefullnesss of such a hook. Can you explain the underlying purpose ?
I refactored a bit to make it clearer.

StephaneMangin: With this light refactoring, we can see that these values are not managed outside this scope.
```
        total_packages = len(packages)
        for index, package in enumerate(packages):
            attributes = self._prepare_attributes(
                picking, package, index, total_packages
            )
            add_item(counter=index, package=package)
        else:
            attributes = self._prepare_attributes(picking)
            add_item()
        return item_list
```

jbaudoux: The stock.quant.package will have a package number defined on it (coming from another module). That value needs to be transmitted to the api instead of a local count.

StephaneMangin: Hum ok then, makes sense.

delivery_postlogistics/tests/common.py
StephaneMangin: @jbaudoux Fixed using the method described in the main PR desscription.

StephaneMangin:
@bizzappdev  Hi, you can find the solution to external URL call avoid by Odoo on the description.

delivery_postlogistics/tests/test_postlogistics.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+from odoo.exceptions import UserError
+from odoo.tests import common
+
+from .common import TestPostlogisticsCommon, recorder
+
+
+# This tag removal will allow to call external API URL
+@common.tagged("-standard")
```
bizzappdev: This tag allows all traffic to the external URL.

From my understanding, VCR will only listen when use_cassette is in context. Additionally, Odoo has explicitly blocked external URLs, likely for a valid reason. Ideally, the URL used in the cassette and the shipping method for test cases should be set to localhost.

Using localhost should not cause any issues, while bypassing Odoo's URL check could introduce a potential risk. Could you confirm if there is a specific reason for allowing an external URL in this case?
@StephaneMangin

StephaneMangin: Yes, when we want to refresh the testcase when the API changes. We want to keep the original URLs and configuration.

Here the call is not done externally, Odoo check the validity of the URL before VCR applies... Which is difficult to get around easily. For now i don't see any risks with this configuration. Later on we could find a way to manage differently this approach or if someone finds it, we would update this module accordingly.

bizzappdev: Thanks for the clarification, I agree that these particular test cases and modules do not pose a risk. However, since this is a generic module, what if someone inherits a test case class or a module method that interacts with the external URL? Wouldn't that introduce a potential risk?

Additionally, since Odoo validates the URL before VCR applies, bypassing this check could lead to unforeseen issues in the future. While there may not be immediate concerns, would it be worth exploring an approach in the future where we mock the API response while still keeping the external URL intact in the cassettes? This might help maintain Odoo’s validation integrity while ensuring test cases remain up to date.

Alternatively, you could consider having separate test cases specifically for API changes and tagging them with a `-standard` tag, which would serve your purpose while also ensuring test cases remain risk-free and enhance the module’s extensibility.

StephaneMangin: UPDATE: this method doesn't work at all... I need to override the check directly to launch the tests (otherwize it's all green ^^ but false positive) I updated the PR

delivery_postlogistics/models/delivery_carrier.py
```diff
@@ -98,6 +93,14 @@ class DeliveryCarrier(models.Model):
         string="ZPL Patch String", default="^XA^CW0,E:TT0003M_.TTF^XZ^XA^CI28"
     )
 
+    def default_get(self, fields_list):
+        res = super().default_get(fields_list)
+        res["postlogistics_default_package_type_id"] = self.env.ref(
+            "delivery_postlogistics.postlogistics_default_package_type",
+            raise_if_not_found=False,
+        ).id
```
leemannd: The raise_if_not_found will raise a Nonetype if the xmlid is not found

test-requirements.txt
```diff
@@ -1 +1,2 @@
-odoo-test-helper
+vcrpy
+odoo_test_helper
```
vvrossem:
```suggestion
odoo_test_helper
vcrpy
```

delivery_postlogistics/README.rst
```diff
@@ -101,31 +100,33 @@ Authors
...
--  Jacques-Etienne Baudoux <je@bcim.be>
+- Jacques-Etienne Baudoux <je@bcim.be>
 
--  `Trobz <https://trobz.com>`__:
+- `Trobz <https://trobz.com>`__:
 
-   -  Jack Le <anlh@trobz.com>
+  - Jack Le <anlh@trobz.com>
+
+- Stéphane Mangin stephane.mangin@camptocamp.com
```
vvrossem, line 121:
```suggestion
- Stéphane Mangin <stephane.mangin@camptocamp.com>
```
You can regenerate the README.rst

delivery_postlogistics/models/delivery_carrier.py
```diff
@@ -111,14 +117,22 @@ def onchange_prod_environment(self):
...
                 carrier.postlogistics_endpoint_url = "https://wedecint.post.ch/"
 
+    def _postlogistics_get_default_custom_package_code(self):
+        # Used while changing the carrier on the stock.package.type Form
+        package_type = self.postlogistics_default_package_type_id
+        if package_type:
+            return ",".join(package_type._get_shipper_package_code_list())
+        # Ultimate fallback
+        return "ECO"
+
```
vvrossem:
```suggestion
    def _postlogistics_get_default_custom_package_code(self):
        # Used while changing the carrier on the stock.package.type Form
        if (package_type := self.postlogistics_default_package_type_id):
            return ",".join(package_type._get_shipper_package_code_list())
        # Ultimate fallback
        return "ECO"
```

delivery_postlogistics/models/stock_picking.py
```diff
@@ -44,8 +44,7 @@ def _get_packages_from_picking(self):
             # moved so take the source one.
             package_ids.add(operation.result_package_id.id or operation.package_id.id)
 
-        packages = self.env["stock.quant.package"].browse(package_ids)
-        return packages
+        return self.env["stock.quant.package"].browse(package_ids)
 
     def get_shipping_label_values(self, label):
         # TODO: consider to depends on base_delivery_carrier_label
```
vvrossem, line 65: is this TODO still relevant?

delivery_postlogistics/postlogistics/web_service.py
```diff
@@ -14,7 +14,7 @@
 import requests
 from PIL import Image
 
-from odoo import _, exceptions
+from odoo import _
```
vvrossem:
```suggestion
```
This can be removed since `*.env._` is used instead

vvrossem (approved):

delivery_postlogistics/models/stock_picking.py
StephaneMangin: Yes but the latter module isn't ready yet

delivery_postlogistics/postlogistics/web_service.py
StephaneMangin: This is the webservice implementation which does not inherits from Models ^^

delivery_postlogistics/postlogistics/web_service.py
```diff
@@ -0,0 +1,487 @@
...
+
+        """
+        results = []
+        picking_carrier = picking.carrier_id
+        access_token = self.get_access_token(picking_carrier)
+
+        # get options
+        lang = self._get_language(picking.partner_id.lang)
+        post_customer = self._prepare_customer(picking)
+        recipient = self._prepare_recipient(picking)
```
bizzappdev, line 349:
```suggestion
        recipient = picking.postlogistics_label_prepare_recipient()
```

bizzappdev (changes requested):
This will remove the DeprecationWarning and we will no longer need to add `WARNING.* DeprecationWarning` in ignore.

delivery_postlogistics/postlogistics/web_service.py
StephaneMangin: I would have kept the retrocompatibility of th API (and remove the old API for a latter major update, 18.0.2.0.0 for instance). Otherwize, the deprecation warning is not usefull though.

StephaneMangin: I realize my comments are not as precise as they should be.

yankinmax (approved):

leemannd (commented):
Are the changes on `stock_picking_deliver_link` intended?

leemannd:
Fixing the pre-commit in -> https://github.com/OCA/delivery-carrier/pull/1004

bizzappdev (approved):

delivery_postlogistics/migrations/18.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,63 @@
...
+        )
+    ]
+    openupgrade.rename_xmlids(cr, xmlids_spec)
+
+
+def _rename_models(cr):
+    models_spec = [
+        (
+            "postlogistics.delivery.carrier.template.option",
+            "delivery.carrier.template.option",
```
leemannd: I suspect this will not work.
1. The model is already created by the module `delivery_carrier_option`

```
2025-03-25 16:20:21,900 98 ERROR prod odoo.sql_db: bad query: b"UPDATE ir_model SET model = 'delivery.carrier.template.option' WHERE model = 'postlogistics.delivery.carrier.template.option'"
ERROR: duplicate key value violates unique constraint "ir_model_obj_name_uniq"
DETAIL:  Key (model)=(delivery.carrier.template.option) already exists.
 
2025-03-25 16:20:21,901 98 ERROR prod OpenUpgrade: Error after 0:00:00.001743 running UPDATE ir_model SET model = 'delivery.carrier.template.option' WHERE model = 'postlogistics.delivery.carrier.template.option' 
2025-03-25 16:20:21,977 98 WARNING prod odoo.modules.loading: Transient module states were reset 
2025-03-25 16:20:21,980 98 ERROR prod odoo.modules.registry: Failed to load registry 
2025-03-25 16:20:21,980 98 CRITICAL prod odoo.service.server: Failed to initialize database `prod`. 
  File "/odoo/src/odoo/service/server.py", line 1318, in preload_registries
    registry = Registry.new(dbname, update_module=update_module)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/decorator.py", line 232, in fun
    return caller(func, *(extras + args), **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/odoo/src/odoo/tools/func.py", line 97, in locked
    return func(inst, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/odoo/src/odoo/modules/registry.py", line 127, in new
    odoo.modules.load_modules(registry, force_demo, status, update_module)
  File "/odoo/src/odoo/modules/loading.py", line 476, in load_modules
    processed_modules += load_marked_modules(env, graph,
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/odoo/src/odoo/modules/loading.py", line 364, in load_marked_modules
    loaded, processed = load_module_graph(
                        ^^^^^^^^^^^^^^^^^^
  File "/odoo/src/odoo/modules/loading.py", line 181, in load_module_graph
    migrations.migrate_module(package, 'pre')
  File "/odoo/src/odoo/modules/migration.py", line 254, in migrate_module
    mod.migrate(self.cr, installed_version)
  File "/odoo/external-src/delivery-carrier/delivery_postlogistics/migrations/18.0.1.0.0/pre-migration.py", line 61, in migrate
    _rename_models(cr)
  File "/odoo/external-src/delivery-carrier/delivery_postlogistics/migrations/18.0.1.0.0/pre-migration.py", line 40, in _rename_models
    openupgrade.rename_models(cr, models_spec)
  File "/usr/local/lib/python3.12/site-packages/openupgradelib/openupgrade.py", line 922, in rename_models
    logged_query(
  File "/usr/local/lib/python3.12/site-packages/openupgradelib/openupgrade.py", line 1730, in logged_query
    cr.execute(query, args)
  File "/odoo/src/odoo/sql_db.py", line 354, in execute
    res = self._obj.execute(query, params)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "ir_model_obj_name_uniq"
DETAIL:  Key (model)=(delivery.carrier.template.option) already exists.
```

AFAIR best is:
- migrate the records from previous model to the new one
- use `merge_models` -> https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L1135 (I never used it)

delivery_postlogistics/models/stock_picking.py
```diff
@@ -0,0 +1,458 @@
...
+        context_attachment = self.env.context.copy()
+        # remove default_type setted for stock_picking
+        # as it would try to define default value of attachement
+        if "default_type" in context_attachment:
+            del context_attachment["default_type"]
+        return (
+            self.env["shipping.label"].with_context(**context_attachment).create(data)
+        )
+
+    def postlogistics_cod_amount(self):
```
yankinmax, line 90: here the `_set_a_default_package` is missing
https://github.com/OCA/delivery-carrier/blob/5a4f7c4eb5f7bd2c6bd0298d69937f0199270c02/delivery_postlogistics/models/stock_picking.py#L78

ivantodorovich (approved):
thanks!

delivery_postlogistics/models/stock_picking.py
yankinmax: I see it's removed now, so nevermind

jbaudoux (approved):

delivery_postlogistics/migrations/18.0.1.0.0/pre-migration.py
jbaudoux: @leemannd Is this issue still valid or is it resolved ?

jbaudoux:
@leemannd Hello. Can you update your review?

delivery_postlogistics/__manifest__.py
```diff
@@ -0,0 +1,41 @@
...
+        "views/res_partner_view.xml",
+        "views/stock.xml",
+    ],
+    "installable": True,
+    "auto_install": False,
+    "application": True,
+    "external_dependencies": {
+        "python": [
+            "openupgradelib",
+            "vcrpy",
```
simahawk: @yankinmax vcrpy IS NOT a dep. It's needed only for tests.
Also, please try to respect the commit msg guidelines.
I'm fixing these issues now and merge. I'm also dropping the change on the index.html file. 
TIP for devs: `export SKIP=oca-gen-addon-readme` to disable pre-commit readme update.

@leemannd commit msg issue applies to you as well :wink:

simahawk:
I'm also removing the fix for test dep because it came from the other PR for postlogistics. https://github.com/OCA/delivery-carrier/pull/974/commits/10676c6f18741d528ab0ff260878c7540e865f42

delivery_postlogistics/__manifest__.py
yankinmax: Hello @simahawk, I've added vcrpy, because another PR build was failing without this dependency.
I'll check the commit message guideline. I thought I was respecting it.
Thanks for your help on this.

simahawk:
Sounds like there's an issue w/ a PO file. @leemannd or @bizzappdev (as you are working on #941 ) would you be able to check what's wrong?

bizzappdev:
@simahawk yes sure we will check soon

PR #942, opened 2025-01-30
[18.0][MIG] partner_delivery_schedule: Migration to 18.0

juancarlosonate-tecnativa:
Migration to v18: name_get is no longer valid and has been replaced by display_name. The bootstrap badges in the reports rendered to pdf do not work, so i removed them.

@Tecnativa TT54201
@CarlosRoca13 @sergio-teruel can you check? thanks!

PR #979, opened 2025-02-12
[18.0][FIX] partner_delivery_schedule: Incorrect module dependency

juancarlosonate-tecnativa:
@Tecnativa

There is an issue in the migration https://github.com/OCA/delivery-carrier/pull/942 i have changed the module dependency so that it can be installed correctly.

@sergio-teruel @CarlosRoca13  take a look when you can, please

### DIFF
--- a/partner_delivery_schedule/__manifest__.py
+++ b/partner_delivery_schedule/__manifest__.py
@@ -11,7 +11,7 @@
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "depends": ["delivery"],
+    "depends": ["stock_delivery"],
     "data": [
         "security/ir.model.access.csv",
         "views/partner_delivery_schedule_view.xml",


CarlosRoca13:
It wasn't changed in the migration... It has had `delivery` since its creation... Can you point out the problem with the installation?

juancarlosonate-tecnativa:
> It wasn't changed in the migration... It has had `delivery` since its creation... Can you point out the problem with the installation?

Delivery no longer depends on sale_stock, so it can't find stock and some of the external xml ids are missing. https://github.com/odoo/odoo/blob/6d13c9a7ca0ba9940beeb0ac07499e48a926ffa0/addons/delivery/__manifest__.py#L14

If you test this module on runboat in version 18 without having installed any other stock modules beforehand, you'll see the issue, it was overlooked during the migration.

CarlosRoca13 (approved):
Thank you very much for the explanation :smiley:

sergio-teruel (approved):
Great!!

pedrobaeza (approved):
/ocabot merge patch

PR #990, opened 2025-02-20
[17.0][MIG] delivery_carrier_multi_zip: Migration to 17.0

victoralmau:
Migration to 17.0

Please @pedrobaeza can you review it?

@Tecnativa TT52409


victoralmau:
IMO you should remove the `post_init_hook` method and try to convert the prefixes (`zip_prefix_ids`) to `delivery.carrier.zip` values because it is not possible. What do you think @pedrobaeza ?

pedrobaeza:
The hook doesn't make sense now in this context, and I don't think we should do any kind of conversion of the prefixes. Add a new selection option for choosing if this is going to be determined by prefixes or ZIP ranges, with the default in prefixes, and that's all.

delivery_carrier_multi_zip/models/delivery_carrier.py
```diff
@@ -0,0 +1,50 @@
...
+    zip_range_ids = fields.One2many(
+        comodel_name="delivery.carrier.zip",
+        inverse_name="carrier_id",
+        string="ZIP codes",
+    )
+
+    def _match_address(self, partner):
+        """Match as well by zip intervals if they are present."""
+        res = super()._match_address(partner)  # it has self.ensure_one()
+        if res and self.zip_range_ids:
```
pedrobaeza: Do this only if the option is selected, not depending on the entries (as they may exist, but the option has been changed).

pedrobaeza (commented):
I think you don't need to depend on `stock_delivery`, or the method `_match_address` is there?

victoralmau:
The `_match_address()` method is in `delivery`, but the `ir.model.access` records that are set for `delivery.carrier.zip` are linked to `stock` (stock.group_stock_user and stock.group_stock_manager), so the `stock_delivery` dependency is necessary. 

This same (or similar) explanation applies to other _delivery_carrier_ modules in this repository in 17.0.

pedrobaeza (approved):

rousseldenis (commented):
@victoralmau It seems this misses DESCRIPTION.rst as section is removed from README

victoralmau:
> @victoralmau It seems this misses DESCRIPTION.rst as section is removed from README

I'm sorry but I don't understand what you are saying, the readme/DESCRIPTION.md file exists (as do others) and the README.rst file is automatically generated with the contents of the readme folder.

PR #986, opened 2025-02-17
 [15.0][MIG] delivery_ups_oca: Migration to 15.0 

jans23:

PR #989, opened 2025-02-20
[17.0] [MIG] delivery_ups_oca

Tisho99:
Standard migration

T-7548

PR #991, opened 2025-02-20
15.0 mig delivery_ups_oca

jans23:
Successor of #986.


delivery_ups_oca/models/product_packaging.py
```diff
@@ -6,4 +6,9 @@
 class ProductPackaging(models.Model):
     _inherit = "product.packaging"
 
-    package_carrier_type = fields.Selection(selection_add=[("ups", "UPS")])
+    package_carrier_type = fields.Selection(
+        selection=[("none", "None"), ("ups", "UPS")],
+        string="Carrier",
+        default="none",
+    )
+    shipper_package_code = fields.Char()
```
Tisho99: What is this new field used for?

delivery_ups_oca/models/product_packaging.py
```diff
@@ -6,4 +6,9 @@
 class ProductPackaging(models.Model):
     _inherit = "product.packaging"
 
-    package_carrier_type = fields.Selection(selection_add=[("ups", "UPS")])
+    package_carrier_type = fields.Selection(
```
Tisho99: I think this should't be changed, you are overriding the selection values and can cause conflicts with other modules that use selection_add

delivery_ups_oca/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,16 @@
+To configure this module, you need to:
+
+#. Add a carrier account with delivery type ``ups`` and fill in your credentials (UPS
```
Tisho99: Isn't 'delivery type' instead of 'carrier account' the model?

jans23: This describes the configuration in the UI where it is actually called Shipping Method and Delivery Method (inconsistently) with Provider "UPS". I changed it accordingly.

delivery_ups_oca/models/product_packaging.py
jans23: I removed it.

delivery_ups_oca/models/product_packaging.py
jans23: Right.

Tisho99: You changed it and then put it back the way it was

delivery_ups_oca/models/product_packaging.py
Tisho99: Mmm the tests now fail, I found that the field was moved to stock.package.type model, but i still don't know why.

P.S.: I am doing a migration of this module to V17, but is still in progress: https://github.com/OCA/delivery-carrier/pull/989

delivery_ups_oca/models/product_packaging.py
Tisho99: Shouldn't this file be deleted if the data and the new package_carrier_type and the data has been moved to the stock.package.type model?

delivery_ups_oca/models/ups_request.py
```diff
@@ -121,7 +121,7 @@ def _quant_package_data_from_picking(self, package, picking, is_package=False):
             },
             "Dimensions": {
                 "UnitOfMeasurement": {"Code": self.package_dimension_code},
-                "Length": str(package.packaging_length),
+                "Length": str(package.length),
```
Tisho99: package is a record of the "stock.package.type" (see line 114), I think packaging_length was the correct field (see image)

image

delivery_ups_oca/models/ups_request.py
```diff
@@ -268,7 +268,7 @@ def _quant_package_data_from_order(self, order):
             "PackagingType": {"Code": self.default_packaging_id.shipper_package_code},
             "Dimensions": {
                 "UnitOfMeasurement": {"Code": self.package_dimension_code},
-                "Length": str(self.default_packaging_id.packaging_length),
+                "Length": str(self.default_packaging_id.length),
```
Tisho99: Same

Tisho99 (approved):
LGTM Code review

PR #993, opened 2025-02-25
[16.0][FIX] delivery_auto_refresh: don't try to create lines over a new id

chienandalu:
Some other modules could trigger the autorefresh even when the order record isn't in the db, and making the line creation fail. We only need it when the record is saved.

cc @Tecnativa TT55206

please review @pilarvargas-tecnativa @pedrobaeza


pilarvargas-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1003, opened 2025-03-24
[18][MIG] delivery_roulier

florian-dacosta:
No major changes, but I did add the tracking ref fields on `stock.quant.package` because of the removal of base_delivery_carrier_label from the dependencies, like explained here : https://github.com/OCA/delivery-carrier/issues/896

This has not been tested with a real carrier implementation yet.


rvalyi (commented):
in decorator.py you used the "fun" variable name. I think "meth" or "func" would be more idiomatic.

florian-dacosta:
> in decorator.py you used the "fun" variable name. I think "meth" or "func" would be more idiomatic.

Ok, I did change it for "method", as func was used already.

delivery_roulier/tests/test_delivery_roulier.py
```diff
@@ -0,0 +1,148 @@
...
+                "type": "zpl2",
+            },
+            "id": 1,
+        }
+    ],
+    "annexes": [{"name": "annexe name", "type": "txt", "data": b"dGVzdCBhbm5leGU="}],
+}
+
+
+class DeliveryRoulierCase(TransactionCase):
```
rousseldenis: FYI, you can use `BaseCommon` Odoo base class to avoid tracking disable line.

rousseldenis (approved):
Could you remove stock_picking_delivery_link changes here ? Thanks

florian-dacosta:
> Could you remove stock_picking_delivery_link changes here ? Thanks

I am not sure how it got there, my mistake, I have removed it.

rvalyi (approved):

PR #1012, opened 2025-04-30
[18.0][MIG] delivery_price_method: Migration to 18.0

pablo-cort-s73:
Migration to 18.0


BT-dmoreno (approved):

BT-tkarpinski (approved):

delivery_price_method/__manifest__.py
```diff
@@ -10,7 +10,7 @@
     "license": "AGPL-3",
     "development_status": "Production/Stable",
     "category": "Delivery",
-    "version": "17.0.1.0.0",
-    "depends": ["delivery"],
+    "version": "18.0.1.0.0",
+    "depends": ["stock_delivery"],
```
rousseldenis, line 14: @pablo-cort-s73 Is that change really needed as it seems to me we can only rely on `delivery` module for this one purpose? Correct ?

BT-dmoreno: @rousseldenis since version 17.0 it's the new module `stock_delivery` who implements the original method `send_shipping`. Since this module is extending that same method, it makes sense for me that the dependency is updated to be using `stock_delivery` instead of just `delivery`.

https://github.com/odoo/odoo/blob/17.0/addons/stock_delivery/models/delivery_carrier.py#L35

It should have been also the case for this module in version 17.0 actually.

pablo-cort-s73: That's right @rousseldenis, sorry for not answering right away, that's the reason for the change in the dependency.

Thanks @BT-dmoreno

BT-dmoreno:
@OCA/logistics-maintainers looks like everything should be fine here. Could you please merge this PR?

hparfr (approved):
/ocabot merge nobump

Issue #1009, opened 2025-04-16
[17.0][FIX] delivery_package_number: recomputation of number_of_packages on manually edits


Tisho99:
The number_of_packages field for a pick can be modified using a wizard or directly in the form view, but the second way does not work because the field is recomputed before.

I am willing to do the FIX in a PR, but i would like to read someone's opinion before

## Module
delivery_package_number

## Describe the bug
The number_of_packages field of the stock.picking in the delivery_package_number module can be modified directly in the form view, but when the picking is confirmed the field is recomputed, and the value of number_of_packages changes.

1. Go to Inventory / Operations / Deliveries.
2. Create a delivery. Set a carrier. Set the number_of_packages field greater to 1.
3. Validate the picking. The number_of_packages would have been recomputed

**Additional context**
If you don't edit the value of number_of_packages in the form, a wizard to edit the field will be displayed on the confirmation of the picking, and then the new value of number_of_packages won't be recomputed. 

**Expected behavior**
Not recompute number_of_packages field if it has been modified also in the form view.

There are 2 ways of fixing it: 

1: Passing the following [context](https://github.com/OCA/delivery-carrier/blob/a481eb6f0d1fbca4531925c112e0a0bd2515553d/delivery_package_number/wizard/stock_number_package_validate_wiz.py#L46-L48) to the picking [confirmation](https://github.com/OCA/delivery-carrier/blob/a481eb6f0d1fbca4531925c112e0a0bd2515553d/delivery_package_number/models/stock_picking.py#L59) of the module.

2: Adding an option in the base_delivery_carrier_label to not automatically create the packages, which is the origin of this error, and what is skipped with the context of the option 1 ([link](https://github.com/OCA/delivery-carrier/blob/a481eb6f0d1fbca4531925c112e0a0bd2515553d/base_delivery_carrier_label/models/stock_picking.py#L72-L75)) 

**Images of the error:**

![Image](https://github.com/user-attachments/assets/f40a18f4-270f-4f0d-8782-df45cc194841)

![Image](https://github.com/user-attachments/assets/fadab40f-8db8-46ea-bb02-cd1d738cdb4d)

Tisho99:
@manuelregidor 

T-8186

Tisho99:
Hello @pedrobaeza 
We are trying to do a PR in the delivery_package_number module to be able to manually edit the number_of_packages field directly in the form view, and I've described the idea

[...]

PR #853, opened 2024-07-09
[16.0][BC] base_delivery_carrier_label: remove _set_a_default_package

hparfr:
This functionality adds a burden on other modules tests like (delivery_package_fee or delivery_package_number).

In some contexts, you want to require a package to be always explicitly created.

This functionality is also now implemented in `delivery_automatic_package` module.

To be merged with `major`

EDIT:

Contains and replace:
- [ ] #855

~~This PR remove set_a_default_package from base_delivery_carrier_label.
Tests are red because it needs some changes in delivery_roulier  #855 
And tests in #855 are also red because it needs this PR.~~

PR #1013, opened 2025-05-06
[17.0] [IMP] base_delivery_carrier_label: add setting to not set default packages in shipper send

Tisho99:
This PR adds a configuration option to not set the default packages when sending a picking to the carrier

The functionality has been removed in v16: https://github.com/OCA/delivery-carrier/pull/853/files

Related with the https://github.com/OCA/delivery-carrier/issues/1009 Issue

T-8186

### DIFF
--- a/base_delivery_carrier_label/README.rst
+++ b/base_delivery_carrier_label/README.rst
@@ -39,6 +39,23 @@ module for other carrier-specific modules.
 .. contents::
    :local:
 
+Configuration
+=============
+
+By default, automatic packaging is enabled.
+
+You can disable it in Inventory / Configuration / Settings, by searching
+"Default Packages".
+
+When active, pickings without assigned packages will be grouped into a
+single one for shipping. This ensures compatibility with connectors that
+require at least one package to generate labels or tracking references.
+
+   **NOTE:** If the Number of Packages field is manually set but no real
+   packages are defined, this feature overrides it and creates only one
+   package. This may cause confusion when relying solely on that field
+   instead of using Odoo’s standard package objects.
+
 Usage
 =====
 
--- a/base_delivery_carrier_label/models/__init__.py
+++ b/base_delivery_carrier_label/models/__init__.py
@@ -2,6 +2,8 @@ from . import delivery_carrier
 from . import delivery_carrier_template_option
 from . import delivery_carrier_option
 from . import stock_move_line
+from . import res_company
+from . import res_config_settings
 from . import stock_picking
 from . import stock_quant_package
 from . import shipping_label
--- a/base_delivery_carrier_label/models/res_company.py
+++ b/base_delivery_carrier_label/models/res_company.py
@@ -1,0 +1,10 @@
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    delivery_set_default_package = fields.Boolean(
+        string="Default Packages",
+        default=True,
+    )
--- a/base_delivery_carrier_label/models/res_config_settings.py
+++ b/base_delivery_carrier_label/models/res_config_settings.py
@@ -1,0 +1,10 @@
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    delivery_set_default_package = fields.Boolean(
+        related="company_id.delivery_set_default_package",
+        readonly=False,
+    )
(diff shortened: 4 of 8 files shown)


Jaimermaccione (changes requested):
@Tisho99 Reviewed the module improvement, and it works correctly in the local environment.  
To make the behavior clearer to end users, I suggest improving the README and the field description. Let me know what you think, in case there's any approach that might be off. Thanks! :

---

##  Configuration

By default, the automatic packaging behavior is **enabled**.

You can disable it from  
**Inventory → Configuration → Settings**, by searching for **"Default Packages"**.

When enabled, if a picking has no assigned packages, Odoo will automatically group all move lines into a **single package** before sending it to the carrier.  
This default behavior ensures that shipping connectors can work properly, as many of them require at least one package to generate labels or tracking numbers.

⚠️ **Note**:  
If a user manually sets a number of packages (e.g., `3`) via the *Number of Packages* field (see [delivery_package_number](https://github.com/OCA/delivery-carrier/tree/17.0/delivery_package_number)) but no actual packages exist yet, this setting will **silently override** it and assign **only one physical package**.

This behavior can be especially confusing when the standard Odoo packaging feature (`stock.quant.package`) is not used.  
In such cases, users rely solely on the *Number of Packages* field to inform the carrier.  
Automatically generating one default package at validation can cause the system to ignore the user-defined value, resulting in only one package being communicated to the carrier, regardless of the original input.

This setting was introduced to avoid these inconsistencies and ensure better control over when and how packages are assigned for shipping.

---

## Field title and help text (revised)

### **Title**  
**Automatically assign a single package before shipping**  
_When sending a picking to the carrier, this option ensures all items are grouped into one package if none exist. This helps shipping connectors that require at least one package for label generation or tracking._

### **Help**  
**If no packages are assigned, the system will group all items into a single package before shipping.**

Tisho99:
@Jaimermaccione 

What you say makes sense and its better documented

However, I think that is too extense, this module is big and its documentation is short. If we put that, half of the module's readme will talk about a small secondary feature and I don't think is appropriate

Jaimermaccione:
@Tisho99 Thanks for the feedback, you’re absolutely right.

I’ve rewritten it to keep it concise while still clarifying the purpose and the edge case that can confuse users. Here’s the new proposed text:

---

### Configuration

By default, automatic packaging is enabled.

You can disable it in **Inventory → Configuration → Settings**, by searching *"Default Packages"*.

When active, pickings without assigned packages will be grouped into a single one for shipping. This ensures compatibility with connectors that require at least one package to generate labels or tracking references.

> ⚠️ **Note:**  
> If the *Number of Packages* field is manually set but no real packages are defined, this feature overrides it and creates only one package.  
> This may cause confusion when relying solely on that field instead of using Odoo’s standard package objects.

---

If you think even this is too long, we could drop the note and just keep a shorter version —

---

### Configuration

By default, automatic packaging is enabled.

You can disable it in **Inventory → Configuration → Settings**, by searching *"Default Packages"*.

When active, pickings without assigned packages will be grouped into a single one for shipping. This ensures compatibility with connectors that require at least one package to generate labels or tracking references.

---


Pick whichever version fits best — short or extended. Happy to go with what works for the README. Thank you!

Tisho99:
@Jaimermaccione I have updated the readme with the first option

Jaimermaccione (changes requested):
@Tisho99 Perfect, thank you very much!

Regarding the delivery_set_default_package field — do you think it’s better to leave the help text and title as they are, rather than updating them to something like the suggestion I shared?

Currently, it says:

Help: "Set Default Packages at Pickings Before Sending it to Shipper"
Title: "Shipping connectors allow to compute accurate shipping costs, print shipping labels and request carrier picking at your warehouse to ship to the customer. Apply shipping connector from delivery methods."

I was thinking of something like this instead:

Title
Automatically assign a single package before shipping
When sending a picking to the carrier, this option ensures all items are grouped into one package if none exist. This helps shipping connectors that require at least one package for label generation or tracking.

Help
If no packages are assigned, the system will group all items into a single package before shipping.

Let me know what you think. Thanks again!

Tisho99:
@Jaimermaccione 

Okay, I've replaced my field description with yours. However, I've moved some of the help to the title to shorten the main description.

Here you can see how it is displayed:

Captura desde 2025-05-13 10-13-17

Do you see it correct now?

Jaimermaccione (approved):
@Tisho99 Looks great! Thank you! LGTM!

manuelregidor (approved):
Technical review. LGTM

Tisho99:

Tisho99:
Hi @pedrobaeza 

Could you have a look?

This PR implements the suggestion Carlos Roca did in the https://github.com/OCA/delivery-carrier/issues/1009 issue

In v16 a similar PR has been merged: https://github.com/OCA/delivery-carrier/pull/853

Thank you

ValentinVinagre (approved):
LGTM 👍🏻

PR #1015, opened 2025-05-08
[17.0] [ADD] delivery_carrier_global_manifest

Tisho99:
This new module adds a general manifest file that can be printed with all carriers.

The base_delivery_carrier_label adds a wizard to print the manifest file, but no manifest can be printed with that module. Only a few carrier modules extend that base_delivery_carrier_label to allow to print a manifest, but only for the added carrier.

An option to print the manifest file for more specific pickings, or just one, could also be added

T-8167


Jaimermaccione (changes requested):
@Tisho99 Functionally reviewed in a local environment. Everything works as expected and the README looks good.

Just one comment regarding the wizard's switch: it might be a bit confusing. While the is_general_manifest_query field correctly toggles depending on whether a carrier is set or not, it's still editable by the user even when a carrier is selected. This can be misleading, especially since generating the file always uses the carrier-specific logic in that case.

image


If the switch is meant to be informative only, I’d recommend making it read-only in all cases to avoid any ambiguity and make it clear that the logic is entirely based on the carrier_id.

Let me know what you think. Thanks!

Tisho99:
@Jaimermaccione I have made the switch readonly. 

Do you see it correct now?

In reality, the field is not necessary and could even be eliminated

Jaimermaccione (approved):
@Tisho99 LGTM! Functional review on local environment.

delivery_carrier_general_manifest/wizards/manifest_wizard.py
```diff
@@ -0,0 +1,91 @@
...
+        return general_carriers
+
+    def _get_general_manifest_pickings(self):
+        self.ensure_one()
+        search_args = [
+            ("carrier_id", "in", self._get_general_manifest_carrier().ids),
+            ("date_done", ">=", self.from_date),
+            ("state", "=", "done"),
+            ("picking_type_code", "=", "outgoing"),
+            # ("carrier_tracking_ref", "!=", False),
```
manuelregidor: Could you remove this line?

delivery_carrier_general_manifest/wizards/manifest_wizard_view.xml
```diff
@@ -0,0 +1,20 @@
...
+    <record id="manifest_wizard_form_inherit_general" model="ir.ui.view">
+        <field name="name">manifest.wizard.form.inherit.general</field>
+        <field name="model">manifest.wizard</field>
+        <field
+            name="inherit_id"
+            ref="base_delivery_carrier_label.manifest_wizard_form"
+        />
+        <field name="arch" type="xml">
+            <field name="carrier_id" position="after">
+                <field name="is_general_manifest_query" widget="boolean_toggle" />
```
manuelregidor: Could you add a message to inform that if no carriers are selected, a general manifest will be generated using all carriers included in the general manifest?

Tisho99: done

manuelregidor:
@Tisho99 I can't see the changes : s

Tisho99:
@manuelregidor I added a help, and now i have removed the line you requested

manuelregidor:
@Tisho99 Please, could you include a message in the view, not a help, so it is clearly visible when the manifiest that is going to be generated is the general one? Thank you.

Tisho99:
@manuelregidor I have included the message directly in the view

manuelregidor (approved):
Tecnical review. LGTM

Tisho99:

delivery_carrier_general_manifest/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2025 Alberto Martínez <alberto.martinez@sygel.es>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Delivery Carrier General Manifest",
+    "summary": "Manifest files for all carriers",
+    "version": "17.0.1.0.0",
+    "category": "Delivery",
+    "website": "https://github.com/OCA/delivery-carrier",
+    "author": "Sygel, Odoo Community Association (OCA)",
+    "maintainers": ["tisho99"],
```
ValentinVinagre: are u sure about that?

ValentinVinagre (changes requested):
Has it been reviewed that this functionality is not already supported by existing carrier modules?
For example: https://github.com/OCA/l10n-spain/tree/17.0/delivery_gls_asm#manifiesto-de-envios
 @manuelregidor

delivery_carrier_general_manifest/__manifest__.py
Tisho99: No, i'll delete it if you want

ValentinVinagre: It's no problem for me. Do you know what it means to be there?

Tisho99: I know something, like I can merge PRs from this module. Is there any documentation of the maintainer duties?

ValentinVinagre: As far as I know, there's no documentation for that part. In short, it's this: you agree to be the maintainer of this module, meaning you'll be automatically pinged if someone opens a PR for this module.

manuelregidor:
@ValentinVinagre The example you mention does have the manifest functionality implemented.

ValentinVinagre:
> @ValentinVinagre The example you mention does have the manifest functionality implemented.

Yes, won't it be duplicate functionality with this module?

manuelregidor:
> > @ValentinVinagre The example you mention does have the manifest functionality implemented.
> 
> Yes, won't it be duplicate functionality with this module?

With this module you can generate individual manifests or a general one if needed. You can also exclude carriers in case you don't need to include them in the general manifest, so I don't see the problem.

ValentinVinagre:
> > > @ValentinVinagre The example you mention does have the manifest functionality implemented.
> > 
> > 
> > Yes, won't it be duplicate functionality with this module?
> 
> With this module you can generate individual manifests or a general one if needed. You can also exclude carriers in case you don't need to include them in the general manifest, so I don't see the problem.

In my opinion, it duplicates the functionality of the carrier manifest. Ideally, this generic functionality would allow for carrier-specific configurations, and the rest of the carrier modules would need to be reviewed to see if they also offer this functionality and a general change made. Otherwise, it involves more maintenance.

Tisho99:
@ValentinVinagre 

I think what duplicates the functionality is not having a general manifest.

What happens if we have 10 carrier modules that needs a manifest? All modules will have to define its own manifest report from scratch 

Your example of delivery_gls_asm does it. The module neither inherits from base_delivery_carrier_label, and as a result the carrier label wizard has also been duplicated

ValentinVinagre:
> delivery_gls_asm

Anything can come from there. Could you briefly review what would need to be changed to make everything work properly? In case of adding this module to OCA

Jaimermaccione (changes requested):
@Tisho99 I've checked the module and everything looks good. The logic behind it is solid, and the general manifest is being generated properly.

#### Just a couple of small things:

- **PDF formatting**: The PDF report doesn’t have a top margin with the company address, so from the second page onward the header gets a bit squashed into the table. Not a blocker, but something to fix it is not complicated.
image

- **README suggestion**: It would be useful to add a quick note explaining how the pickings are selected for the general manifest, just to make life easier for whoever uses or reviews the module later (It might be obvious, but I think it's better to spell it out clearly.) Here are the conditions as I understand them (correct me if I’m wrong)::

  > A delivery (picking) will be included in the general manifest only if all of the following conditions are met:
  > 
  > - **Included carrier**: The delivery has a carrier marked with "Include in General Manifest".
  > - **Done date**: The `date_done` of the picking is within the selected date range in the wizard.
  > - **State**: The delivery is in **"Done"** state.
  > - **Operation type**: Only **outgoing deliveries** are considered (picking type = `outgoing`).

That’s it! Solid work overall, just a couple of tweaks and it's good to go.

Tisho99:
@Jaimermaccione I have updated the readme. The header thing is a issue of the base report layout, not of this report

Jaimermaccione (approved):
@Tisho99 Perfect, then! Functionally valid and tested in a local environment. Thanks!

HaraldPanten (approved):
@ValentinVinagre Do we have your technical approval?

THX.

delivery_carrier_general_manifest/wizards/manifest_wizard.py
```diff
@@ -0,0 +1,78 @@
...
+    general_manifest_query = fields.Boolean(
+        string="General Manifest",
+        help="A general manifest query generates a generic general manifest "
+        "report using all carriers configured to be included on it",
+    )
+
+    @api.depends("general_manifest_query")
+    def _default_general_carrier_ids(self):
+        return self.env["delivery.carrier"].search(
+            [("include_in_general_manifest", "=", True)]
```
ValentinVinagre: adds multi-company support.

ValentinVinagre (changes requested):

delivery_carrier_general_manifest/wizards/manifest_wizard.py
```diff
@@ -0,0 +1,81 @@
...
+        help="A general manifest query generates a generic general manifest "
+        "report using all carriers configured to be included on it",
+    )
+
+    @api.depends("general_manifest_query")
+    def _default_general_carrier_ids(self):
+        return self.env["delivery.carrier"].search(
+            [
+                ("include_in_general_manifest", "=", True),
+                ("company_id", "in", [False] + self.env.companies.ids),
```
ValentinVinagre: why self.env.companies? the wizard don't have one field company_id ?

Tisho99: No, it hasn't

ValentinVinagre: Then a company field should be created, so that it takes the one activated by the user by default.

Tisho99: I disagree, i think that should be in the base module and not in this one

ValentinVinagre: I think it's correct to add it to the base module adding the functionality.

Tisho99: @ValentinVinagre Updated. Can you review?

delivery_carrier_general_manifest/wizards/manifest_wizard.py
```diff
@@ -0,0 +1,91 @@
...
+        for rec in self:
+            rec.general_carrier_ids = (
+                self.env["delivery.carrier"]
+                .search(
+                    [
+                        ("include_in_general_manifest", "=", True),
+                        ("company_id", "in", [False, rec.company_id.id]),
+                    ]
+                )
+                .ids
```
ValentinVinagre: If it doesn't find any records, does it give an error?

delivery_carrier_general_manifest/wizards/manifest_wizard.py
```diff
@@ -0,0 +1,91 @@
...
+class ManifestWizard(models.TransientModel):
+    _inherit = "manifest.wizard"
+
+    carrier_id = fields.Many2one(
+        required=False,
+    )
+    general_carrier_ids = fields.Many2many(
+        string="General Carriers",
+        comodel_name="delivery.carrier",
+        domain=[("include_in_general_manifest", "=", True)],
```
ValentinVinagre: domain and compute?

ValentinVinagre (commented):
Some comments

delivery_carrier_general_manifest/wizards/manifest_wizard_view.xml
```diff
@@ -0,0 +1,31 @@
...
+        <field name="arch" type="xml">
+            <field name="carrier_id" position="before">
+                <field name="general_manifest_query" widget="boolean_toggle" />
+                <field
+                    name="general_carrier_ids"
+                    invisible="not general_manifest_query"
+                    required="general_manifest_query"
+                    widget="many2many_tags"
+                />
+                <field name="company_id" invisible="not general_manifest_query" />
```
ValentinVinagre: add the multi-company group

delivery_carrier_general_manifest/wizards/manifest_wizard.py
Tisho99: The compute is used to calculate the default carriers, the user can unpick that carriers. The domain is to forbid the user selecting carriers that are not included in the general manifest

delivery_carrier_general_manifest/wizards/manifest_wizard.py
Tisho99: No, it does not give an error

delivery_carrier_general_manifest/wizards/manifest_wizard_view.xml
Tisho99: Added

delivery_carrier_general_manifest/wizards/manifest_wizard.py
ValentinVinagre: > The compute is used to calculate the default carriers, the user can unpick that carriers. The domain is to forbid the user selecting carriers that are not included in the general manifest

Shouldn't you also take into account the company you choose?

delivery_carrier_general_manifest/wizards/manifest_wizard.py
Tisho99: I have tried it, but the domain does not work, the company is not being updated in the domain, I assume that's because the model is transient

Apart from that, there is a record rule in carriers that do not let you select them if you have not their company active, so i think that the company does not need to be added to the domain

delivery_carrier_general_manifest/wizards/manifest_wizard.py
Tisho99: I have tried it, but the domain does not work, the company is not being updated in the domain, I assume that's because the model is transient

Apart from that, there is a record rule in carriers that do not let you select them if you have not their company active, so i think that the company does not need to be added to the domain

ValentinVinagre (approved):
LGTM 👍🏻

ValentinVinagre:
@rousseldenis Do you see this new functionality as correct?

delivery_carrier_general_manifest/wizards/manifest_wizard.py
```diff
@@ -0,0 +1,91 @@
...
+                        ("include_in_general_manifest", "=", True),
+                        ("company_id", "in", [False, rec.company_id.id]),
+                    ]
+                )
+                .ids
+            )
+
+    def _get_general_manifest_pickings(self):
+        self.ensure_one()
+        search_args = [
```
rousseldenis: @ValentinVinagre @Tisho99 My only concern is about this to retrieve the correct pickings.

Have you ever evaluated this : https://github.com/OCA/stock-logistics-transport/tree/16.0/shipment_advice

As it allows to load outgoing pickings in a shipment advice object. That allows you to be sure the pickings you try to print manifest are really the good ones.

jbaudoux (commented):
I was a bit confused by `general manifest`which is an official document for customs different from what you are doing. Maybe `global manifest`would be more suitable?

ValentinVinagre:
> I was a bit confused by `general manifest`which is an official document for customs different from what you are doing. Maybe `global manifest`would be more suitable?

It's a document used to review all departures according to carriers, etc.
There's no problem changing the name to global manifest if it's more understandable or appropriate.

delivery_carrier_general_manifest/wizards/manifest_wizard.py
ValentinVinagre: Hmm, I think they're different functionalities. The idea is that at the end of the day, or at some point during the day, the person in charge can create a report/printout of everything that has been delivered to carriers and to which ones.
I think the "shipment_advice" module is more complex.

jbaudoux: I also think shipment advice is more suitable for truck loading while this lighter module can be enough for parcel carriers

Tisho99:
Thanks for the review, @rousseldenis and @jbaudoux.

I've renamed the module to delivery_carrier_global_manifest as suggested.

Does it look better to you now? Have you seen another possible improvement?

jbaudoux:
> Thanks for the review, @rousseldenis and @jbaudoux.
> 
> I've renamed the module to delivery_carrier_global_manifest as suggested.
> 
> Does it look better to you now? Have you seen another possible improvement?

@Tisho99 looks like you need to fix the ES translation file also ;)

jbaudoux (approved):

HaraldPanten:
@jbaudoux @rousseldenis Could we merge this one, please?

All comments have been considered.

THX.

rousseldenis:
> @jbaudoux @rousseldenis Could we merge this one, please?
> 
> All comments have been considered.
> 
> THX.

Indeed, full pack :smiley: 

/ocabot merge nobump

PR #1017, opened 2025-05-21
[18.0][FIX] delivery_price_method: Skip base functionality

david-s73:
- The change has been made because it skipped the base functionality to check if the shipment has to be free or not if it exceeded the price set in the field `free_over` I leave the exact line of the Odoo code, as it is forcing the `delivery_type` to be the same as this in the `price_method` then does not make the `if` when it should be done.
[Odoo code](https://github.com/odoo/odoo/blob/18.0/addons/delivery/models/delivery_carrier.py#L282)

### DIFF
--- a/delivery_price_method/models/delivery_carrier.py
+++ b/delivery_price_method/models/delivery_carrier.py
@@ -42,3 +42,12 @@ class DeliveryCarrier(models.Model):
                 del rate["tracking_number"]  # remove offending key
                 res[index].update(rate)
         return res
+
+    def _get_price_from_picking(self, total, weight, volume, quantity, wv=0.0):
+        if (
+            self.price_method == "base_on_rule"
+            and self.free_over
+            and total >= self.amount
+        ):
+            return 0.0
+        return super()._get_price_from_picking(total, weight, volume, quantity, wv)
--- a/delivery_price_method/models/delivery_carrier.py
+++ b/delivery_price_method/models/delivery_carrier.py
@@ -21,7 +21,12 @@ class DeliveryCarrier(models.Model):
         # Trick the method for using all the upstream code for the
         # price computation in case of using fixed or base_on_rule.
         previous_method = False
-        if self.price_method in ("fixed", "base_on_rule"):
+        if self.price_method in ("fixed", "base_on_rule") and (
+            not self.free_over
+            or not self.amount
+            or self._compute_currency(order, order.amount_total, "pricelist_to_company")
+            < self.amount
+        ):
             previous_method = self.delivery_type
             self.sudo().delivery_type = self.price_method
         res = super().rate_shipment(order)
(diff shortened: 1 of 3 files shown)


david-s73:
@rousseldenis Can you check the fix when you can?

david-s73:
@pedrobaeza Can you check the fix when you can?

delivery_price_method/models/delivery_carrier.py
```diff
@@ -21,7 +21,9 @@ def rate_shipment(self, order):
         # Trick the method for using all the upstream code for the
         # price computation in case of using fixed or base_on_rule.
         previous_method = False
-        if self.price_method in ("fixed", "base_on_rule"):
+        if self.price_method in ("fixed", "base_on_rule") and (
+            not self.free_over or not self.amount or order.amount_total < self.amount
```
rousseldenis: Why not using `self._compute_currency()` as in base code ?

david-s73: I understand that you mean changing the `order.amount` to `self._compute_currency()` right?

pedrobaeza:
Can you add a test for the case?

david-s73:
@pedrobaeza I have already taken the tests.

pedrobaeza:
Why the extra commit?

david-s73:
@pedrobaeza Because it's another fix but from the same module, I put it in this pr.

pedrobaeza:
Well, starting with, the commit message is not correct: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message, and if it's other module, it's better to put it in another PR.

david-s73:
@pedrobaeza I have already modified the commit message, and since it is from the same module, I have put it in the same pr.

pedrobaeza (approved):

david-s73:
@pedrobaeza Merged?

pedrobaeza:
Pending from @rousseldenis' approval

Reyes4711-S73 (approved):
LGTM

Reyes4711-S73:
@rousseldenis @OCA/logistics-maintainers Please, can you merge this PR?

david-s73:
@rousseldenis Can you approve it?

delivery_price_method/models/delivery_carrier.py
rousseldenis: @david-s73 Yes, because `self.amount` should be compared with currency comparison.

rousseldenis:
@david-s73 sorry, I answered late but was very busy. One point is still questionning for me.

delivery_price_method/models/delivery_carrier.py
david-s73: @rousseldenis I have already made the changes you requested.

rousseldenis (approved):

PR #1023, opened 2025-06-12
[18.0][FIX]delivery_postlogistics: Update __manifest__.py for "application": False,

andreampiovesana:
"application": False,

### DIFF
--- a/delivery_postlogistics/__manifest__.py
+++ b/delivery_postlogistics/__manifest__.py
@@ -31,7 +31,7 @@
     ],
     "installable": True,
     "auto_install": False,
-    "application": True,
+    "application": False,
     "external_dependencies": {
         "python": [
             "openupgradelib",


rousseldenis:
@yvaucher @jbaudoux @grindtildeath An idea why this was set ?

jbaudoux (approved):

jbaudoux (changes requested):
It's like that on odoo delivery_xxx modules. So makes sense to have it here as well

jbaudoux:
> @yvaucher @jbaudoux @grindtildeath An idea why this was set ?

because it's like that on odoo delivery_xxxx connectors modules

yvaucher:
I assume that at the time when Odoo was a bit smaller we probably viewed this as a feature complete implementation for one carrier. Thus the choice was to give visibility as an application to those modules.

IIRC it has its own settings page and icon.

However as for bank integrations it can be argued it's not what we call an application in Odoo but just a carrier integration.

PR #1024, opened 2025-06-23
[15.0][IMP] delivery_purchase: Take into account the delivered quantity to calculate the shipping cost

CarlosRoca13:
Before these changes, the shipping cost set on the delivery note always took into account the quantity from the purchase order.

After making these changes, the behavior can be modified so that if the system parameter "delivery_purchase.use_delivered_qty_to_set_cost" is set to True, the received quantity will be taken into account when receiving partial quantities on the delivery note.

cc @Tecnativa TT56831

ping @pedrobaeza @victoralmau

### DIFF
--- a/delivery_purchase/README.rst
+++ b/delivery_purchase/README.rst
@@ -39,6 +39,13 @@ applying the delivery method to purchases.
 .. contents::
    :local:
 
+Configuration
+=============
+
+If you want to recompute the delivery cost on picking validation based
+on received quantity, you need to check the option
+"Recompute Delivery Price When Validate Picking"
+
 Usage
 =====
 
@@ -86,6 +93,7 @@ Contributors
   * Ernesto Tejeda
   * Pedro M. Baeza
   * Vicent Cubells
+  * Carlos Roca
 
 Maintainers
 ~~~~~~~~~~~
--- a/delivery_purchase/models/delivery_carrier.py
+++ b/delivery_purchase/models/delivery_carrier.py
@@ -3,6 +3,7 @@
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError, ValidationError
+from odoo.tools.misc import str2bool
 
 
 class DeliveryCarrier(models.Model):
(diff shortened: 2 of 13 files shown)


delivery_purchase/README.rst
```diff
@@ -39,6 +39,13 @@ applying the delivery method to purchases.
 .. contents::
    :local:
 
+Configuration
+=============
+
+If you want the shipping cost to take the delivered quantity into account,
```
pedrobaeza: The parameter should be to recompute the delivery cost on validation, not this one. And please add it to the res.config screen.

delivery_purchase/models/delivery_carrier.py
```diff
@@ -159,4 +162,20 @@ def _purchase_get_price_available(self, order):
             order.company_id,
             order.date_order or fields.Date.today(),
         )
+        return weight, volume, quantity, total
+
+    def _purchase_get_price_available(self, order):
```
pedrobaeza, line 169: As you have the control over the whole module, no need to pass contexts here. Just do 2 methods, one for the PO, and another for the picking.

CarlosRoca13:
Changes done @pedrobaeza

delivery_purchase/models/res_config_settings.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2025 Tecnativa - Carlos Roca
```
pedrobaeza: This and the view should go into `wizards` directory, being a transient model. Odoo is not complying with this guideline though.

delivery_purchase/models/delivery_carrier.py
```diff
@@ -159,4 +164,27 @@ def _purchase_get_price_available(self, order):
...
+        self.ensure_one()
+        self = self.sudo()
+        order = order.sudo()
+        lines = order.order_line
+        qty_field = "product_uom_qty"
+        weight, volume, quantity, total = self._get_delivery_data(
+            order, lines, qty_field
+        )
```
pedrobaeza: Don't put variables with only one usage (the same for the rest of the code):

```suggestion
        order = order.sudo()
        weight, volume, quantity, total = self._get_delivery_data(
            order, order.order_line, "product_uom_qty"
        )
```

Do the code more readable not having to follow which is the value of such local variables.

CarlosRoca13:
Changed @pedrobaeza

pedrobaeza (approved):
The part of amounts is not really correct, as it's still using purchase order amounts, but let's continue this way for the moment.

victoralmau (approved):

PR #1025, opened 2025-06-24
[15.0][ADD] delivery_purchase_multi_destination: New module to use the multi destination type to set the delivery cost on purchases

CarlosRoca13:
cc @Tecnativa TT56831

ping @pedrobaeza @victoralmau


pedrobaeza:
I know you have called it according the purchase semantics, but both the carrier view and the other module (`delivery_multi_destination`) are called the other way, so this one should be `delivery_purchase_multi_destination`.

delivery_purchase_multi_destination/models/delivery_carrier.py
```diff
@@ -0,0 +1,65 @@
...
+from odoo import _, models
+from odoo.exceptions import ValidationError
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = "delivery.carrier"
+
+    def purchase_rate_shipment(self, order):
+        # We have to override this method for getting the proper price
+        # according destination on sales orders.
```
pedrobaeza:
```suggestion
        # according destination on purchase orders.
```

delivery_purchase_multi_destination/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,8 @@
+To configure delivery methods with multiple origins:
```
pedrobaeza: This are the fragments from `delivery_multi_destination`. Here you have to put that "refer to the module ... for the configuration steps".

delivery_purchase_multi_destination/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,6 @@
+This module allows to set different price rules depending on the origin.
```
pedrobaeza: You have to put here that this is the glue module for using multi destinations for purchase sendings.

pedrobaeza (changes requested):
Please add tests.

CarlosRoca13:
Done @pedrobaeza

delivery_purchase_multi_destination/models/purchase_order.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2025 Tecnativa - Carlos Roca
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    @api.depends("partner_id")
```
pedrobaeza, line 13: Add a explanation comment:

```suggestion
    # Introduce this extra dependency in the compute
    @api.depends("partner_id")
```

delivery_purchase_multi_destination/models/delivery_carrier.py
```diff
@@ -0,0 +1,66 @@
...
+                            picking_res = super(
+                                DeliveryCarrier, subcarrier
+                            ).purchase_send_shipping(p)
+                            break
+                        except Exception:  # pylint: disable=except-pass
+                            pass
+                        finally:
+                            p.carrier_id = carrier
+                if not picking_res:
+                    raise ValidationError(_("There is no matching delivery rule."))
```
pedrobaeza, line 70: Why adding this and the try/except?

CarlosRoca13:
@pedrobaeza I've added comments on all the points you mentioned to ensure everything is clear.

pedrobaeza (approved):

victoralmau (approved):

PR #879, opened 2024-09-11
[17.0][MIG] delivery_purchase: Migration to 17.0

SodexisTeam:

PR #1026, opened 2025-06-25
[17.0][MIG] delivery_purchase: Migration to 17.0

CarlosRoca13:
cc @Tecnativa TT52416

This PR supersedes PR https://github.com/OCA/delivery-carrier/pull/879 by addressing the changes requested by @rousseldenis and incorporating the changes introduced in PR https://github.com/OCA/delivery-carrier/pull/1024.

ping @pedrobaeza @victoralmau


rousseldenis:
@CarlosRoca13 thanks for this. When supersedes, if possible, you should keep the existing work that has been done by keeping original commits. Thanks

CarlosRoca13:
> @CarlosRoca13 thanks for this. When supersedes, if possible, you should keep the existing work that has been done by keeping original commits. Thanks

You're right, I attributed the authorship of my first commit to the author of the original PR.

pedrobaeza (commented):
Tested on runboat + code review.

Several comments to improve:

- Please put the commit for #1024 before the migration ones.
- The delivery method and carrier price can be changed after confirming the PO, which doesn't seem correct, specially as this is not propagated to open pickings.
- A PO line is always written when validating the picking. This behavior may be not wanted for some users. Can this be optional by a parameter (also for 15)?

CarlosRoca13:
Done @pedrobaeza

pedrobaeza (approved):

victoralmau (approved):

PR #1021, opened 2025-06-03
[ADD][18.0] delivery_carrier_picking_valid

mmequignon:

PR #1031, opened 2025-07-02
[ADD][18.0] delivery_carrier_picking_valid_packaging_weight

mmequignon:
based on https://github.com/OCA/delivery-carrier/pull/1021

### DIFF
--- a/delivery_carrier_picking_valid_packaging_weight/models/stock_move.py
+++ b/delivery_carrier_picking_valid_packaging_weight/models/stock_move.py
@@ -1,0 +1,53 @@
+# Copyright 2025 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import api, fields, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    estimated_shipping_weight = fields.Float(
+        compute="_compute_estimated_shipping_weight"
+    )
+
+    def _get_estimated_shipping_quantity(self):
+        self.ensure_one()
+        return self.product_uom_qty
+
+    def _compute_estimated_shipping_weight(self):
+        for record in self:
+            product_qty = record._get_estimated_shipping_quantity()
+            weight = record.product_id.get_total_weight_from_packaging(product_qty)
+            record.estimated_shipping_weight = weight
+
+    @api.model
+    def read_group(
+        self,
+        domain,
+        fields,
+        groupby,
+        offset=0,
+        limit=None,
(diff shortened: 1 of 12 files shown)


delivery_carrier_picking_valid_packaging_weight/models/stock_picking.py
```diff
@@ -0,0 +1,19 @@
...
+    _inherit = "stock.picking"
+
+    estimated_shipping_weight = fields.Float(
+        compute="_compute_estimated_picking_weight"
+    )
+
+    def _compute_estimated_picking_weight(self):
+        for record in self:
+            record.estimated_shipping_weight = sum(
+                self.move_ids.mapped("estimated_shipping_weight")
```
simahawk, line 18: :fire: 
```suggestion
                record.move_ids.mapped("estimated_shipping_weight")
```

delivery_carrier_picking_valid_packaging_weight/models/stock_move.py
```diff
@@ -0,0 +1,53 @@
...
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    estimated_shipping_weight = fields.Float(
+        compute="_compute_estimated_shipping_weight"
+    )
+
+    def _get_estimated_shipping_quantity(self):
+        self.ensure_one()
```
simahawk, line 15: this is your own private method to be used in the loop, make no sense to call `ensure_one` IMO

delivery_carrier_picking_valid_packaging_weight/tests/test_picking_valid.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2025 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+
+from odoo.addons.delivery_carrier_picking_valid.tests.common import Common
```
simahawk, line 5: you should fix the use of BaseCommon there ;)

simahawk (changes requested):

jbaudoux (changes requested):
This module is useless.
Base module is using `picking._get_estimated_weight()` and delivery_total_weight_from_packaging is already ensuring the glue with product_total_weight_from_packaging

delivery_carrier_picking_valid_packaging_weight/models/stock_picking.py
mmequignon: nice catch, thanks

PR #1033, opened 2025-07-14
[14.0][IMP] delivery_schenker: Request Data in raised Error

raumschmiede-joshuaL:
For easier error handling it is good to have the request data in the error that is raised

### DIFF
--- a/delivery_schenker/models/schenker_request.py
+++ b/delivery_schenker/models/schenker_request.py
@@ -1,6 +1,8 @@
 # Copyright 2021 Tecnativa - David Vidal
+# Copyright 2025 Raumschmiede GmbH
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+import json
 import logging
 from xml.etree import ElementTree as ET
 
@@ -62,6 +64,8 @@ class SchenkerRequest:
                     response = service(vals)
                 else:
                     response = service(**vals)
+
+                data = self._dump_request_data(vals)
                 try:
                     root = ET.fromstring(response.text)
                     error_text = next(root.iter("faultstring")).text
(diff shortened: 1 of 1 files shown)


mt-software-de:
Thanks for the contribution. Could you please check pre-commit?

mt-software-de:
@jbaudoux @chienandalu @hildickethan Could you maybe review this?

delivery_schenker/models/schenker_request.py
```diff
@@ -72,17 +76,27 @@ def _process_reply(self, service, vals=None, send_as_kw=False):
...
+                    msg = e.message + _("\nRequest Data:\n%s") % data
+                    raise Fault(msg)
         return response
 
+    def _dump_request_data(self, data):
+        res = json.dumps(data, indent="    ", default=str)
+        if self.access_key:
+            res = res.replace(self.access_key, "<REMOVED>")
+
+        return res
```
mt-software-de, line 98: I would prefer to not use str.replace, instead removing the actual keys which holding the access key.

mt-software-de (changes requested):

delivery_schenker/models/schenker_request.py
raumschmiede-joshuaL: The API endpoint calls prepare the data in different ways.

`_tracking_api_credentials` has the key "AccessKey" in the 1st level of keys,
`_shipping_api_credentials` in applicationArea and with a lowercase A.
`_send_shipping`, which uses `_shipping_api_credentials`, puts another wrapper around.

So for each case I had to check other keys. Converting it to a string and replacing it is easier

hildickethan (approved):

mt-software-de (approved):

raumschmiede-joshuaL:
@hparfr, can you merge this, please?

hparfr (approved):
/ocabot merge minor

PR #1038, opened 2025-07-24
[18.0][FIX]`delivery_postlogistics`: Replace ``pack.env`` by ``self.env`` because ``pack`` parameter can be ``None``

Camille0907:
In ``stock.picking::postlogistics_label_prepare_attributes`` method, parameter ``pack`` is optionnal. If it is not set, ``pack.env._(`` triggers a traceback. Here we replace it by ``self.env._(``, for it to work in any case


yankinmax (approved):

hparfr (approved):
/ocabot merge minor

PR #1045, opened 2025-08-28
 [18.0][FIX] delivery_postlogistics: Fix package id on label

wouitmil:
Ensure that the package_id is set

It's needed when stock_picking._generate_postlogistics_label is called from a generate_label on a stock_picking_batch

### DIFF
--- a/delivery_postlogistics/models/stock_picking.py
+++ b/delivery_postlogistics/models/stock_picking.py
@@ -61,17 +61,6 @@ Please use _get_quant_packages_from_picking instead."
 
         return self.env["stock.quant.package"].browse(package_ids)
 
-    def get_shipping_label_values(self, label):
-        # TODO: consider to depends on base_delivery_carrier_label
-        self.ensure_one()
-        return {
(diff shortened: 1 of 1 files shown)


delivery_postlogistics/models/stock_picking.py
```diff
@@ -61,17 +61,6 @@ def _get_quant_packages_from_picking(self):
 
         return self.env["stock.quant.package"].browse(package_ids)
 
-    def get_shipping_label_values(self, label):
```
leemannd, line 64: This is already defined in the module `delivery_carrier_shipping_label`. Thanks

leemannd (approved):

delivery_postlogistics/models/stock_picking.py
wouitmil: Exact, thats why it's removed as delivery_carrier_shipping_label is a dependency of delivery_postlogistics

hparfr (approved):
/ocabot merge minor

PR #1057, opened 2025-09-30
[16.0][ADD] delivery_dachser: New module

victoralmau:
New module

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT57270


victoralmau:
Finally, this has been validated. Can you review it?

pedrobaeza (approved):

delivery_dachser/models/delivery_carrier.py
```diff
@@ -0,0 +1,119 @@
...
+            )
+            attachments = []
+            if response.get("label"):
+                attachments = [
+                    (
+                        "dachser.pdf",
+                        (base64.b64decode(response.get("label"))),
+                    )
+                ]
+            picking.message_post(body=_(""), attachments=attachments)
```
carlos-lopez-tecnativa:
```suggestion
            picking.message_post(body="", attachments=attachments)
```

delivery_dachser/tests/test_delivery_dachser.py
```diff
@@ -0,0 +1,92 @@
...
+        self.picking.delivery_state = False
+        self.picking.tracking_state_update()
+        self.assertTrue(self.picking.delivery_state)
+        self.assertTrue(self.picking.tracking_state_history)
+
+    def test_04_dachser_picking_cancel_shipment(self):
+        if not self.carrier_dachser.dachser_api_key:
+            self.skipTest("Without Dachser Api Key")
+        self.picking.carrier_tracking_ref = "test"
+        with self.assertRaises(UserError):
```
carlos-lopez-tecnativa: I preferer use `assertRaisesRegex` indicate to expected message to be more specfiic error.

carlos-lopez-tecnativa (approved):
LGTM, just a minor comment, non-blocking. If you can apply it, that would be great.

delivery_dachser/models/delivery_carrier.py
victoralmau: With this change, the message 
`delivery_dachser/models/delivery_carrier.py:89: [C8107(translation-required), DeliveryCarrier.dachser_send_shipping] String parameter on "message_post" requires translation. Use body=_('')`
 is displayed.

I will make the change and add the corresponding exception.

pedrobaeza: IMO, the problem is to post an empty message instead of simply adding the attachments. Or post a short message indicating that the label is downloaded.

victoralmau: Are there any similar examples in other carrier modules: https://github.com/OCA/delivery-carrier/blob/ff81a75fb41dd14cfa2b6deae785a72acd417031/delivery_correos_express/models/delivery_carrier.py#L187

pedrobaeza: Yeah, but that doesn't mean that is correct...

PR #1060, opened 2025-10-16
[16.0][FIX] delivery_purchase: Dropshipping compatibility (purchase order)

victoralmau:
Dropshipping compatibility (purchase order)

Use case example:
- Create a sales order and use the Dropshipping route
- Set the delivery address as France, for example
- Create a rule-based carrier only available for France
- Confirm the sales order
- Set the previously created carrier on the purchase order
- The carrier must be applied correctly, using the delivery address (France) as a reference

@Tecnativa TT58174


pedrobaeza (commented):
Oh, some missing pieces...

/ocabot merge patch

pedrobaeza (approved):

PR #1061, opened 2025-10-17
[16.0][IMP] delivery_purchase: Don't update the shipping cost of confirmed POs

victoralmau:
Don't update the shipping cost of confirmed POs

@Tecnativa TT58520

### DIFF
--- a/delivery_purchase/tests/test_delivery_purchase.py
+++ b/delivery_purchase/tests/test_delivery_purchase.py
@@ -1,6 +1,8 @@
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
+from odoo import Command
 from odoo.tests import Form
+from odoo.tools import mute_logger
 
 from odoo.addons.base.tests.common import BaseCommon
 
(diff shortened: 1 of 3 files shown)


pedrobaeza (changes requested):
I don't think this should go into this direction, but the purchase price must be readonly in the very moment it's confirmed, and from that moment, the delivery price is controlled in the pickings. If not, you will have a mix of prices depending on the number of receipts you do.

delivery_purchase/tests/test_delivery_purchase.py
```diff
@@ -89,6 +90,48 @@ def test_purchase_delivery_line_invoice_status(self):
...
+    def test_purchase_picking_price(self):
+        carrier = self.env["delivery.carrier"].create(
+            {
+                "name": "Carrier Rule custom",
+                "product_id": self.delivery_product.id,
+                "delivery_type": "base_on_rule",
+                "price_rule_ids": [
+                    (
+                        0,
+                        0,
```
pedrobaeza:
```suggestion
                    Command.Create(
```

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):

PR #1062, opened 2025-10-17
[18.0][FIX] delivery_carrier_pricelist: prevent ``AccessError``

SilvioC2C:


henrybackman (approved):

delivery_carrier_pricelist/models/delivery_carrier.py
```diff
@@ -25,18 +30,34 @@ class DeliveryCarrier(models.Model):
...
+        if self.invoice_policy != "pricelist" or self.delivery_type == "pricelist":
+            return super().rate_shipment(order)
+
+        # Prevent ``AccessError`` if needed: by default, module ``delivery`` only allows
+        # Sales / Administrator users to have ``write`` access on ``delivery.carrier``,
+        # but lots of modules use this method, even for automations on ``sale.order``
+        # creation/updates (eg: module ``delivery_auto_refresh`` or OCA/sale-workflow
+        # ``sale_order_carrier_auto_assign``); therefore, if necessary, we go ``sudo()``
+        # before updating/reverting the delivery type on the carrier
+        try:
```
simahawk: I don't understand what is the point to check for permission and then bypassing it...
This is a technical thing that must be executed w/ sudo, that's it.

simahawk (changes requested):

delivery_carrier_pricelist/models/delivery_carrier.py
simahawk: as we said: just use `sudo` when you set the fields

delivery_carrier_pricelist/models/delivery_carrier.py
```diff
@@ -1,8 +1,12 @@
 # Copyright 2020 Camptocamp
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+from logging import getLogger
+
 from odoo import api, fields, models
 
+_logger = getLogger(__name__)
```
simahawk: to be discarded

delivery_carrier_pricelist/models/delivery_carrier.py
```diff
@@ -25,18 +29,23 @@ class DeliveryCarrier(models.Model):
...
+        # Odoo uses ``getattr(self, '%s_rate_shipment' % self.delivery_type)`` in its
+        # base method ``rate_shipment()`` to lookup which function to use, so we
+        # temporarily change the ``delivery_type`` if needed.
+
+        # Quick check: if the invoice policy is not "pricelist", or the delivery type is
+        # already set to "pricelist", we don't have to do anything
+        if self.invoice_policy != "pricelist" or self.delivery_type == "pricelist":
+            return super().rate_shipment(order)
+
+        # Use ``sudo()`` to prevent ``AccessError`` when updating the delivery type
```
simahawk, line 46: IMO we should finally propose a change to odoo core avoid this stuff.
I would propose the addition of `_rate_shipment_get_handler` instead of using getattr directly.
This way, in v19 or 20 we can get rid of this hack.
Also note that some specific carrier implementation rely on computed fields depending on `delivery_type` so we might trigger unnecessary changes.

simahawk:
Updated

delivery_carrier_pricelist/models/delivery_carrier.py
simahawk: added

simahawk (approved):

jbaudoux (approved):

PR #1065, opened 2025-10-25
[18.0][IMP] delivery_carrier_picking_valid_dangerous_goods: availability

jbaudoux:
* Move restriction on dangerous goods inside standard 'Availability' tab
* Apply also on SO => TODO Add test *- EDIT: done*

### DIFF
--- a/delivery_carrier_picking_valid_dangerous_goods/models/delivery_carrier.py
+++ b/delivery_carrier_picking_valid_dangerous_goods/models/delivery_carrier.py
@@ -1,4 +1,5 @@
 # Copyright 2025 Camptocamp SA
+# Copyright 2025 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 from odoo import fields, models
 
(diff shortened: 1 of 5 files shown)


jbaudoux:
rebased and test added on SO

delivery_carrier_picking_valid_dangerous_goods/models/delivery_carrier.py
```diff
@@ -11,19 +11,24 @@ class DeliveryCarrier(models.Model):
...
     def _match_picking(self, picking):
-        return super()._match_picking(picking) and self._match_dangerous_goods(picking)
+        return super()._match_picking(picking) and self._match_dangerous_goods(
+            picking.move_ids.product_id
+        )
 
-    def _match_dangerous_goods(self, picking):
-        # Returns True if picking is compliant with carrier regarding limited
-        # amounts of dangerous goods
+    def _match_dangerous_goods(self, products):
```
mmequignon, line 28: I always found this part weird, this is limited amount, but we aren't checking qties, weight or whatever. If we find any, the picking/order isn't valid.

mmequignon (approved):

PR #1068, opened 2025-10-27
[17.0][FIX] delivery_cbl: add super() call in get_manifest_file of manifest_wizard

Tisho99:
The get_manifest_file() function of the manifest_wizard model can be inherited in multiple models to add a custom manifest file depending on the carrier type, just like we do in this module.

A super() call should be added in this inheritance to ensure module interoperability.

The _get_cbl_picking() function only returns CBL picking, so there is no problem restricting the delivery type.

### DIFF
--- a/delivery_cbl/wizard/manifest_wizard.py
+++ b/delivery_cbl/wizard/manifest_wizard.py
@@ -25,6 +25,8 @@ class ManifestWizard(models.TransientModel):
 
     def get_manifest_file(self):
         self.ensure_one()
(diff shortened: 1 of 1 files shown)


Anxo82:
Technical review ok :+1:

Anxo82 (approved):

Tisho99:
Hello @pedrobaeza 

This PR is related to the other one for delivery_cbl that you reviewed earlier.

Could you please review this one as well?

Thank you!

pedrobaeza (approved):
/ocabot merge patch

PR #1070, opened 2025-10-29
[18.0][ADD] delivery_pre_shipping: New module to send a delivery to the carrier before the picking is validated

carlos-lopez-tecnativa:
TT58440


pedrobaeza:
I would remove the `carrier` word from the module name.

pedrobaeza:
And reading the context README, I would rename it to `delivery_pre_shipping`, or at least, `delivery_manual_shipping`, to use the same terminology as Odoo.

delivery_pre_shipping/readme/CONTEXT.md
```diff
@@ -0,0 +1,4 @@
+In some cases or industries, it is necessary to obtain a tracking reference before validating the picking.
+
+For example, when a company sells products through an external marketplace that requires a tracking number for the delivery process even though the company has not yet shipped the products, as they might still be out of stock and require resupply from a vendor. 
+The picking may be validated the next day or several days later, but a tracking number is still needed to report it to the marketplace.
```
pedrobaeza:
```suggestion

The picking may be validated the next day or several days later, but a tracking number is still needed to report it to the marketplace.
```

delivery_pre_shipping/views/stock_picking_views.xml
```diff
@@ -0,0 +1,21 @@
...
+            name="inherit_id"
+            ref="stock_delivery.view_picking_withcarrier_out_form"
+        />
+        <field name="arch" type="xml">
+            <xpath expr="//button[@name='send_to_shipper']" position="after">
+                <button
+                    name="action_send_to_shipper_manually"
+                    type="object"
+                    string="Send to shipper manually"
+                    invisible="carrier_tracking_ref or delivery_type in ['fixed', 'base_on_rule'] or not delivery_type or state != 'assigned' or picking_type_code == 'incoming'"
```
pedrobaeza: The state condition is correct? I would say that it should be `in ('draft', 'cancel', 'done')`

pedrobaeza:
Thinking twice about this, in fact the way Odoo intends to do this is through flexible routes + checking the mark "Generate Shipping Labels", and having 2 pickings when this happens. Please reflect this in the README, but it's not bad to have this alternative anyway.

pedrobaeza (approved):

sergio-teruel:
@carlos-lopez-tecnativa ¿Se puede revisar ya?

carlos-lopez-tecnativa:
> @carlos-lopez-tecnativa ¿Se puede revisar ya?

Yes, please. This PR is ready for review and merge.

carlos-lopez-tecnativa:
> Thinking twice about this, in fact the way Odoo intends to do this is through flexible routes + checking the mark "Generate Shipping Labels", and having 2 pickings when this happens. Please reflect this in the README, but it's not bad to have this alternative anyway.

Added to the README, but this feature only works when you have stock available. In this case, this module is intended to work when you do not have stock yet and you have not validated the picking.

sergio-teruel (approved):

delivery_pre_shipping/views/stock_picking_views.xml
```diff
@@ -0,0 +1,27 @@
...
+            name="inherit_id"
+            ref="stock_delivery.view_picking_withcarrier_out_form"
+        />
+        <field name="arch" type="xml">
+            <xpath expr="//button[@name='send_to_shipper']" position="after">
+                <button
+                    name="action_send_to_shipper_manually"
+                    type="object"
+                    string="Send to shipper manually"
+                    invisible="carrier_tracking_ref or delivery_type in ['fixed', 'base_on_rule'] or not delivery_type or state in ['draft', 'cancel', 'done'] or picking_type_code not in ['outgoing', 'internal']"
```
christian-ramos-tecnativa, line 16: Sorry for the delayed review but I'm not sure about adding to the invisible condition to `picking_type_code not in ['outgoing', 'internal']`, some delivery carriers like the [postlogistics](https://github.com/OCA/delivery-carrier/blob/18.0/delivery_postlogistics/models/stock_picking.py#L241) one is capable of using the carrier for outgoing and incoming deliveries

delivery_pre_shipping/views/stock_picking_views.xml
```diff
@@ -0,0 +1,27 @@
...
+                />
+            </xpath>
+            <!-- Allow to cancel the shipment if it is sent to the shipper by mistake -->
+            <xpath expr="//button[@name='cancel_shipment']" position="attributes">
+                <attribute
+                    name="invisible"
+                >not carrier_tracking_ref or delivery_type in ['fixed', 'base_on_rule'] or not delivery_type or state in ['draft', 'cancel'] or picking_type_code not in ['outgoing', 'internal']</attribute>
```
christian-ramos-tecnativa: Same comment here as above, even worse as this make incoming deliveries not able to be canceled

christian-ramos-tecnativa (changes requested):

delivery_pre_shipping/views/stock_picking_views.xml
carlos-lopez-tecnativa: Sorry, but in this case, the goal of this module is to send a picking before it is validated, while keeping the same Odoo logic. In the codebase, you can send the picking to the carrier for internal transfers or deliveries, but not for incoming shipments.

https://github.com/odoo/odoo/blob/112170149b7dd8a7bbb18972ac1a5dc6eaf6abf4/addons/stock_delivery/models/stock_picking.py#L96
https://github.com/odoo/odoo/blob/112170149b7dd8a7bbb18972ac1a5dc6eaf6abf4/addons/stock_delivery/views/delivery_view.xml#L47

delivery_pre_shipping/views/stock_picking_views.xml
carlos-lopez-tecnativa: In this case, I saw that Odoo displays the button when there is a carrier_tracking_ref and does not consider the picking type. Therefore, to prevent interference with other modules, I preferred to create a new button with a specific invisible attribute. Please let me know if this is OK for you.

https://github.com/odoo/odoo/blob/112170149b7dd8a7bbb18972ac1a5dc6eaf6abf4/addons/stock_delivery/views/delivery_view.xml#L28

delivery_pre_shipping/views/stock_picking_views.xml
christian-ramos-tecnativa: My bad, I mixed the `send_to_shipper` with the `print_return_label` button, as that will require a new `action_print_return_label_manually` we can leave it as it is

christian-ramos-tecnativa (approved):
LGTM

eduezerouali-tecnativa (approved):

eduezerouali-tecnativa:
Could we merge this?

carlos-lopez-tecnativa:
ping @pedrobaeza

PR #1074, opened 2025-11-03
[18.0][FIX] delivery_postlogistics: drop transaction rollback/commit

mmequignon:
When sending a request for multiple packages, postlogistics might return an error for one of those.
However, nobody is invoiced while one of the successful is scanned by swisspost. It means that it's safe if we do not save the successful labels when one fails.
Raising an exception when a label fails lets odoo manage the rollback. When user fixed the package, he can then try to confirm the picking again, and generate new labels.

Tests will come.


TDu (approved):
LG, makes more sense

jbaudoux (approved):

PR #1080, opened 2025-11-14
[18.0][ADD] delivery_carrier_report_to_printer: New module that allows setting a report on delivery.carrier and enables configuring it to print directly

carlos-lopez-tecnativa:
TT58969
@Tecnativa @pedrobaeza @sergio-teruel @christian-ramos-tecnativa could you please review this?


delivery_carrier_report_to_printer/models/delivery_carrier.py
```diff
@@ -0,0 +1,47 @@
...
+        domain = [("res_id", "in", pickings.ids), ("res_model", "=", pickings._name)]
+        # Capture the created attachments during the shipping process
+        current_attachments = self.env["ir.attachment"].search(domain, order="id")
+        res = super().send_shipping(pickings)
+        new_attachments = (
+            self.env["ir.attachment"].search(domain, order="id") - current_attachments
+        )
+        if self.label_report_id and new_attachments:
+            self._print_carrier_label(new_attachments)
+        return res
```
christian-ramos-tecnativa, line 27: I think we should avoid both searches for  cases where `label_report_id` is not set checking the field first
```suggestion
        if not self.label_report_id:
            return super().send_shipping(pickings)
        domain = [("res_id", "in", pickings.ids), ("res_model", "=", pickings._name)]
        # Capture the created attachments during the shipping process
        current_attachments = self.env["ir.attachment"].search(domain, order="id")
        res = super().send_shipping(pickings)
        new_attachments = (
            self.env["ir.attachment"].search(domain, order="id") - current_attachments
        )
        if new_attachments:
            self._print_carrier_label(new_attachments)
        return res
```

christian-ramos-tecnativa (changes requested):

delivery_carrier_report_to_printer/models/delivery_carrier.py
carlos-lopez-tecnativa: Done, thanks.

KDRdev (approved):

jans23:
Tested and its working fine for us.

sergio-teruel (approved):

christian-ramos-tecnativa (approved):

PR #1099, opened 2026-01-06
[19.0] [MIG] delivery_carrier_agency: Migration to 19.0

BhaveshHeliconia:


delivery_carrier_agency/tests/test_carrier_agency.py
```diff
@@ -1,14 +1,16 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import Command
+from odoo.tests import tagged
 
 from odoo.addons.base.tests.common import BaseCommon
 
 
+@tagged("post_install", "-at_install")
```
florian-dacosta: Why does it need to change on post install ?

BhaveshHeliconia: I added the post_install tag while I was trying to figure out why the tests weren’t running. In Odoo 18, tests worked with -i module_name --test-enable, but in Odoo 19, if the module is already installed, tests only run with -u module_name --test-enable. At that time, I wasn’t sure what was going wrong, so I added the tag while testing different options.

That was only for debugging, and I forgot to remove it afterward. I’ll remove the post_install tag now. Sorry about that, and thanks for catching it.

florian-dacosta (approved):

florian-dacosta:
@hparfr If you want to review

hparfr (approved):
/ocabot merge nobump

PR #1002, opened 2025-03-21
[17.0][MIG] delivery_auto_refresh: Migration to 17.0

victoralmau:
Migration to 17.0

Locked by:
- [x] `sale_order_carrier_auto_assign`: https://github.com/OCA/sale-workflow/pull/3635

Please @chienandalu can you review it?

@Tecnativa TT52316

PR #1106, opened 2026-01-26
[FIX][18.0] delivery_auto_refresh: Show Delivery Carrier in sale form

Shide:
Restore missing view that shows carrier field under payment terms in Sale Order form view.

image

MT-13352 @moduon @chienandalu @Gelojr @fcvalgar please review if you want 😄


Shide:
In order to change carrier when partner changes, the line of the SO must be dropped and another one with the correct carrier will be created.

chienandalu (commented):
Not sure this module is the right place to make this field visible again. It was dropped in the refactor made in the migration to v17 (and surely it wasn't the proper module in the first place anyway): https://github.com/OCA/delivery-carrier/pull/1002/changes/dec8f2816c9a0ff41e98d763496fab35b68cdeaf#diff-04560e886f8ea29a506caeb3d290d5c46f6d944c7da493df68867038a627d7afL11-L13

PR #1108, opened 2026-02-04
[18.0][IMP] delivery_state: Add field to store raw tracking response

carlos-lopez-tecnativa:
The purpose of this field is to store tracking information in raw format, because some carriers stop responding through the API after a certain period of time. In cases where the information is required later (for example, audits or documentation requests), this field can be used to retrieve tracking data that may no longer be available via the API.

Note: Each delivery module must implement the logic to retrieve and store this information.

TT59692
@Tecnativa @sergio-teruel @pedrobaeza could you please review this?

### DIFF
--- a/delivery_state/models/stock_picking.py
+++ b/delivery_state/models/stock_picking.py
@@ -17,6 +17,8 @@ class StockPicking(models.Model):
         string="Delivery Date",
         readonly=True,
     )
+    # Technical field to store raw tracking data from the carrier API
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Now that I see this, should the rest of the fields be `copy=False` probably.

pedrobaeza (approved):

christian-ramos-tecnativa (approved):
LGTM

PR #1109, opened 2026-02-05
[18.0][IMP] delivery_ups_oca: Save the tracking response

carlos-lopez-tecnativa:
Depends on:

- [x] https://github.com/OCA/delivery-carrier/pull/1108

TT59692
@Tecnativa @sergio-teruel @pedrobaeza could you please review this?

### DIFF
--- a/delivery_ups_oca/models/ups_request.py
+++ b/delivery_ups_oca/models/ups_request.py
@@ -429,8 +429,9 @@ class UpsRequest:
                             warn=warning.get("message"),
                         )
                     )
-            picking.write({"tracking_json": shipment})
-        except Exception:
+            picking.write({"tracking_json": shipment, "pod_error": False})
+        except Exception as ex:
+            picking.write({"pod_error": str(ex)})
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Should the whole response be stored? What are the extra data that are not decoded on the later code?

carlos-lopez-tecnativa:
> Should the whole response be stored? What are the extra data that are not decoded on the later code?

I do not know exactly which JSON key contains the tracking information because UPS does not generate a POD, so I think it is better to store the entire response so that we can later generate a report or something based on this field. The current tracking information is only saved in a text field and cannot be manipulated.

pedrobaeza:
Yeah, but if the tracking information is enough, the rest is not needed, just junk, but well, as you prefer.

carlos-lopez-tecnativa:
ping @pedrobaeza

pedrobaeza (approved):

eduezerouali-tecnativa (approved):

PR #1110, opened 2026-02-05
[18.0][MIG] delivery_carrier_multi_zip: Migration to 18.0

Anxo82:
@HaraldPanten @luis-ron 

- [T-9291]


Jaimermaccione (approved):
@Anxo82 Functional review completed in a local environment.

Tests performed:
- Created multiple delivery methods with several ZIP code ranges, successfully confirming that the module allows the configuration it claims.
- Tested the standard Odoo delivery method assignment workflow.
- Verified that ZIP range filtering works as expected by testing contacts with intermediate ZIP codes  
  (e.g. delivery method range **39770–39775** and contact ZIP **39772**).

All tests indicate the module works correctly.

**LGTM!**

ValentinVinagre (approved):
LGTM 👍🏻

Anxo82:
@pedrobaeza or @jbaudoux could you please review this PR for merge when you have a moment?
Thanks.

jbaudoux (commented):
A reason for depending on stock_delivery and not just delivery?

Anxo82:
@jbaudoux I’ve reviewed the module and it appears to only depend on delivery. I couldn’t find any usage of stock_delivery features, so I think the dependency could be changed accordingly.
@pedrobaeza as the module was originally implemented by your team, please let me know if you see any reason to keep it.
Thanks.

pedrobaeza:
The module was done before the `delivery` split, so it's possible that there's no reason to keep the other dependency, but you have to check if needed.

Anxo82:
@jbaudoux @pedrobaeza I’ve reviewed the module and updated the dependencies to only include delivery, as I couldn’t find any usage of features provided by stock_delivery.
Thanks.

pedrobaeza:
If @jbaudoux doesn't have any other comment, we can merge.

PR #1124, opened 2026-02-17
[18.0][IMP] delivery_state: Add logic to notify when retrieving the POD fails.

carlos-lopez-tecnativa:
Each delivery_* module must fill the new field when necessary.

TT59692

### DIFF
--- a/delivery_state/models/stock_picking.py
+++ b/delivery_state/models/stock_picking.py
@@ -3,7 +3,9 @@
 # Copyright 2020 Tecnativa - David Vidal
 # Copyright 2022 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-from odoo import api, fields, models
+from markupsafe import Markup
+
+from odoo import SUPERUSER_ID, api, fields, models
 
 
 class StockPicking(models.Model):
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I think this should be notified to the administrators channel instead.

carlos-lopez-tecnativa:
> I think this should be notified to the administrators channel instead.

Sorry, but I do not understand what you mean. Which channel are you referring to by `administrators channel`?

pedrobaeza:
It's something new starting with 18:

imagen

It's intended for notifying things like this to administrators.

carlos-lopez-tecnativa:
> It's something new starting with 18:
> It's intended for notifying things like this to administrators.

Done

delivery_state/models/stock_picking.py
```diff
@@ -86,6 +97,29 @@ def _update_delivery_state(self):
...
+    def _build_message_pod_error(self):
+        message = self.env._(
+            "<b>Errors while fetching POD for carrier %s:</b><br/>"
+            "Please review the details below "
+            "and take the necessary actions to resolve these issues.:<br/>",
+            self.carrier_id.name,
+        )
+        message += "<ul>"
+        for picking in self:
+            picking_url = f"<a href='/odoo/deliveries/{picking.id}'>{picking.name}</a>"
```
pedrobaeza: AFAIK, there's a method for obtaining this link directly without hardcoding things.

carlos-lopez-tecnativa: Done, thanks

pedrobaeza (approved):

sergio-teruel (approved):

PR #1113, opened 2026-02-10
[18.0][IMP] delivery_ups_oca: More features

sudhir-erpharbor:
- Added Insurance & Unit Test Cases
- Added API Logging
- Added Paperless Invoice & Unit Test Cases
- Added Residential Address Logic & Unit Test Cases
- Phone Number is Not Mandatory & Related Test Cases
- Added Negotiated Rates & Unit Test Cases
- Added functionality to get the shipping label in PDF format and also attach it in Chatter as an internal note
- Hide unavailable shipping methods
- Retry authentication failures

PR #1127, opened 2026-02-27
[18.0][IMP] delivery_ups_oca: Convert label from GIF to PDF

carlos-lopez-tecnativa:
UPS can return the label in GIF format instead of PDF. When trying to print the image directly from the attachment using the base_report_to_printer module, it is not possible because GIF is not supported. In this case, convert the label from GIF to PDF.

TT58440
@Tecnativa @pedrobaeza @carlosdauden @sergio-teruel could you please review this?

### DIFF
--- a/delivery_ups_oca/models/delivery_carrier.py
+++ b/delivery_ups_oca/models/delivery_carrier.py
@@ -4,6 +4,11 @@
 # Copyright 2024 Sygel - Manuel Regidor
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+import base64
+from io import BytesIO
+
+from PIL import Image
+
 from odoo import fields, models
 
 from .ups_request import UpsRequest
@@ -19,7 +24,13 @@ class DeliveryCarrier(models.Model):
         },
     )
     ups_file_format = fields.Selection(
-        selection=[("GIF", "GIF"), ("ZPL", "ZPL"), ("EPL", "EPL"), ("SPL", "SPL")],
+        selection=[
+            ("PDF", "PDF"),
+            ("GIF", "GIF"),
+            ("ZPL", "ZPL"),
+            ("EPL", "EPL"),
+            ("SPL", "SPL"),
+        ],
         default="GIF",
         string="File format",
     )
@@ -156,7 +167,14 @@ class DeliveryCarrier(models.Model):
         val_list = []
         for label in labels:
             format_code = label["format_code"].upper()
-            attachment_name = f"{label['tracking_ref']}-{format_code}.{format_code}"
+            # When the label format is PDF(GIF), convert it to PDF.
+            if format_code == "GIF" and self.ups_file_format == "PDF":
+                label["datas"] = self._convert_gif_to_pdf(label["datas"])
+                label["format_code"] = "PDF"
+                format_code = "PDF"
+            attachment_name = (
+                f"{label['tracking_ref']}-{format_code}.{format_code.lower()}"
+            )
             val_list.append(
                 self._prepare_ups_label_attachment(
                     picking,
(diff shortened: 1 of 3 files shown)


carlos-lopez-tecnativa:
I noticed that this PR https://github.com/OCA/delivery-carrier/pull/1113 adds a similar feature, but with more changes. In the future, they could be unified. @sudhir-erpharbor, could you please review this PR as well while your PR is being finalized?

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -156,6 +161,16 @@ def _create_ups_label(self, picking, labels):
...
+            # When the label format is GIF, convert it to PDF.
+            if format_code == "GIF":
+                img_decoded = base64.b64decode(label["datas"])
+                image_string = BytesIO(img_decoded)
+                im = Image.open(image_string)
+                label_result = BytesIO()
+                im.save(label_result, "PDF")
+                label["datas"] = base64.b64encode(label_result.getvalue()).decode()
+                label["format_code"] = "PDF"
+                format_code = "PDF"
```
pedrobaeza, line 174: No need to do these 2 variable assignment, as it's not being used further.

```suggestion
```

carlos-lopez-tecnativa: > No need to do these 2 variable assignment, as it's not being used further.

Are used in the `attachment_name`

pedrobaeza: OK, but only the second one. And taking the occasion, I don't usually like extensions in upper, so it may be changed.

carlos-lopez-tecnativa: If I do not update `label["format_code"]`, it may have an incorrect value if any code reads this value from there. To keep consistency, I prefer to change it.

I also updated the code to save the file extension in lowercase.

pedrobaeza: OK.

pedrobaeza:
Check CI.

christian-ramos-tecnativa:
We are assuming here that no one is using the GIF option (and may be true), but could be better to add a new option `PDF` that internaly transform the GIF into PDF. WDYT

pedrobaeza:
Good idea. We can add an entry in the dictionary.

carlos-lopez-tecnativa:
> We are assuming here that no one is using the GIF option (and may be true), but could be better to add a new option `PDF` that internaly transform the GIF into PDF. WDYT

Done

christian-ramos-tecnativa (approved):
LGTM, tests errors doesn't seems related

TheRealRoyalTS (commented):
Thanks for the contribution, this is a useful improvement!

Two small things before merge:

1. **Missing `resolution` in `im.save()`** — `im.save(label_result, "PDF")` defaults to 72 DPI. UPS GIF labels are 828×1164 px designed for a 6×4" label at 236 DPI, so the printed PDF will come out at the wrong physical size. Fix: `im.save(label_result, "PDF", resolution=236.0)`

2. **Pillow not declared in `__manifest__.py`** — Since Pillow is now a hard runtime dependency it should be listed so Odoo can warn on missing packages:
   ```python
   "external_dependencies": {"python": ["Pillow"]},
   ```

pedrobaeza:
Pillow is already declared in standard Odoo: https://github.com/odoo/odoo/blob/18.0/requirements.txt#L50-L53

TheRealRoyalTS:
> Pillow is already declared in standard Odoo: https://github.com/odoo/odoo/blob/18.0/requirements.txt#L50-L53

You're right!
I was comparing with PR #1113 where Pillow is explicitly listed in external_dependencies in the manifest.

carlos-lopez-tecnativa:
> 1. **Missing `resolution` in `im.save()`** — `im.save(label_result, "PDF")` defaults to 72 DPI. UPS GIF labels are 828×1164 px designed for a 6×4" label at 236 DPI, so the printed PDF will come out at the wrong physical size. Fix: `im.save(label_result, "PDF", resolution=236.0)`

Done, please update your review

TheRealRoyalTS (approved):
LGTM

pedrobaeza (approved):
/ocabot merge minor

PR #1134, opened 2026-03-14
[18.0][IMP] delivery_ups_oca: Sanitize phone/mobile before send data to carrier

sergio-teruel:
If the phone number is stored as +34 122 12 12 12, it is sent to the carrier in this format, which may cause an error due to the number being too long.
The formatting from the phone_validation module (without adding a direct dependency) is used in E164 format.

cc @Tecnativa

ping @carlosdauden @carlos-lopez-tecnativa

### DIFF
--- a/delivery_ups_oca/models/delivery_carrier.py
+++ b/delivery_ups_oca/models/delivery_carrier.py
@@ -3,8 +3,9 @@
 # Copyright 2023 ForgeFlow, S.L. - Jordi Ballester
 # Copyright 2024 Sygel - Manuel Regidor
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+import re
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 from .ups_request import UpsRequest
 
@@ -209,3 +210,24 @@ class DeliveryCarrier(models.Model):
         self.ensure_one()
         ups_request = UpsRequest(self)
         ups_request._get_new_token()
+
+    @api.model
+    def sanitize_phone_ups(self, phone_number):
+        """Sanitize a phone number for UPS usage."""
+        if not phone_number:
+            return phone_number
+        sanitize_fn = getattr(self, "_phone_format", None)
+        if sanitize_fn:
+            phone_sanitized = sanitize_fn(number=phone_number)
+            if phone_sanitized:
+                return phone_sanitized
+        if phone_number:
+            phone_number = re.sub(r"[^\d+]", "", phone_number)
+            # Clean plus character is there is any in the number in othe positions.
+            # We only want the first plus
+            phone_number = (
+                "+" + phone_number.replace("+", "")
+                if "+" in phone_number
+                else phone_number
+            )
+        return phone_number
--- a/delivery_ups_oca/models/ups_request.py
+++ b/delivery_ups_oca/models/ups_request.py
@@ -139,7 +139,9 @@ class UpsRequest:
             Name=(partner.parent_id or partner).name,
             AttentionName=partner.name,
             TaxIdentificationNumber=partner.vat,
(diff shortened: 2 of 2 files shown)


victoralmau (approved):

carlos-lopez-tecnativa (changes requested):
I’m not sure about this change. Please take into account that the method [_phone_format_number ](https://github.com/odoo/odoo/blob/1b5072c0e0af6be340389e0c429ca370d8dc169d/addons/phone_validation/models/models.py#L111) returns False if validation is not possible (for example, when the number is invalid).


Before this change, with this phone number, the data was sent correctly to UPS.
Image


With this commit, I get this error.
Image

Image

I understand that the phone number +34 122 12 12 12 is invalid, but my point is that I now have different behavior depending on whether the phone is valid or not. The goal of this PR is to handle cases where the phone number has more than 15 characters.

sergio-teruel:
Change done! thanks @carlos-lopez-tecnativa

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -209,3 +209,18 @@ def ups_update_token(self):
...
+        Default format is E164
+        """
+        sanitize_fn = getattr(self, "_phone_format", None)
+        if phone_number and sanitize_fn:
+            phone_sanitized = sanitize_fn(number=phone_number)
+            if phone_sanitized:
+                phone_number = phone_sanitized
+            else:
+                # At least remove any spaces if the phone number validation fails
+                phone_number = phone_number.replace(" ", "")
```
carlosdauden: Can you replace "()-" too in this point?

carlosdauden (approved):
Minor comment

sergio-teruel:
Better clean phone number...

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -209,3 +210,24 @@ def ups_update_token(self):
...
         ups_request = UpsRequest(self)
         ups_request._get_new_token()
+
+    @api.model
+    def sanitize_phone_ups(self, phone_number):
+        """Sanitize a phone number for UPS usage."""
+        if not phone_number:
+            return phone_number
+        sanitize_fn = getattr(self, "_phone_format", None)
+        if sanitize_fn and sanitize_fn:
```
carlos-lopez-tecnativa:
```suggestion
        if sanitize_fn:
```

carlos-lopez-tecnativa (changes requested):

sergio-teruel:
Changed @carlos-lopez-tecnativa

sergio-teruel:
ping @carlos-lopez-tecnativa

carlos-lopez-tecnativa (approved):

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -209,3 +210,24 @@ def ups_update_token(self):
         self.ensure_one()
         ups_request = UpsRequest(self)
         ups_request._get_new_token()
+
+    @api.model
+    def sanitize_phone_ups(self, phone_number):
```
pedrobaeza, line 215: Why is this put here instead of ups_request and why it's public?

sergio-teruel:
>> Why is this put here instead of ups_request
Because I am accessing a method in the `models` module without the module I depend on being installed

>> why it's public
Because I access it from another class

pedrobaeza:
On Odoo ORM, public means callable from RPC, which is what should be avoided, but you can call from other model any method, being prefixed with _ or not. And anyway, I don't get why putting this method in this crowded model of it's only called from the other one.

sergio-teruel:
Today is my final contribution to OCA, thanks to people like this... 

cc @Tecnativa

PR #1135, opened 2026-03-16
[14.0][IMP] delivery_send_to_shipper_at_operation: Prevent Duplicate Trackings

raumschmiede-joshuaL:
In some cases the carrier API can return the same tracking number for a PACK backorder that is already on a done PACK picking

### DIFF
--- a/delivery_send_to_shipper_at_operation/models/stock_picking.py
+++ b/delivery_send_to_shipper_at_operation/models/stock_picking.py
@@ -1,6 +1,7 @@
 # Copyright 2021 Camptocamp SA
 # Copyright 2023 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # Copyright 2024 Michael Tietz (MT Software) <mtietz@mt-software.de>
+# Copyright 2026 Raumschmiede GmbH
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
 from lxml import etree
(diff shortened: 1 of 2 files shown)


raumschmiede-joshuaL:
@mt-software-de, @sebalix and @jbaudoux, can you check?

vuwnevska (approved):
apply pre-commit, LG

raumschmiede-joshuaL:
> apply pre-commit, LG

@vuwnevska, already done. The check fails because of a missing module. Something is broken in the pre-commit setup

raumschmiede-joshuaL:
pre-commit check will be fixed after PR https://github.com/OCA/delivery-carrier/pull/1141 was merged

delivery_send_to_shipper_at_operation/models/stock_picking.py
```diff
@@ -68,7 +69,8 @@ def _prepare_values_send_to_ship_at_operation(self, related_ship):
         related_ship.ensure_one()
         carrier_tracking_ref = related_ship.carrier_tracking_ref
         if carrier_tracking_ref:
-            carrier_tracking_ref += "," + self.carrier_tracking_ref
+            if self.carrier_tracking_ref not in carrier_tracking_ref.split(","):
+                carrier_tracking_ref += "," + self.carrier_tracking_ref
```
mt-software-de, line 73: IMO. This should not be the case, that a tracking ref is the same for multiple transfers. 
That's due to a specific implementation on your end, but I don't think your change will cause any problems either. So why not just. 


```suggestion
            if self.carrier_tracking_ref not in carrier_tracking_ref:
                carrier_tracking_ref += "," + self.carrier_tracking_ref
```

mt-software-de (changes requested):

delivery_send_to_shipper_at_operation/models/stock_picking.py
raumschmiede-joshuaL: If an API returned tracking ref "ABC" and later for the backorder ref "AB", it thinks "AB" was already added to the OUT picking. Eventhough only "ABC" was added.
Will never happen but just in case. I can change it if you want

mt-software-de: ok.

mt-software-de (approved):
LGTM.

@OCA/logistics-maintainers can someone merge this?

hparfr (commented):
/ocabot merge minor

PR #1137, opened 2026-03-17
[16.0][FIX] delivery_roulier_laposte_fr:  crash in _laposte_fr_get_parcel when done quantity is 0

oujbih:
The product_qty field does not exist on stock.move.line in 16.0 so we removed the corresponding code.


hparfr (approved):

dreispt (approved):

oujbih:
replace with : https://github.com/OCA/delivery-carrier/pull/1168

PR #1138, opened 2026-03-18
[19.0][ADD] delivery_dpd_portugal

dreispt:


delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+TEST_BASE_URL = "https://qabusiness.dpd.pt/api/v1/pt"
+PROD_BASE_URL = "https://business.dpd.pt/api/v1/pt"
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = "delivery.carrier"
+
+    delivery_type = fields.Selection(
+        selection_add=[("dpd_portugal", "DPD Portugal")],
+        ondelete={"dpd_portugal": "set default"},
```
raumschmiede-joshuaL, line 22: Personally I like the country code more than the the country name:


```suggestion
        selection_add=[("dpd_pt", "DPD Portugal")],
        ondelete={"dpd_pt": "set default"},
```

Less to write on each new field and method.

The name of the module could be changed too, but not necessary. Not everyone knows the country code. But if you are inside the module, you will understand that PT = Portugal

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+    )
+
+    # DPD Portugal API Configuration
+    dpd_portugal_prod_environment = fields.Boolean(
+        string="Production Environment",
+        default=False,
+        help="Use production API instead of test environment",
+    )
```
raumschmiede-joshuaL, line 30: I think you can replace this with standard field `prod_environment`. Your field sounds like as it does the exact same

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+            if carrier.delivery_type == "dpd_portugal":
+                carrier.supports_shipping_insurance = (
+                    carrier.dpd_portugal_insurance_enabled
+                )
+        return result
+
+    def dpd_portugal_rate_shipment(self, order):
+        """Get shipping rates from DPD Portugal."""
+        if self.delivery_type != "dpd_portugal":
+            return super().rate_shipment(order)
```
raumschmiede-joshuaL, line 121:
```suggestion
```

`dpd_portugal_rate_shipment` is called inside of `rate_shipment`:
https://github.com/odoo/odoo/blob/35bc748256de1b5eac81007a62d43dc663726ab6/addons/delivery/models/delivery_carrier.py#L322
It is a carrier-specific implementation. `dpd_portugal_rate_shipment` will always be called only for carriers that have `dpd_portugal` as `delivery_type`. So the check is not necessary either

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+                ),
+                "warning_message": False,
+            }
+
+    def dpd_portugal_send_shipping(self, pickings):
+        """Send shipment to DPD Portugal and generate labels."""
+        if self.delivery_type != "dpd_portugal":
+            return super().send_shipping(pickings)
+
+        self.ensure_one()
```
raumschmiede-joshuaL, line 164: Same as with `rate_shipment`.
`ensure_one` is done by `send_shipping`, no need to do it here again:
https://github.com/odoo/odoo/blob/35bc748256de1b5eac81007a62d43dc663726ab6/addons/stock_delivery/models/stock_picking.py#L126

```suggestion
```

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+                        "type": "binary",
+                        "datas": base64.b64encode(label_content),
+                        "res_model": "stock.picking",
+                        "res_id": picking.id,
+                    }
+                )
+                picking.message_post(
+                    body=self.env._("DPD Portugal label generated"),
+                    attachment_ids=[attachment.id],
+                )
```
raumschmiede-joshuaL, line 208: Something I learned recently: `message_post` can create attachments on its own. Here in delivery_schenker it is implemented like this:
https://github.com/OCA/delivery-carrier/blob/2630fc80634d4806ee4903fa805abe039a77de5d/delivery_schenker/models/delivery_carrier.py#L467-L473

If you want you can try it. Less code to write.

You are using the attachment later. Maybe `message_post` somehow returns the ID of the created attachment. If not, keep your code

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+
+            # Prepare shipment data
+            shipment_data = self._prepare_dpd_portugal_shipment_data(picking)
+
+            # Create shipment using account-based request
+            shipment_response = self._dpd_portugal_request_with_account(
+                "shipment/create", shipment_data, picking
+            ).json()
+
+            if shipment_response.get("success", False):
```
raumschmiede-joshuaL, line 179: With the opposite check you can move the `UserError` to the top and you indent the code one to the left as you can remove the `else`

```suggestion
            if not shipment_response.get("success"):
```

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+                        "res_id": picking.id,
+                    }
+                )
+                picking.message_post(
+                    body=self.env._("DPD Portugal label generated"),
+                    attachment_ids=[attachment.id],
+                )
+
+                # Send email if enabled
+                if self.dpd_portugal_send_label_email and picking.partner_id.email:
```
raumschmiede-joshuaL, line 211: Same here. Check for the opposite and do a `continue`. Move the result.append from line 226 above line 211

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+        :return: tracking URL or False
+        """
+        if picking.carrier_tracking_ref:
+            return f"https://www.dpd.pt/rastreamento?parcelNumber={picking.carrier_tracking_ref}"
+        return False
+
+    def dpd_portugal_tracking_state_update(self, picking):
+        """Update tracking state from DPD Portugal."""
+        if self.delivery_type != "dpd_portugal" or not picking.carrier_tracking_ref:
+            return super().tracking_state_update(picking)
```
raumschmiede-joshuaL, line 255: Same as `rate_shipment`. No check and no super call needed

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+
+        tracking_data = self._dpd_portugal_request_with_account(
+            f"shipment/track/{picking.carrier_tracking_ref}", {}, picking
+        ).json()
+
+        if "status" in tracking_data:
+            # Update picking tracking state
+            picking.tracking_state = tracking_data["status"]
+            picking.tracking_state_history = [
```
raumschmiede-joshuaL, line 264: `tracking_state_history` is a text field from `delivery_state`. You are missing the dependency to this module. `delivery_schenker` as reference: 
https://github.com/OCA/delivery-carrier/blob/2630fc80634d4806ee4903fa805abe039a77de5d/delivery_schenker/models/delivery_carrier.py#L542-L556


Please add a unit test for this

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+                )
+                picking.carrier_tracking_ref = False
+            else:
+                raise UserError(
+                    self.env._(
+                        "Failed to cancel shipment: %(message)s",
+                        message=response.get("message", "Unknown error"),
+                    )
+                )
+
```
raumschmiede-joshuaL, line 312: Same as in line 179. Revert the `if` to remove the `else`:

```suggestion
            if not response.get("success"):
                raise UserError(
                    self.env._(
                        "Failed to cancel shipment: %(message)s",
                        message=response.get("message", "Unknown error"),
                    )
                )

            picking.message_post(
                body=self.env._(
                    "DPD Portugal shipment %(tracking_ref)s cancelled",
                    tracking_ref=picking.carrier_tracking_ref,
                )
            )
            picking.carrier_tracking_ref = False

```

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+    def _prepare_dpd_portugal_shipment_data(self, source):
```
raumschmiede-joshuaL, line 313: IMO it is better to always put the carrier-specific part of the name to the front:

```suggestion
    def _dpd_portugal_prepare_shipment_data(self, source):
```

With that all defs look the same and you know you can always type `_dpd_portual_` first to get the right name.

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+        """Prepare shipment data for DPD Portugal API."""
+        # Get sender and recipient information
+        if source.partner_id and source.warehouse_id:
+            # Stock picking
+            sender = source.warehouse_id.partner_id or source.company_id.partner_id
+            recipient = source.partner_id
+            weight = source.shipping_weight
+        elif source.partner_shipping_id:
```
raumschmiede-joshuaL, line 321: This will raise an error if `warehouse_id` is not set on the picking as it does not have the field `partner_shipping_id`.
And I can't find the code that defines `warehouse_id` for `stock.picking`.

It is better to have a `_dpd_portugal_prepare_shipment_data` that accepts sender and recipient and param, which is called by `_dpd_portugal_prepare_shipment_data_sale_order` and `_dpd_portugal_prepare_shipment_data_picking` that pass the right data to it.

Same goes with other values like name and origin in lines 383 and 384. Make a base def that accepts these values as param,
or prepare the result without these values. In `_dpd_portugal_prepare_shipment_data_sale_order` get the result of the base def and set the values in there

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+            sender = source.company_id.partner_id
+            recipient = source.partner_shipping_id
+            weight = source.total_weight
+        else:
+            raise ValidationError(
+                self.env._("Invalid source for shipment data preparation")
+            )
+
+        # Get package dimensions
+        # In Odoo 19, packages are accessed via move_line_ids.result_package_id
```
raumschmiede-joshuaL, line 332: New to me. But comment not needed

```suggestion
```

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+        packages = source.move_line_ids.result_package_id
```
raumschmiede-joshuaL, line 333: This line should fail when you pass a sales order as param. That's why a base def is better

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+        if packages:
+            # Use first package dimensions
+            package = packages[0]
+            dimensions = {
+                "length": package.length or 10,
+                "width": package.width or 10,
+                "height": package.height or 10,
+            }
```
raumschmiede-joshuaL, line 341: Weird that the API only accepts one package. But I think if you have multiple packages in Odoo, you should sum up the dimensions of each. If the sum is 0, use the default dimensions in ilne 352. 
If DPD Portugal is fine with the default dimensions

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+                "insurance": {
+                    "enabled": self.dpd_portugal_insurance_enabled,
+                    "amount": self.dpd_portugal_insurance_amount,
+                },
+            },
+        }
+
+        return shipment_data
+
+    def _validate_dpd_portugal_picking(self, picking):
```
raumschmiede-joshuaL, line 403: Check for fields is not necessary. It seems that you expect that another module needs to be installed that adds `warehouse_id` to the picking as you rely on it. But you have no dependency to it. You can use `picking.picking_type_id.warehouse_id`.
Check on missing picking partner can be done where you call this def

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+                )
+            )
+
+        # Validate addresses
+        self._validate_dpd_portugal_address(
+            picking.warehouse_id.partner_id or picking.company_id.partner_id, "sender"
+        )
+        self._validate_dpd_portugal_address(picking.partner_id, "recipient")
+
+    def _validate_dpd_portugal_address(self, partner, address_type):
```
raumschmiede-joshuaL, line 424: How does the API behave if one of the fields you check in this def is not set? Will the API raise an error or will it accept wrong data? Can you test this?
Usually it should check the fields, that's why this def is not needed

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+                self.env._(
+                    "Missing required fields for %(address_type)s address: %(fields)s",
+                    address_type=address_type,
+                    fields=", ".join(missing_fields),
+                )
+            )
+
+    def _dpd_portugal_get_api_url(self):
+        """Get API URL based on environment."""
+        return PROD_BASE_URL if self.dpd_portugal_prod_environment else TEST_BASE_URL
```
raumschmiede-joshuaL, line 442: As you need the URLs only in this def, remove the constants and add the strings here


```suggestion
        return "https://business.dpd.pt/api/v1/pt" if self.dpd_portugal_prod_environment else "https://qabusiness.dpd.pt/api/v1/pt"
```

delivery_dpd_portugal/models/delivery_carrier.py
```diff
@@ -0,0 +1,506 @@
...
+                auth=auth,
+                timeout=30,
+            )
+        except requests.exceptions.RequestException as exc:
+            raise UserError(
+                self.env._("DPD Portugal connection error: %(error)s", error=exc)
+            ) from exc
+
+        if response.status_code not in (200, 201):
+            error_msg = self._dpd_portugal_extract_error_message(response)
```
raumschmiede-joshuaL, line 482: It's easier to remove `_dpd_portugal_extract_error_message` and to use `response.text` here. Doesn't matter whether it is a JSON or not

delivery_dpd_portugal/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1,8 @@
+- [Open Source Integrators](https://www.opensourceintegrators.com):
+  - Daniel Reis
+
+---
+
+**Note:** This module was developed with the assistance of AI language models,
```
raumschmiede-joshuaL, line 6: Interesting. This explains a few things

delivery_dpd_portugal/tests/test_delivery_dpd_portugal.py
```diff
@@ -0,0 +1,127 @@
...
+                "product_id": self.env.ref("delivery.product_product_delivery").id,
+            }
+        )
+
+    def test_carrier_creation(self):
+        """Test DPD Portugal carrier creation."""
+        self.assertEqual(self.dpd_carrier.delivery_type, "dpd_portugal")
+        self.assertEqual(self.dpd_carrier.dpd_portugal_service_type, "standard")
+        self.assertFalse(self.dpd_carrier.dpd_portugal_prod_environment)
+        self.assertTrue(self.dpd_carrier.dpd_portugal_cod_enabled)
```
raumschmiede-joshuaL, line 59: This test is useless. All fields are set in the `setUp`, no need to check them

delivery_dpd_portugal/tests/test_delivery_dpd_portugal.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def test_can_generate_return(self):
+        """Test return shipment capability."""
+        self.assertTrue(self.dpd_carrier.can_generate_return)
+
+    def test_supports_shipping_insurance(self):
+        """Test insurance support."""
+        self.assertTrue(self.dpd_carrier.supports_shipping_insurance)
```
raumschmiede-joshuaL, line 67: This is useless. The check below is fine

delivery_dpd_portugal/tests/test_delivery_dpd_portugal.py
```diff
@@ -0,0 +1,127 @@
...
+        self.dpd_carrier._validate_dpd_portugal_address(self.customer, "recipient")
+
+        # Invalid address (missing required fields)
+        invalid_partner = self.env["res.partner"].create({"name": "Invalid Partner"})
+        with self.assertRaises(ValidationError):
+            self.dpd_carrier._validate_dpd_portugal_address(
+                invalid_partner, "recipient"
+            )
+
+    def test_api_url_selection(self):
```
raumschmiede-joshuaL, line 102: Not really useful. Remove it

delivery_dpd_portugal/tests/test_delivery_dpd_portugal.py
```diff
@@ -0,0 +1,127 @@
...
+                "partner_id": self.customer.id,
+                "picking_type_id": self.env.ref("stock.picking_type_out").id,
+                "location_id": self.env.ref("stock.stock_location_stock").id,
+                "location_dest_id": self.env.ref("stock.stock_location_customers").id,
+                "carrier_id": self.dpd_carrier.id,
+                "dpd_portugal_cod_amount": 150.0,
+            }
+        )
+
+        self.assertEqual(picking.dpd_portugal_cod_amount, 150.0)
```
raumschmiede-joshuaL, line 127: The field is not computed, it is set a few line above it. Useless check, remove it

delivery_dpd_portugal/tests/test_delivery_dpd_portugal.py
```diff
@@ -0,0 +1,127 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+from odoo.exceptions import ValidationError
+from odoo.tests import TransactionCase, tagged
+
+
+@tagged("post_install", "-at_install")
+class TestDeliveryDPDPortugal(TransactionCase):
```
raumschmiede-joshuaL, line 9: There are no unit tests that check the actual carrier API functionality like tracking state and shipment creation.
You can mock `requests.post` to prevent that the API is called, you can modify the result to raise HTTP errors and you can check the JSON data that was passed to `post`, so the shipment data and tracking data.

Please add tests to cover these cases. Try the API, check in which cases the API returns an error and re-create this behaviour with the mocked `post` to test how the module behaves.

With these tests everyone of us can execute the tests to know whether the main part of your implementation actually works as we do not need credentials for the API

delivery_dpd_portugal/views/delivery_carrier.xml
```diff
@@ -0,0 +1,60 @@
...
+    </record>
+
+    <!-- DPD Portugal Carrier List View -->
+    <record id="view_delivery_carrier_list_dpd_portugal" model="ir.ui.view">
+        <field name="name">delivery.carrier.list.dpd_portugal</field>
+        <field name="model">delivery.carrier</field>
+        <field name="inherit_id" ref="delivery.view_delivery_carrier_tree" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='delivery_type']" position="after">
+                <field name="dpd_portugal_service_type" optional="show" />
```
raumschmiede-joshuaL, line 56: Why do you want to show the field in the carrier list view? Remove it or make it `optional="hide"`

raumschmiede-joshuaL (changes requested):

dreispt:
@raumschmiede-joshuaL Thank you for the thorough review.
Changes were made based on it.
Unfortunately I won't be able to add unit tests that check the actual carrier API functionality at this point.

delivery_dpd_portugal/migrations/19.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,37 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.tools import sql
+
+
+def migrate(cr, version):
```
raumschmiede-joshuaL, line 6: The migration is needed only if the module is already in the OCA repo and some people could already have it installed.
Do you have the module installed on a production database of you or one of your customers? If yes, I think it is ok to keep the migration script. But in this case I don't know whether it is ok to add a migration script

delivery_dpd_portugal/views/delivery_carrier.xml
```diff
@@ -0,0 +1,48 @@
...
+            <xpath expr="//page[@name='destination']" position="before">
+                <page
+                    string="DPD Portugal Configuration"
+                    name="dpd_portugal_configuration"
+                    invisible="delivery_type != 'dpd_pt'"
+                >
+                    <group>
+                        <group string="Configuration">
+                            <field name="prod_environment" />
+                        </group>
```
raumschmiede-joshuaL, line 18: Can be removed as the field is already displayed by Odoo

delivery_dpd_portugal/models/delivery_carrier.py
raumschmiede-joshuaL: I implemented another carrier recently. Adding the dimensions of each package does not make sense. A package with dimensions 10x10x10 cm and another with 20x20x20 cm will not result in a package that is 30x30x30 cm big. Is more complex.

Can you check the API again? I highly doubt that it works like this. Or they expect for each package a separate API call

raumschmiede-joshuaL (changes requested):
There are still changes left that I requested. Can you add comments to my review issues to know why you didn't change them?


> Unfortunately I won't be able to add unit tests that check the actual carrier API functionality at this point.

Why not? The most important code of the module does not seem to work, I assume only by reviewing it

raumschmiede-joshuaL:
@dreispt, what modules did you install locally or on a test system to test the module in the frontend? Did you test every method that calls the DPD API? I counted 5 places where the API is called.

Your changes will not work with the Odoo base and the OCA modules, seems that you had modules that changed the behaviour.

And it looks like that a part of your new changes were made by AI, like parts of the code before. Please stop this or at least review the AI-made code more precisely

PR #1140, opened 2026-03-20
[14.0][FIX] delivery_schenker: Use BID for Tracking State Update

raumschmiede-joshuaL:
The fetched tracking number is the Schenker booking ID. BID is the corresponding code in the tracking API.

With code cu the API returns nothing since March 2024. With BID it returns data for recent pickings in our Odoo.

Documentation: 
image

Can't really tell the difference, the documentation has no more description for it.

 
**EDIT**
I asked Schenker via mail. Got an answer. They say BID is better. cu could be used differently depending on country and product and can change over time

### DIFF
--- a/delivery_schenker/models/delivery_carrier.py
+++ b/delivery_schenker/models/delivery_carrier.py
@@ -1,5 +1,6 @@
 # Copyright 2021 Tecnativa - David Vidal
 # Copyright 2023 Michael Tietz (MT Software) <mtietz@mt-software.de>
+# Copyright 2026 Raumschmiede GmbH
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from lxml import etree
 
@@ -622,7 +623,7 @@ class DeliveryCarrier(models.Model):
         self.ensure_one()
         return {
             "reference": picking.carrier_tracking_ref,
-            "reference_type": "cu",
+            "reference_type": "BID",
             "booking_type": self.schenker_booking_type,
         }
 
--- a/delivery_schenker/models/schenker_request.py
+++ b/delivery_schenker/models/schenker_request.py
@@ -183,7 +183,7 @@ class SchenkerRequest:
         return {"AccessKey": self.access_key, "in": {}}
 
     def _get_tracking_states(
-        self, reference=False, reference_type="cu", booking_type="land"
+        self, reference=False, reference_type="BID", booking_type="land"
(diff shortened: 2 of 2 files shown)


raumschmiede-joshuaL:
@mt-software-de and @hparfr, can you check?

mt-software-de:
Thx. Could you please update the repo template?
https://github.com/OCA/oca-addons-repo-template?tab=readme-ov-file#how-to-use

mt-software-de (approved):
LGTM.

Could you add a description whats the difference between cu and BID.
Maybe also the screenshot of the schenker api documentation to this PR description?

raumschmiede-joshuaL:
I'm testing the API a bit. If the API does not return anything for a picking, it would be good to set the `delivery_state`  to `no_update`. Then all old pickings for which the state is still fetched are fixed. New pickings, that even do not have a final state in the Schenker portal will be set to done after the API stopped returning a new status for them.
Will add a commit for this soon

raumschmiede-joshuaL:
@mt-software-de, I moved the repo update into PR https://github.com/OCA/delivery-carrier/pull/1141 so that it can be merged faster

mt-software-de:
How did you know you had to change it to BID? The documentation doesn't really say much about that.

raumschmiede-joshuaL:
> How did you know you had to change it to BID? The documentation doesn't really say much about that.

I tested the API.
Today I tested all other available codes. Only BID returns something.

But the API returns for all codes nothing if the shipment booking has not been submitted yet in the Schenker portal, so when flag `schenker_submit_booking` is not set on the carrier. I guess the first tracking status will be available as soon as the shipment has been submitted. I will test it more to be sure.
A further check is probably needed in the tracking fetching to cover this. Something like if picking was set to done more than 20 days ago and still no tracking status via API, only then set the `delivery_state` to `no_update`.

Nevertheless code BID makes more sense, regardless of the submit flag

raumschmiede-joshuaL:
@mt-software-de, I added a migration to set the delivery state on the pickings that will most likely get no more tracking state updates.

**EDIT**
Deleted it. Will add another PR to set a day limit to stop fetching a tracking state for pickings when there is no new or none from the API

mt-software-de (approved):
LGTM. Thanks for the explanation.

pedrobaeza (approved):
/ocabot merge patch

PR #1148, opened 2026-03-30
[14.0][IMP] delivery_state: Exclude Pickings without Delivery State

raumschmiede-joshuaL:
In our system we have `delivery_send_to_shipper_at_operation` and `deliver_schenker` installed. We create the Schenker labels on the PACK step. When the PACK picking creates a backorder and we send the backorder to the schenker API, its tracking number is added to the OUT picking next to tracking number of the first PACK picking. Then we set the OUT picking to done.

The cron of `delivery_state` finds all 3 pickings and tries to fetch the state. For the 2 PACK pickings it is fine, but the API call for the OUT picking fails. This is the log:

image

The API does not like 2 tracking numbers.

The cron should only fetch the tracking state for pickings that were sent to the carrier API.
My fix ensures that `delivery_state` is compatible with `delivery_send_to_shipper_at_operation`. [Here](https://github.com/OCA/delivery-carrier/blob/fed707047d70833de0802c7ae03aaa8a092c04f0/delivery_send_to_shipper_at_operation/models/stock_picking.py#L87) the module skips the super call that would call later the `send_shipping` of `delivery_state` that would set `date_shipped` on the OUTs:
image

That's why checking `date_shipped` is enough to make it compatible.
**EDIT**
Or to check for other delivery states

### DIFF
--- a/delivery_state/models/stock_picking.py
+++ b/delivery_state/models/stock_picking.py
@@ -76,6 +76,14 @@ class StockPicking(models.Model):
         copy=False,
     )
 
+    def _get_delivery_states_in_progress(self):
+        return [
+            DELIVERY_STATE_SHIPPING_RECORDED,
+            DELIVERY_STATE_IN_TRANSIT,
+            DELIVERY_STATE_INCIDENCE,
+            DELIVERY_STATE_WH_DELIVERED,
+        ]
+
     def tracking_state_update(self):
         """Call to the service provider API which should have the method
         defined in the model as:
@@ -132,12 +140,8 @@ class StockPicking(models.Model):
                 ("state", "=", "done"),
                 (
                     "delivery_state",
-                    "not in",
-                    [
-                        DELIVERY_STATE_CUS_DELIVERED,
-                        DELIVERY_STATE_CANCELED,
-                        DELIVERY_STATE_NO_UPDATE,
-                    ],
+                    "in",
+                    self._get_delivery_states_in_progress(),
                 ),
                 # These won't ever autoupdate, so we don't want to evaluate them
                 ("delivery_type", "not in", [False, "fixed", "base_on_rule"]),
--- a/delivery_state/models/stock_picking.py
+++ b/delivery_state/models/stock_picking.py
@@ -9,6 +9,14 @@ from markupsafe import Markup
 
 from odoo import _, api, fields, models
 
+DELIVERY_STATE_SHIPPING_RECORDED = "shipping_recorded_in_carrier"
+DELIVERY_STATE_IN_TRANSIT = "in_transit"
+DELIVERY_STATE_CANCELED = "canceled_shipment"
+DELIVERY_STATE_INCIDENCE = "incidence"
+DELIVERY_STATE_CUS_DELIVERED = "customer_delivered"
+DELIVERY_STATE_WH_DELIVERED = "warehouse_delivered"
+DELIVERY_STATE_NO_UPDATE = "no_update"
+
 
 class StockPicking(models.Model):
     _inherit = "stock.picking"
@@ -39,13 +47,13 @@ class StockPicking(models.Model):
     )
     delivery_state = fields.Selection(
         selection=[
-            ("shipping_recorded_in_carrier", "Shipping recorded in carrier"),
-            ("in_transit", "In transit"),
-            ("canceled_shipment", "Canceled shipment"),
-            ("incidence", "Incidence"),
-            ("customer_delivered", "Customer delivered"),
-            ("warehouse_delivered", "Warehouse delivered"),
-            ("no_update", "No more updates from carrier"),
+            (DELIVERY_STATE_SHIPPING_RECORDED, "Shipping recorded in carrier"),
+            (DELIVERY_STATE_IN_TRANSIT, "In transit"),
+            (DELIVERY_STATE_CANCELED, "Canceled shipment"),
+            (DELIVERY_STATE_INCIDENCE, "Incidence"),
+            (DELIVERY_STATE_CUS_DELIVERED, "Customer delivered"),
+            (DELIVERY_STATE_WH_DELIVERED, "Warehouse delivered"),
+            (DELIVERY_STATE_NO_UPDATE, "No more updates from carrier"),
         ],
         string="Carrier State",
         tracking=True,
@@ -89,7 +97,11 @@ class StockPicking(models.Model):
             days = carrier.days_fetch_tracking_state_update
             if (
                 picking.delivery_state
-                in ["customer_delivered", "canceled_shipment", "no_update"]
+                in [
+                    DELIVERY_STATE_CUS_DELIVERED,
+                    DELIVERY_STATE_CANCELED,
+                    DELIVERY_STATE_NO_UPDATE,
+                ]
                 or days <= 0
             ):
                 continue
@@ -103,7 +115,7 @@ class StockPicking(models.Model):
                 date_tracking_started = picking.date_done
 
             if date_tracking_started <= datetime.now() - timedelta(days=days):
-                picking.delivery_state = "no_update"
+                picking.delivery_state = DELIVERY_STATE_NO_UPDATE
 
         # Filter pickings with errors and notify
         pickings_with_errors = self.filtered("pod_error")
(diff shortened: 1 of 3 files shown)


raumschmiede-joshuaL:
@chienandalu and @ValentinVinagre, can you check?

raumschmiede-joshuaL:
@mt-software-de

delivery_state/models/stock_picking.py
```diff
@@ -66,6 +67,8 @@ def _update_delivery_state(self):
                 ),
                 # These won't ever autoupdate, so we don't want to evaluate them
                 ("delivery_type", "not in", [False, "fixed", "base_on_rule"]),
+                # This is set only for the pickings that were sent to the carrier API
+                ("date_shipped", "!=", False),
```
raumschmiede-joshuaL: The other idea would be to change the check on `delivery_state`. It is not set on OUT pickings. Either by checking for non-final states:

```suggestion
"in",
["shipping_recorded_in_carrier", "in_transit", "incidence"],
),
# These won't ever autoupdate, so we don't want to evaluate them
("delivery_type", "not in", [False, "fixed", "base_on_rule"]),
```

Or by adding an empty string/False to the list:


```suggestion
"not in",
["", False, "customer_delivered", "canceled_shipment", "no_update"],
),
# These won't ever autoupdate, so we don't want to evaluate them
("delivery_type", "not in", [False, "fixed", "base_on_rule"]),
```

mt-software-de: Like we already discussed. IMO. It is not needed to check the date_shipped. If you change the domain to `in ["shipping_recorded_in_carrier", "in_transit", "incidence"]` it is working and it is more accurate.

raumschmiede-joshuaL:
@jbaudoux, can you check?

delivery_state/models/stock_picking.py
raumschmiede-joshuaL: Implemented it like you said

raumschmiede-joshuaL:
@mt-software-de, can you check again? I added a new commit

delivery_state/models/stock_picking.py
```diff
@@ -1,9 +1,18 @@
 # Copyright 2020 Trey, Kilobytes de Soluciones
 # Copyright 2020 FactorLibre
 # Copyright 2020 Tecnativa - David Vidal
+# Copyright 2026 Raumschmiede GmbH
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import api, fields, models
 
+DS_SHIPPING_RECORDED = "shipping_recorded_in_carrier"
```
mt-software-de: I like it more if it is written out, it makes easier to read.

```suggestion
DELIVERY_STATE_SHIPPING_RECORDED = "shipping_recorded_in_carrier"
```

delivery_state/models/stock_picking.py
```diff
@@ -28,19 +37,34 @@ class StockPicking(models.Model):
...
         tracking=True,
         readonly=True,
     )
 
+    def _get_delivery_states_completed(self):
+        return [
+            DS_CANCELED,
+            DS_CUS_DELIVERED,
+            DS_NO_UPDATE,
+        ]
```
mt-software-de: This isn't in use?

delivery_state/models/stock_picking.py
```diff
@@ -61,8 +85,8 @@ def _update_delivery_state(self):
                 ("state", "=", "done"),
                 (
                     "delivery_state",
-                    "not in",
-                    ["customer_delivered", "canceled_shipment", "no_update"],
+                    "in",
+                    self._get_delivery_states_in_progress(),
```
mt-software-de, line 144: So the difference is that you now make explicit which state to include,
which leads to the side effect that no transfer will be found which delivery_state is False?
If so can you reflect this in you commit message?

mt-software-de (changes requested):
Thx. Just minor.

ValentinVinagre:
> @chienandalu and @ValentinVinagre, can you check?

Sorry, 14.0 is a very old version for us.

delivery_state/models/stock_picking.py
raumschmiede-joshuaL: Yes, it's not in use. Only `_get_delivery_states_in_progress` is in use, in the cron domain. I thought it makes more sense to implement it to have the counterpart to `_get_delivery_states_in_progress`.  
Even if you could use `_get_delivery_states_in_progress` with the opposite operator in a domain

delivery_state/models/stock_picking.py
raumschmiede-joshuaL: I added a new sentence to it

delivery_state/models/stock_picking.py
mt-software-de: I wouldn't implement something which isn't in use. You will find this code in the test report as not covered.

raumschmiede-joshuaL: Ok. I removed it

raumschmiede-joshuaL:
@mt-software-de, one last review

**EDIT**
Not yet. I saw that I have to change some values to the new constants

raumschmiede-joshuaL:
@mt-software-de, now a code review is possible

mt-software-de (approved):
LGTM

raumschmiede-joshuaL:
@OCA/logistics-maintainers can you merge?

delivery_state/tests/test_delivery_state.py
```diff
@@ -169,23 +179,63 @@ def test_days_fetch_tracking_state_update(self):
...
+
+        # Set PICK to done
+        pick = self.sale.picking_ids.filtered(lambda p: p.state == "assigned")
+        for ml in pick.move_line_ids:
+            ml.qty_done = ml.product_uom_qty
+        pick.button_validate()
+        pick.carrier_id = self.carrier_test
+
+        # Set PACK to done. Call carrier API with this picking
+        pick = self.sale.picking_ids.filtered(lambda p: p.state == "assigned")
```
jbaudoux: What about renaming to `pack` ?
```suggestion
        pack = self.sale.picking_ids.filtered(lambda p: p.state == "assigned")
```

delivery_state/tests/test_delivery_state.py
```diff
@@ -169,23 +179,63 @@ def test_days_fetch_tracking_state_update(self):
...
+        for ml in pick.move_line_ids:
+            ml.qty_done = ml.product_uom_qty
+        pick.carrier_id = self.carrier_test
+        pick.button_validate()
+        pack_pick = pick
+
+        # SET OUT to done
+        pick = self.sale.picking_ids.filtered(lambda p: p.state == "assigned")
```
jbaudoux: and here `ship`
```suggestion
        ship = self.sale.picking_ids.filtered(lambda p: p.state == "assigned")
```

delivery_state/tests/test_delivery_state.py
```diff
@@ -169,23 +179,63 @@ def test_days_fetch_tracking_state_update(self):
...
+        for ml in pick.move_line_ids:
+            ml.qty_done = ml.product_uom_qty
+        pick.button_validate()
+        pick.carrier_id = self.carrier_test
```
jbaudoux: For simulating a real case, the carrier should be set on the ship before validation

delivery_state/tests/test_delivery_state.py
```diff
@@ -169,23 +179,63 @@ def test_days_fetch_tracking_state_update(self):
...
+            autospec=True,
+        ) as mocked:
+            self.env.ref(
+                "delivery_state.ir_cron_delivery_state"
+            ).method_direct_trigger()
+
+            # As only the PACK picking was sent to the carrier API only for this
+            # the tracking state must be updated even if the other pickings have
+            # a carrier set
+            mocked.assert_called_once_with(pack_pick)
```
jbaudoux:
```suggestion
            mocked.assert_called_once_with(pack)
```

delivery_state/tests/test_delivery_state.py
raumschmiede-joshuaL: I wanted to simulate it like it can happen in `delivery_send_to_shipper_at_operation`. Carrier has no effect on the OUT, but OUT has a carrier. Cron must only find the PACK

raumschmiede-joshuaL:
@jbaudoux, can you do another review? I accidentally changed the operator in the cron domain during rebase. Now it should be fine

delivery_state/tests/test_delivery_state.py
jbaudoux: Ah yes, v14... :)
Maybe you could add a migration note for >=18.0 that the carrier should be set on the SO before confirmation and that pack rule should propagate the carrier, instead of setting the carrier manually on each picking.

raumschmiede-joshuaL: I tested it to be sure. I found another bug in the test setup. `pack` was not set at all because the product had no stock.  Now it's better.

I added a note: 
image

raumschmiede-joshuaL:
@mt-software-de, please review again

mt-software-de (approved):
LGTM

raumschmiede-joshuaL:
@OCA/logistics-maintainers can you check again?

jbaudoux (approved):

PR #1149, opened 2026-03-30
[14.0][IMP][BP] delivery_state: POD Fields, no Copy of some Fields and Tracking Update Savepoint

raumschmiede-joshuaL:
Various backports of commits from version 18.0

Relates to PR https://github.com/OCA/delivery-carrier/pull/1148. Because of the wrong tracking number on the OUT the Schenker API raises an error which results in that the tracking state can not be fetched for any other picking. With the savepoint this is fixed.

The other commits relate to the savepoint commit, that's why I backported them, too.

The commit that adds `pod_error` also adds the message post for the admin channel. The channel does not exist in version 14.0. I changed to code so that it works with 14.0. Should I keep this part? As long as no channel exists, this part will not do anything. Or should I add one as XML record in `delivery_state`?

- `tracking_json` was added as `fields.Json`. This does not exist in 14.0, I changed it to `fields.Char`
- `not pod_error` does not work for `invisible` in the view. I changed it to `attrs`
- In the last commit that adds the savepoint I had to change the check whether a cron is calling this. `cron_id` is not set in 14.0, only `lastcall`

### DIFF
--- a/delivery_state/models/stock_picking.py
+++ b/delivery_state/models/stock_picking.py
@@ -16,6 +16,8 @@ class StockPicking(models.Model):
         string="Delivery Date",
         readonly=True,
     )
+    # Technical field to store raw tracking data from the carrier API
+    tracking_json = fields.Json(readonly=True, copy=False)
     tracking_state = fields.Char(
         string="Tracking state",
         readonly=True,
--- a/delivery_state/models/stock_picking.py
+++ b/delivery_state/models/stock_picking.py
@@ -40,6 +40,16 @@ class StockPicking(models.Model):
         tracking=True,
         readonly=True,
     )
+    pod_file = fields.Binary(
+        string="Proof of Delivery File",
+        readonly=True,
+        copy=False,
+    )
+    pod_filename = fields.Char(
+        string="Proof of Delivery Filename",
+        readonly=True,
+        copy=False,
+    )
 
     def tracking_state_update(self):
         """Call to the service provider API which should have the method
--- a/delivery_state/models/stock_picking.py
+++ b/delivery_state/models/stock_picking.py
@@ -17,7 +17,7 @@ class StockPicking(models.Model):
         readonly=True,
     )
     # Technical field to store raw tracking data from the carrier API
-    tracking_json = fields.Json(readonly=True, copy=False)
+    tracking_json = fields.Char(readonly=True, copy=False)
     tracking_state = fields.Char(
         string="Tracking state",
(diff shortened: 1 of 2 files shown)


raumschmiede-joshuaL:
@carlos-lopez-tecnativa and @pedrobaeza, you did the most conversation on the PRs for 18.0. Can you check?

delivery_state/models/stock_picking.py
```diff
@@ -69,3 +104,31 @@ def _update_delivery_state(self):
             ]
         )
         pickings.tracking_state_update()
+
+    def _send_message_pod_error(self):
+        channel_admin = self.env.ref("mail.channel_admin", raise_if_not_found=False)
```
carlos-lopez-tecnativa: You need to adapt some part of the code. This channel does not exist in version 14 yet; it is available from v18+.

carlos-lopez-tecnativa (changes requested):

delivery_state/models/stock_picking.py
raumschmiede-joshuaL: Yes, that's a question I wrote in my first post. If this part should be kept, I can create a mail channel in `delivery_state` . Not sure about it

raumschmiede-joshuaL:
@mt-software-de

delivery_state/models/stock_picking.py
mt-software-de: Normally you should backport with a cherry-pick the changes and then create an own commit for migrating it. 
IMO you should
- backport as it is and then create an own channel
or
- backport it and (remove the logic or just post it on the record)

raumschmiede-joshuaL: Ok. I will adjust the commits.
As there is no default channel to notify admins about technical issues I think it is better to post the message on the record. A channel only for this seems a bit too much

raumschmiede-joshuaL: @mt-software-de and @carlos-lopez-tecnativa, can you check again?

raumschmiede-joshuaL:
@pedrobaeza, can you review, too?

pedrobaeza:
Sorry, 14.0 is a very old version for us.

mt-software-de (approved):
LGTM.

mt-software-de:
@OCA/logistics-maintainers since this is just a BP, could some merge it?

hparfr (approved):
/ocabot merge major

PR #1150, opened 2026-03-30
[FIX] delivery_dropoff_site : name_search not working in sale for customer

saxomanu:
In sale, search by customer's name don't work without this


BhaveshHeliconia (approved):
LGTM!

Prana111025 (changes requested):
Change

PR #1152, opened 2026-03-31
[14.0][IMP] delivery_state: Specify Number of Days without new Tracking

raumschmiede-joshuaL:
For us it happened that the Schenker API did not return a new tracking state (because of https://github.com/OCA/delivery-carrier/pull/1140). Even with the other PR merged old pickings will never get a tracking state as the Schenker API stops to return data after a certain amount of time.
To avoid calling the API every day without fetching new data, you can set on the carrier how many days it will fetch the tracking status updates for a picking. If no new status has been fetched after these days, `delivery_state` is set to `no_update` to prevent further API calls.

The changes are not breaking. As long as the days field is not set, nothing will happen with the pickings.
Carrier modules that depend on `delivery_state` does not have to be updated for this

### DIFF
--- a/delivery_state/models/delivery_carrier.py
+++ b/delivery_state/models/delivery_carrier.py
@@ -1,5 +1,6 @@
 # Copyright 2020 Trey, Kilobytes de Soluciones
 # Copyright 2020 FactorLibre
+# Copyright 2026 Raumschmiede GmbH
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import fields, models
 
@@ -8,6 +9,10 @@ class DeliveryCarrier(models.Model):
     _inherit = "delivery.carrier"
 
     track_carrier_state = fields.Boolean(default=True)
+    days_fetch_tracking_state_update = fields.Integer(
+        help="The tracking state is fetched for a picking for this number of days.\n"
+        "If no final state has been set, the delivery state will be set to 'no_update'"
+    )
 
     def send_shipping(self, pickings):
         res = super().send_shipping(pickings)
--- a/delivery_state/models/stock_picking.py
+++ b/delivery_state/models/stock_picking.py
@@ -1,7 +1,10 @@
 # Copyright 2020 Trey, Kilobytes de Soluciones
 # Copyright 2020 FactorLibre
 # Copyright 2020 Tecnativa - David Vidal
+# Copyright 2026 Raumschmiede GmbH
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from datetime import datetime, timedelta
+
 from markupsafe import Markup
 
 from odoo import _, api, fields, models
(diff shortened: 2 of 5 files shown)


raumschmiede-joshuaL:
@mt-software-de

delivery_state/models/delivery_carrier.py
```diff
@@ -8,6 +9,11 @@ class DeliveryCarrier(models.Model):
     _inherit = "delivery.carrier"
 
     track_carrier_state = fields.Boolean(default=True)
+    days_with_no_new_tracking_state = fields.Integer(
+        help="If set and higher than 0: If a picking has not received a new tracking "
+        "state from the carrier within this number of days, "
+        "its delivery state will be set to 'no_update'"
+    )
```
mt-software-de, line 16:
```suggestion
    days_ignore_tracking_state_update = fields.Integer(
        help="If set and higher than 0: If a picking has not received a new tracking "
        "state from the carrier within this number of days, "
        "its delivery state will be set to 'no_update'"
    )
```

delivery_state/models/stock_picking.py
```diff
@@ -26,6 +29,10 @@ class StockPicking(models.Model):
         string="Tracking state history",
         readonly=True,
     )
+    date_latest_tracking_state = fields.Datetime(
+        readonly=True,
+        copy=False,
+    )
```
mt-software-de, line 35: I don't think a new field is needed.

delivery_state/models/stock_picking.py
```diff
@@ -48,8 +55,32 @@ def tracking_state_update(self):
...
         for picking in self.filtered("carrier_id"):
             method = "%s_tracking_state_update" % picking.delivery_type
-            if hasattr(picking.carrier_id, method):
-                getattr(picking.carrier_id, method)(picking)
+            carrier = picking.carrier_id
+
+            if not hasattr(carrier, method):
+                continue
+
+            getattr(carrier, method)(picking)
```
mt-software-de, line 63: IMO you can also check for the days if the method doesn't exists.
```suggestion
            carrier = picking.carrier_id
            if hasattr(carrier, method):
                getattr(carrier, method)(picking)
```
We should rather set the state to `no_update` if the the method does not exists. But this should be handled in an own PR.

delivery_state/models/stock_picking.py
```diff
@@ -48,8 +55,32 @@ def tracking_state_update(self):
...
+                not last_date
+                and picking.delivery_state == "shipping_recorded_in_carrier"
+            ):
+                last_date = picking.date_shipped
+            if (
+                last_date
+                and last_date <= datetime.now() - timedelta(days=days)
+                and not picking.date_delivered
+            ):
+                picking.delivery_state = "no_update"
```
mt-software-de, line 83: IMO a new date field is not needed just use `date_shipped or date_done`, this simplifies the code a lot. 
Also i would check in the top of your changes if the state has changed. 

```suggestion
            if state_has_changed:
                continue
            days = carrier.days_with_no_new_tracking_state
            if days <= 0:
                continue
            date_shipped = picking.date_shipped or picking.date_done
            if last_date <= datetime.now() - timedelta(days=days):
                picking.delivery_state = "no_update"
```

mt-software-de (changes requested):

raumschmiede-joshuaL:
One idea would be to set `date_latest_tracking_state` in `delivery_state` instead of expecting that the carrier implementation sets it. It is set when the saved delivery state is different than the delivery state after the carrier implementation was called.
But I removed it to have less code.

With this the behaviour has changed. Now it is like this: With 5 days set, the tracking is fetched for a picking for 5 days. After that it will stop fetching states for the picking. If no final state has been set in the last call, `delivery_state` is set to  `no_update`.
Previously it was like this: With 5 days set, the tracking is fetched for a picking for 5 days. If a new state has been set, it waited for another 5 days again. If 5 days has passed between 2 different states, it will stop fetching states for the picking by setting it to `no_update`

delivery_state/models/stock_picking.py
raumschmiede-joshuaL: Created PR https://github.com/OCA/delivery-carrier/pull/1194 for this

raumschmiede-joshuaL:
@mt-software-de and @hparfr, please review here, too

hparfr (approved):
Lgtm

mt-software-de (approved):
LGTM

mt-software-de:
@OCA/logistics-maintainers could you merge this one?

jbaudoux (approved):

PR #1153, opened 2026-04-01
[19.0][MIG] delivery_sending: Migration to 19.0

pablo-cort-s73:
This PR migrates the `delivery_sending` module from 13.0 to 19.0.

#### Major changes:

*   **Manifest & Dependencies**:
    *   Updated version to `19.0.1.0.0`.
    *   Updated external dependency from `suds` to `suds-py3` for Python 3.12+ compatibility.
    *   Added missing dependencies: `delivery` and `stock_delivery`.
*   **Models**:
    *   Migrated carrier packaging extension from `product.packaging` to `stock.package.type` (Standard refactoring for Odoo 17+).
    *   Updated `ondelete` policies for required selection fields.
    *   Refactored ORM calls and translations to use `self.env._()`.
    *   Fixed linting issues (Ruff E501, UP031, Pylint-Odoo W8113, W8161, E8106, etc.).
*   **Views**:
    *   Removed `attrs` and replaced them with Odoo 17+ `invisible`/`required` attributes.
    *   Updated inherited view IDs (moved from `delivery` to `stock_delivery`).
    *   Moved `groups` from the view record to the specific elements (required for inherited views).
*   **Tests**:
    *   Migrated base class to `TransactionCase`.
    *   Implemented `unittest.mock.patch` for the `suds` Client to avoid external network calls during tests (mandatory for v17+).
    *   Updated field names (`move_ids`, `quantity`) to match the new Stock ORM.
    *  Added more tests to cover missed lines


Reyes4711-S73 (approved):
LGTM

sbiosca-s73 (approved):

Reyes4711-S73:
@OCA/logistics-maintainers Please, can you review/merge this PR

hparfr:
hi, 
please reword the commit `[IMP] $module: pre-commit auto fixes]`

delivery_sending/__manifest__.py
```diff
@@ -3,14 +3,19 @@
...
     "website": "https://github.com/OCA/delivery-carrier",
     "author": "Impulso Diagonal, Tecnativa, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "depends": ["delivery_package_number", "delivery_state"],
-    "external_dependencies": {"python": ["suds"]},
+    "depends": [
+        "delivery",
+        "stock_delivery",
```
BhaveshHeliconia: Why were these two additional dependencies added?

BhaveshHeliconia (changes requested):

delivery_sending/__manifest__.py
pablo-cort-s73: `stock_delivery` is required directly by this module: it provides the carrier-related fields used on `stock.picking`, the `package_carrier_type` field extended by the module, and the picking form view that we inherit.

`delivery_package_number` and `delivery_state` are also direct functional dependencies: the former provides `number_of_packages`, used in the Sending API request, while the latter updates the shipment state after sending or cancelling it.

However, `delivery` is already a dependency of `stock_delivery`, so its explicit declaration here is redundant. I’ll remove it.

BhaveshHeliconia: Thanks for the explanation. One thing I noticed is that `delivery_package_number` already depends on `stock_delivery`, as shown in its manifest:
[[delivery_package_number manifest](https://github.com/OCA/delivery-carrier/blob/e34dad3811070f23a34ab27ac43e323f457feb07/delivery_package_number/__manifest__.py?utm_source=chatgpt.com#L13)](https://github.com/OCA/delivery-carrier/blob/e34dad3811070f23a34ab27ac43e323f457feb07/delivery_package_number/__manifest__.py?utm_source=chatgpt.com#L13)

Since this module already depends on `delivery_package_number`, `stock_delivery` should already be installed as a transitive dependency. Is there any specific reason why we need to declare `stock_delivery` as a direct dependency here as well?

PR #1156, opened 2026-04-06
[18.0] [IMP] delivery_ups_oca: Save the POD, Signature and tracking_state

carlos-lopez-tecnativa:
By passing the parameter `returnPOD=true`, the API can return the POD in HTML format.
By passing the parameter returnSignature=true, the API can return the Signature in PNG format.

TT59692

### DIFF
--- a/delivery_ups_oca/models/delivery_carrier.py
+++ b/delivery_ups_oca/models/delivery_carrier.py
@@ -236,6 +236,7 @@ class DeliveryCarrier(models.Model):
             picking_vals = {
                 "delivery_state": response["delivery_state"],
                 "tracking_state_history": response["tracking_state_history"],
+                "tracking_state": response.get("tracking_state"),
             }
             if response.get("pod"):
                 picking_vals.update(
--- a/delivery_ups_oca/models/delivery_carrier.py
+++ b/delivery_ups_oca/models/delivery_carrier.py
@@ -233,8 +233,19 @@ class DeliveryCarrier(models.Model):
         ):
             ups_request = UpsRequest(self)
             response = ups_request.tracking_state_update(picking)
-            picking.delivery_state = response["delivery_state"]
-            picking.tracking_state_history = response["tracking_state_history"]
+            picking_vals = {
+                "delivery_state": response["delivery_state"],
+                "tracking_state_history": response["tracking_state_history"],
+            }
+            if response.get("pod"):
+                picking_vals.update(
+                    {
+                        "pod_filename": f"ups_pod_{picking.carrier_tracking_ref}.html",
+                        "pod_file": response["pod"],
+                        "pod_error": False,
+                    }
+                )
+            picking.write(picking_vals)
 
     def ups_update_token(self):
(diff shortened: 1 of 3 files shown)


delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -204,6 +204,15 @@ def ups_tracking_state_update(self, picking):
             response = ups_request.tracking_state_update(picking)
             picking.delivery_state = response["delivery_state"]
             picking.tracking_state_history = response["tracking_state_history"]
+            picking.tracking_state = response.get("tracking_state")
+            if response.get("pod"):
```
pedrobaeza, line 210: There are a lot of write over the record. Why don't group all of them?

carlos-lopez-tecnativa: Done

jans23:
In order to test this, does the shipping method require any specific setting or is a POD returned for every shipment?

carlos-lopez-tecnativa:
> In order to test this, does the shipping method require any specific setting or is a POD returned for every shipment?

To test this, you must enable the `ups_tracking_state_update_sync=True` field and have a picking that has been delivered to the customer.

You can either:

1. Run the cron job at https://github.com/OCA/delivery-carrier/blob/eecc3d262fa9b0547cd11cbcf0b40f739f0726c1/delivery_state/data/ir_cron_data.xml#L4
2. Or click the Update tracking state button in the picking form view: https://github.com/OCA/delivery-carrier/blob/eecc3d262fa9b0547cd11cbcf0b40f739f0726c1/delivery_state/views/stock_picking_views.xml#L30

I’m not sure whether this can be tested in a test environment; we tested it in a production environment.

KDRdev (changes requested):
Please squash your commits, otherwise LGTM :)

carlos-lopez-tecnativa:
> Please squash your commits, otherwise LGTM :)

Each commit adds a specific feature; for this reason, I kept the commits split. I’m not sure whether it would be a better approach to squash them into a single commit. What do you think, @pedrobaeza?

pedrobaeza (approved):
There's no strict criteria for this. Seeing the mix of diff lines, I think in this case we can keep it separated for being clearer in each of the changes.

jans23:
We tested this feature successfully in our production system. It works great!

One possible improvement would be to rewrite the local paths of `./ups_logo.gif` and `./ups_banner.gif` to global paths which would enable displaying these images because the HTML file is downloaded in any case.

carlos-lopez-tecnativa:
> One possible improvement would be to rewrite the local paths of `./ups_logo.gif` and `./ups_banner.gif` to global paths which would enable displaying these images because the HTML file is downloaded in any case.

I don’t understand this, but if you can improve it, feel free to create a follow-up PR to implement your suggestion here.

KDRdev (approved):

carlos-lopez-tecnativa:
ping @pedrobaeza

PR #1158, opened 2026-04-07
[16.0][MIG] delivery_carrier_default_tracking_url: Migration to 16.0

DorianMAG:


delivery_carrier_default_tracking_url/models/sale_order.py
```diff
@@ -0,0 +1,13 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class SaleOrder(models.Model):
+
+    _inherit = "sale.order"
+
+    default_tracking_url = fields.Char(
```
flotho, line 11: Could you elaborate the use of this field?
Either remove it or use an xml to display the info

flotho (changes requested):

PR #1162, opened 2026-04-12
delivery_carrier_label_default: icon was not a valid svg, but instead an html downloaded from github

sebastienbeau:
Fix broken icon. If you take a look to the svg, it's not an svg it's an html page of github 

@hparfr @florian-dacosta


pedrobaeza (approved):
Fast-tracking for fixing the problem:

/ocabot merge patch

PR #1163, opened 2026-04-13
delivery_carrier_label_default: icon was not a valid svg, but instead an html downloaded from github

sebastienbeau:
@pedrobaeza same as here : https://github.com/OCA/delivery-carrier/pull/1162


pedrobaeza (approved):
Uf, too old branch..

sebastienbeau:
yes really old, but it's just to avoid blocking error when building oca stat

pedrobaeza:
Maybe it's better to change the module to prevent this problem, but meanwhile, let's merge it manually, as there's no current support in CI for these versions.

PR #1167, opened 2026-04-13
[17.0][MIG] delivery_roulier_dpd_fr: Migration to 17.0

dreispt:


florian-dacosta (approved):
LGTM code review only

@dreispt Note that in roulier (the library), there was a big change with a new api, so a new implementation of dpd has been done : https://github.com/akretion/roulier/tree/master/roulier/carriers/dpd_fr
Both are supported for now, but on the long run, the first one (dpd_fr_soap) will be dropped.

There is an none OCA implementation on v14 here : https://github.com/akretion/delivery-carrier/commit/42a6eda9fc4e84ea210a1b7950f2b948a1e1084a
It  does not really change anything as all the logic is in the library.

Not blocking if you want to follow with old roulier implementation in 17 but I think we will change in v18 then.

dreispt:
Superseded by #1175

dreispt:
@florian-dacosta Thank you for the review and pointers.
In our initial testing if this we were unable to create a shipping label.
So I decided to try the code you shared.
It is migrated to 17.0 with #1175 

It should be able to work properly, correct?
I just noticed a new version of `roulier` lib was published in PyPi, perhaps that was why it wasn't working before?

Thank you

PR #1170, opened 2026-04-15
[18.0][ADD] stock_picking_delivery_package_type_carrier

henrybackman:
Add dedicated carrier field to stock picking package type to allow
filtering packages based on individual carriers instead of just delivery
carrier types

Had to modify the `stock_picking_delivery_package_type_domain` tests due to the FakeModelLoader issue


stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
```diff
@@ -0,0 +1,44 @@
...
+                        "Package type '%(package_type)s' is not valid for carrier"
+                        " '%(carrier)s'."
+                    )
+                    % {
+                        "package_type": package_type.name,
+                        "carrier": package_carriers[0].name,
+                    }
+                )
+
+    def write(self, vals):
```
TDu, line 39: The check should also be done at creation.

stock_picking_delivery_package_type_carrier/models/stock_package_type.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2026 Camptocamp SA (https://www.camptocamp.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class PackageType(models.Model):
+    _inherit = "stock.package.type"
+
+    package_carrier_id = fields.Many2one(
```
jbaudoux, line 10: There should already be a carrier code. Can you filter the available carriers based on this?

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
```diff
@@ -0,0 +1,44 @@
+from odoo import _, models
+from odoo.exceptions import ValidationError
+
+
+class StockQuantPackage(models.Model):
+    _inherit = "stock.quant.package"
+
+    def _check_package_type_carrier_compatibility(self, package_type):
```
jbaudoux, line 8: Can you set as constraint instead of hooking write?

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
```diff
@@ -0,0 +1,44 @@
...
+                        "package_type": package_type.name,
+                        "carrier": package_carriers[0].name,
+                    }
+                )
+
+    def write(self, vals):
+        package_type_id = vals.get("package_type_id")
+        if package_type_id:
+            package_type = self.env["stock.package.type"].browse(package_type_id)
+            self._check_package_type_carrier_compatibility(package_type)
```
jbaudoux, line 43: You need to do the check after the call to super that could alter what will be done. That's why you need to replace by a constraint

jbaudoux (changes requested):
I would only keep tbe restriction in the choose delivery package (put in pack). The constraint don't bring value and could be more problematic than usefi

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
```diff
@@ -0,0 +1,44 @@
...
+
+        for package in self:
+            pkg_id = package.id
+            domain = [
+                "|",
+                ("result_package_id", "=", pkg_id),
+                ("package_id", "=", pkg_id),
+            ]
+            package_carriers = (
+                self.env["stock.move.line"]
```
TDu, line 21: Can't we use `package.move_line_ids` to find the related move lines ?

stock_picking_delivery_package_type_carrier/views/stock_package_type.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="stock_package_type_form" model="ir.ui.view">
+        <field name="name">stock.package.carrier.id.form</field>
+        <field name="model">stock.package.type</field>
+        <field name="inherit_id" ref="stock.stock_package_type_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='delivery']" position="inside">
+                <group>
+                    <field name="package_carrier_id" />
```
TDu, line 10: Probably could have a domain to filter the possible carrier based on `package_carrier_type`

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
henrybackman: The package doesn't have move_line_ids and not sure how to get to those from the package:
```
(Pdb++) pp package.read()
[{'company_id': False,
  'create_date': datetime.datetime(2026, 4, 29, 12, 24, 40, 810341),
  'create_uid': (1, 'OdooBot'),
  'display_name': 'PACK0000005',
  'id': 5,
  'location_id': False,
  'name': 'PACK0000005',
  'owner_id': False,
  'pack_date': datetime.date(2026, 4, 29),
  'package_type_id': False,
  'package_use': 'disposable',
  'quant_ids': [],
  'shipping_weight': 0.0,
  'valid_sscc': False,
  'weight': 0.0,
  'weight_is_kg': True,
  'weight_uom_name': 'kg',
  'weight_uom_rounding': 0.01,
  'write_date': datetime.datetime(2026, 4, 29, 12, 24, 40, 810341),
  'write_uid': (1, 'OdooBot')}]
```

stock_picking_delivery_package_type_carrier/models/stock_package_type.py
henrybackman: `PackageType` only has `package_carrier_type` but this is not granular enough. For example, "PostLogistics" carrier type includes specific carriers that might not be compatible with the package type. Like Thermo packages.

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
henrybackman: Changed as constraint

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
henrybackman: Changed as constraint so it covers both write and create

stock_picking_delivery_package_type_carrier/views/stock_package_type.xml
henrybackman: Good point, added the domain

stock_picking_delivery_package_type_carrier/models/stock_package_type.py
jbaudoux: @henrybackman What I mean is to add a domain on this field like
`[("delivery_type", "=", package_carrier_type)] if package_carrier_type not in ("none", False) else []`
so that if you have a `package_carrier_type` on the package type, then you can only link it to carrier with the corresponding `delivery_type`

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
```diff
@@ -0,0 +1,43 @@
...
+        dedicated_carrier = package_type.package_carrier_id
+        if not dedicated_carrier:
+            return
+
+        for package in self:
+            pkg_id = package.id
+            domain = [
+                "|",
+                ("result_package_id", "=", pkg_id),
+                ("package_id", "=", pkg_id),
```
jbaudoux: The source package shouldn't be validated with the picking carrier. Only the result package.

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
jbaudoux: Maybe he means you could have implemented the reverse o2m relation field `result_move_line_ids` instead of calling search

stock_picking_delivery_package_type_carrier/models/stock_package_type.py
jbaudoux: Ah you did it already in the view :)

stock_picking_delivery_package_type_carrier/views/stock_package_type.xml
```diff
@@ -0,0 +1,18 @@
...
+    <record id="stock_package_type_form" model="ir.ui.view">
+        <field name="name">stock.package.carrier.id.form</field>
+        <field name="model">stock.package.type</field>
+        <field name="inherit_id" ref="stock.stock_package_type_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='delivery']" position="inside">
+                <group>
+                    <field
+                        name="package_carrier_id"
+                        domain="[('delivery_type', '=', package_carrier_type)]"
```
jbaudoux: When there is no carrier integration, the value is "none" and the delivery_type can be fixed or based on rules (but there are other module adding other options, so I would leave it open in this case)
```suggestion
                        domain="[('delivery_type', '=', package_carrier_type)] if package_carrier_type != 'none' else []"
```

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
henrybackman: Ah yes, now I got it. I added the result_move_line_ids and used that for getting the carrier ids :+1:

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
henrybackman: Fixed by switching from search to the `result_move_line_ids` field

stock_picking_delivery_package_type_carrier/views/stock_package_type.xml
henrybackman: Good point, changed this

henrybackman:
Had to modify the stock_picking_delivery_package_type_domain tests due to the FakeModelLoader issue

stock_picking_delivery_package_type_domain/tests/common.py
```diff
@@ -6,78 +6,75 @@
...
-        cls.delivery = cls.delivery_obj.create(
+        self.delivery = self.delivery_obj.create(
             {
                 "name": "Test",
                 "delivery_type": "test",
-                "product_id": cls.product_delivery.id,
+                "product_id": self.product_delivery.id,
             }
         )
-        cls.partner = cls.env["res.partner"].create(
```
jbaudoux, line 40: Starting from here, it can remain in `setUpClass` as it doesn't depend on DeliveryCarrier, StockPackageType

jbaudoux (commented):
Make a separate commit for changes on module `stock_picking_delivery_package_type_domain` and put it before the commit for the new module that depends on it

stock_picking_delivery_package_type_carrier/wizards/choose_delivery_package.py
```diff
@@ -0,0 +1,17 @@
...
+
+class ChooseDeliveryPackage(models.TransientModel):
+    _inherit = "choose.delivery.package"
+
+    def _compute_package_type_domain(self):
+        package_carrier = self.picking_id.carrier_id
+        if not package_carrier:
+            return super()._compute_package_type_domain()
+        # if package has specific carrier defined, use that for the domain
+        domain = [("package_carrier_id", "=", package_carrier.id)]
```
jbaudoux: You need to AND with the domain returned by super()
Also allow package type without restriction
```suggestion
        domain = [("package_carrier_id", "in", (package_carrier.id, False))]
```

stock_picking_delivery_package_type_carrier/models/stock_package_type.py
```diff
@@ -0,0 +1,13 @@
...
+        "delivery.carrier",
+        string="Dedicated carrier",
```
jbaudoux: You need to be able to list multiple values
```suggestion
    package_carrier_ids = fields.Many2many(
        "delivery.carrier",
        string="Delivery Methods",
        help="Restrict this package type to listed delivery methods. When empty, it will be available to all delivery methods of the carrier",
```

stock_picking_delivery_package_type_carrier/views/stock_package_type.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="stock_package_type_form" model="ir.ui.view">
+        <field name="name">stock.package.carrier.id.form</field>
+        <field name="model">stock.package.type</field>
+        <field name="inherit_id" ref="stock.stock_package_type_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='delivery']" position="inside">
```
jbaudoux: Place it after `package_carrier_type` you rely on.
```suggestion
            <xpath expr="//field[@name='package_carrier_type']" position="after">
```

stock_picking_delivery_package_type_carrier/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Stock Picking Delivery Package Type Carrier",
+    "summary": """
+        Restricts package type selection and assignment
+        to package types dedicated to the picking carrier.""",
+    "version": "18.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "Camptocamp SA, Odoo Community Association (OCA)",
```
jbaudoux: You can put at least me as maintainer?

```suggestion
    "author": "Camptocamp SA, BCIM, Odoo Community Association (OCA)",
    "maintainers": ["jbaudoux"],
```

stock_picking_delivery_package_type_domain/tests/common.py
henrybackman: Thanks, moved those back to the setUp()

stock_picking_delivery_package_type_carrier/wizards/choose_delivery_package.py
henrybackman: Modified the domain to work with the Many2many field and include the package types without carrier id

stock_picking_delivery_package_type_carrier/models/stock_package_type.py
henrybackman: Changed to Many2many

stock_picking_delivery_package_type_carrier/views/stock_package_type.xml
henrybackman: Moved:
image

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def _check_package_type_carrier_compatibility(self, package_type):
+        dedicated_carriers = package_type.package_carrier_ids
+        if not dedicated_carriers:
+            return
+
+        for package in self:
+            package_carriers = package.result_move_line_ids.mapped(
+                "picking_id.carrier_id"
+            )
```
jbaudoux: No need for mapped and do not check move lines that are not being processed
```suggestion
            package_carriers = package.result_move_line_ids.filtered(lambd ml: ml.state in ("partially_available", "assigned")).picking_id.carrier_id
            )
```

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
```diff
@@ -0,0 +1,41 @@
...
+            if not package_carriers:
+                continue
+            if not any(carrier in dedicated_carriers for carrier in package_carriers):
```
jbaudoux: if the package is used in multiple pickings with different carrier, its package type should be valid for all pickings (i.e. carrier)

```suggestion
            if not all(carrier in dedicated_carriers for carrier in package_carriers):
```

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
```diff
@@ -0,0 +1,41 @@
...
+                        "package_type": package_type.name,
+                        "carrier": package_carriers[0].name,
+                    }
+                )
+
+    @api.constrains("package_type_id")
+    def _check_package_type_id_carrier_compatibility(self):
+        for package in self:
+            if package_type := package.package_type_id:
+                package._check_package_type_carrier_compatibility(package_type)
```
jbaudoux: Just wondering why having this method and not put it all in this constrain ?

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
henrybackman: Changed

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
henrybackman: Good point. I made a test for this case and changed it to `all`

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
henrybackman: I think originally it was due to trying to limit the indentations. I moved it in the constrain and removed some unnecessary indentations there

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
```diff
@@ -0,0 +1,37 @@
...
+            if not dedicated_carriers:
+                continue
+            package_carriers = package.result_move_line_ids.filtered(
+                lambda ml: ml.state in ("partially_available", "assigned")
+            ).picking_id.carrier_id
+            if not package_carriers:
+                continue
+            if not all(carrier in dedicated_carriers for carrier in package_carriers):
+                carrier_names = ", ".join(carrier.name for carrier in package_carriers)
+                msg = _(
```
jbaudoux:
```suggestion
                msg = self.env._(
```

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
```diff
@@ -0,0 +1,37 @@
...
+                    "Package type '%(package_type)s' is not valid for carriers"
+                    " '%(carrier)s'."
+                ) % {
+                    "package_type": package_type.name,
+                    "carrier": carrier_names,
+                }
```
jbaudoux: should be part of the call to `_`
```suggestion
                , package_type=package_type.name, carrier=carrier_names)
```

jbaudoux:
@henrybackman Squash the fixup and it should be good :)

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
henrybackman: This I somehow always miss :sweat_smile: Thanks for catching!

stock_picking_delivery_package_type_carrier/models/stock_package_type.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2026 Camptocamp SA (https://www.camptocamp.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class PackageType(models.Model):
+    _inherit = "stock.package.type"
+
+    package_carrier_ids = fields.Many2many(
```
divad1196: Name is not clear on the intent and can easily conflict. Would prefix it `allowed_`

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
```diff
@@ -0,0 +1,36 @@
...
+                continue
+            dedicated_carriers = package_type.package_carrier_ids
+            if not dedicated_carriers:
+                continue
+            package_carriers = package.result_move_line_ids.filtered(
+                lambda ml: ml.state in ("partially_available", "assigned")
+            ).picking_id.carrier_id
+            if not package_carriers:
+                continue
+            if not all(carrier in dedicated_carriers for carrier in package_carriers):
```
divad1196: You can use set(-like) operations:
```python
carriers_not_allowed = package_carriers - dedicated_carriers
# Short circuit to reduce nesting
if not carriers_not_allowed:
    continue
...
raise ...
```

NOTE: you are already assured that `dedicated_carriers` is not empty from previous test

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
```diff
@@ -0,0 +1,36 @@
...
+                carrier_names = ", ".join(carrier.name for carrier in package_carriers)
```
divad1196: Use `mapped`
`", ".join(carriers_not_allowed.mapped('name'))`

NOTE: the current code would list all carriers as not valid even if only 1 is not valid. It should re-use the filtered result from before, then one I called `carriers_not_allowed`

stock_picking_delivery_package_type_carrier/tests/test_choose_delivery_package_type.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class TestChooseDeliveryPackageType(CommonChooseDeliveryPackage, BaseCommon):
+    def test_domain_gets_correct_package_type(self):
+        sale = self._create_sale()
+        sale.action_confirm()
+        self.assertTrue(sale.picking_ids)
+        carrier = sale.picking_ids.carrier_id
+        res = sale.picking_ids.action_put_in_pack()
+        model = res.get("res_model")
```
divad1196: Took a few seconds to understand the intent.
AFAIK, there is sady no standard function like `record_from_action_dict` in odoo`, but an utility function here would make it more readable

stock_picking_delivery_package_type_carrier/tests/test_choose_delivery_package_type.py
```diff
@@ -0,0 +1,96 @@
...
+        context = res.get("context")
```
divad1196: `context = res.get("context") or {}`
Should not happen, but if `context` is not provided (e.g. poped from dict), then the unpacking will crash

stock_picking_delivery_package_type_carrier/tests/test_choose_delivery_package_type.py
```diff
@@ -0,0 +1,96 @@
...
+        self.assertEqual("choose.delivery.package", model)
+        self.wizard = self.env[model].with_context(**context).create({})
+        self.assertTrue(self.wizard)
+        self.assertEqual(
+            self.wizard.package_type_domain,
+            [
+                "|",
+                ("package_carrier_ids", "=", False),
+                ("package_carrier_ids", "in", [carrier.id]),
```
divad1196: `[carrier.id]` -> `carrier.ids`

stock_picking_delivery_package_type_carrier/tests/test_choose_delivery_package_type.py
```diff
@@ -0,0 +1,96 @@
...
+
+        picking = sale.picking_ids
+        carrier = picking.carrier_id
+
+        # package type selectable for this picking
+        matching_type = self.env["stock.package.type"].create(
+            {
+                "name": "Type Matching Carrier",
+                "package_carrier_type": carrier.delivery_type,
+                "package_carrier_ids": [Command.set([carrier.id])],
```
divad1196: `carrier.ids`, same things for similar cases.

stock_picking_delivery_package_type_carrier/tests/test_choose_delivery_package_type.py
```diff
@@ -0,0 +1,96 @@
...
+                "name": "Type Other Carrier",
+                "package_carrier_type": carrier.delivery_type,
+                "package_carrier_ids": [Command.set([other_carrier.id])],
+            }
+        )
+
+        res = picking.action_put_in_pack()
+        wizard_model = res.get("res_model")
+        wizard_context = res.get("context")
+        wizard = self.env[wizard_model].with_context(**wizard_context).create({})
```
divad1196: Utility function that I mentioned before would make it readable and also remove duplicate code

stock_picking_delivery_package_type_carrier/tests/test_stock_quant_package.py
```diff
@@ -0,0 +1,109 @@
...
+                "product_id": self.product.id,
+                "product_uom_id": self.product.uom_id.id,
+                "picking_id": picking.id,
+                "quantity": 1,
+                "location_id": self.env.ref("stock.stock_location_stock").id,
+                "location_dest_id": self.env.ref("stock.stock_location_customers").id,
+            }
+        )
+
+        package = self.env["stock.quant.package"].create({})
```
divad1196: Any reason to not create the package before the `move_line`? This way you could just pass `result_package_id` on `create`.

stock_picking_delivery_package_type_carrier/tests/test_stock_quant_package.py
```diff
@@ -0,0 +1,109 @@
...
+        move_line.result_package_id = package
+        return package, picking
+
+    def test_constraint_allows_package_type_without_dedicated_carrier(self):
+        package, picking = self._create_package_for_picking_carrier()
```
divad1196: This function is ~~almost~~ always called. Shouldn't it be in a setupclass method ?

stock_picking_delivery_package_type_carrier/wizards/choose_delivery_package.py
```diff
@@ -0,0 +1,25 @@
...
+class ChooseDeliveryPackage(models.TransientModel):
+    _inherit = "choose.delivery.package"
+
+    def _compute_package_type_domain(self):
+        package_carrier = self.picking_id.carrier_id
+        domain = super()._compute_package_type_domain() or []
+        if not package_carrier:
+            return domain
+        # if package has specific carrier defined, use that for the domain
+        package_carrier_domain = OR(
```
divad1196: Why not just 
```python
package_carrier_domain = [
    "|",
    ....
]
```

like you did previously ? Here you control the domain portion completely.
But the `AND` that comes next makes perfect sense

stock_picking_delivery_package_type_carrier/wizards/choose_delivery_package.py
```diff
@@ -0,0 +1,25 @@
...
+            [
+                [("package_carrier_ids", "=", False)],
+                [("package_carrier_ids", "in", [package_carrier.id])],
+            ]
+        )
+        domain = AND([domain, package_carrier_domain])
+        for wizard in self:
```
divad1196: 3 options:
- idiomatic: `self.package_type_domain = domain`
  Odoo loops automatically
- explicit: `self.update(package_type_domain=domain)`
- explicit, more common but immediate write: `self.write(package_type_domain=domain)`

divad1196 (commented):
Nothing critical, mostly tips

I would at least recommend the `carrier_names` change because if you have carriers A, B and C and only C is not allowed, then you will show to the use "A, B and C are not allwoed"

divad1196 (approved):

stock_picking_delivery_package_type_carrier/models/stock_package_type.py
henrybackman: I've changed the field name to `allowed_package_carrier_ids`

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
henrybackman: Good point, changed this to use the set operation

stock_picking_delivery_package_type_carrier/models/stock_quant_package.py
henrybackman: Done

stock_picking_delivery_package_type_carrier/tests/test_choose_delivery_package_type.py
henrybackman: Helper function made a lot of sense, implemented that to use by both of the tests

stock_picking_delivery_package_type_carrier/tests/test_choose_delivery_package_type.py
henrybackman: True, but in test context I will expect to always have a context

stock_picking_delivery_package_type_carrier/tests/test_stock_quant_package.py
henrybackman: No reason, modified this to create the package before the move_line creation

stock_picking_delivery_package_type_carrier/tests/test_stock_quant_package.py
henrybackman: Yes, this made more sense in `setUp()` as it is needed for each test. I modified the test to use that instead of the separate helper method

stock_picking_delivery_package_type_carrier/wizards/choose_delivery_package.py
henrybackman: I used `OR` to have consistent syntax as I was using the `AND` as well. It looked clearer to me this way

stock_picking_delivery_package_type_carrier/wizards/choose_delivery_package.py
henrybackman: Good point, I changed to the `self.package_type_domain = domain`

stock_picking_delivery_package_type_carrier/wizards/choose_delivery_package.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class ChooseDeliveryPackage(models.TransientModel):
+    _inherit = "choose.delivery.package"
+
+    def _compute_package_type_domain(self):
+        package_carrier = self.picking_id.carrier_id
+        domain = super()._compute_package_type_domain() or []
+        if not package_carrier:
+            return domain
```
jbaudoux: Call to super does not return a domain

```suggestion
        res = super()._compute_package_type_domain()
        for wizard in self:
            package_carrier = wizard.picking_id.carrier_id
            if not package_carrier:
                continue
```

stock_picking_delivery_package_type_carrier/wizards/choose_delivery_package.py
```diff
@@ -0,0 +1,24 @@
...
+        # if package has specific carrier defined, use that for the domain
+        package_carrier_domain = OR(
+            [
+                [("allowed_package_carrier_ids", "=", False)],
+                [("allowed_package_carrier_ids", "in", [package_carrier.id])],
+            ]
+        )
+        domain = AND([domain, package_carrier_domain])
+        self.package_type_domain = domain
```
jbaudoux:
```suggestion
            domain = AND([wizard.package_type_domain, package_carrier_domain])
            wizard.package_type_domain = domain
        return res
```

jbaudoux (changes requested):
compute method does not return the computed value

stock_picking_delivery_package_type_carrier/views/stock_package_type.xml
```diff
@@ -0,0 +1,16 @@
...
+    <record id="stock_package_type_form" model="ir.ui.view">
+        <field name="name">stock.package.carrier.id.form</field>
+        <field name="model">stock.package.type</field>
+        <field name="inherit_id" ref="stock.stock_package_type_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='package_carrier_type']" position="after">
+                <field
+                    name="allowed_package_carrier_ids"
+                    domain="[('delivery_type', '=', package_carrier_type)] if package_carrier_type != 'none' else []"
+                />
```
jbaudoux, line 13: Use widget many2many_tags

jbaudoux:
Replaced by https://github.com/OCA/delivery-carrier/pull/1214

PR #1177, opened 2026-04-23
[19.0][MIG] delivery_purchase: Migration to 19.0

Andrii9090-tecnativa:
Migration to 19.0

I changed the dependency from `purchase` to `purchase_stock` because the `purchase_id` field is added by the `purchase_stock` module. Without this change, OCA and Odoo tests fail

image

@victoralmau @pedrobaeza can you review this?

@Tecnativa
TT61933


BhaveshHeliconia (approved):
Code review LGTM!

delivery_purchase/models/delivery_carrier.py
```diff
@@ -35,8 +35,9 @@ def purchase_rate_shipment(self, order):
                 and (order.amount_total + order.delivery_price) >= self.amount
             ):
                 res["warning_message"] = self.env._(
-                    "The shipping is free since the order amount exceeds %.2f."
-                ) % (self.amount)
+                    "The shipping is free since the order amount exceeds {amount:.2f}.",
```
pedrobaeza: There's no need of this change, as this forces to retranslate everything.

Andrii9090-tecnativa: But without this change the pr-commit failed 
`delivery_purchase/models/delivery_carrier.py:37: [W8301(translation-not-lazy), DeliveryCarrier.purchase_rate_shipment] Use lazy % or .format() or % formatting in odoo._ functions`

BhaveshHeliconia: Yes, agreed.

pedrobaeza: But what you have to change is not this one, but to pass `self.amount` as argument, not with `%`

Andrii9090-tecnativa: self.amount is passed as an argument to self.env._() (line 39), similar to how it is used in Odoo core [self.env._](https://github.com/OCA/OCB/blob/19.0/addons/l10n_account_withholding_tax/models/account_withholding_line.py#L366)

pedrobaeza: Yes, right now you are passing via argument, but you have changed the original translation string. Just do the same, but without changing the string.

pedrobaeza (approved):

victoralmau (approved):
Functional review OK

PR #1181, opened 2026-05-11
[18.0] [FIX] delivery_roulier: Fix tests and add common

paradoxxxzero:
`odoo_test_helper` model faking should now be done at `setUp` (https://github.com/OCA/odoo-test-helper/pull/41).

This also split roulier test setup in a common.


sebastienbeau (approved):
LGTM (code review)

PR #1185, opened 2026-05-19
[19.0][MIG] partner_delivery_zone_calendar

glitchov:
Migration from 14.0


gurneyalex (approved):

partner_delivery_zone_calendar/tests/test_partner_delivery_zone_calendar.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.addons.base.tests.common import BaseCommon
+
+
+class TestPartnerDeliveryZoneCalendar(BaseCommon):
```
BhaveshHeliconia, line 7: This module have only orm fields so for that no need any test case

BhaveshHeliconia (changes requested):

partner_delivery_zone_calendar/tests/test_partner_delivery_zone_calendar.py
glitchov: It's never useless to run the code. The test name is obvious.

BhaveshHeliconia: I understand the point about running the code. My concern was more that the test currently only validates basic ORM field definitions without covering any functional scenario or business logic, so the practical value is limited.

rousseldenis (approved):

rousseldenis:
@BhaveshHeliconia

BhaveshHeliconia (approved):

rousseldenis:
@sbidoul

sbidoul:
name reserved on PyPI

PR #1189, opened 2026-05-20
[18.0][IMP] delivery_ups_oca: phone is optional

KDRdev:


jans23:
@carlos-lopez-tecnativa your review would be most appreciated.

delivery_ups_oca/__manifest__.py
```diff
@@ -4,12 +4,12 @@
 {
     "name": "Delivery UPS OCA",
     "summary": "Integrate UPS webservice",
-    "version": "18.0.1.1.0",
+    "version": "18.0.1.1.1",
```
carlos-lopez-tecnativa: You do not need to change the manifest version; this is done automatically when the PR is merged.

delivery_ups_oca/models/ups_request.py
```diff
@@ -150,6 +149,11 @@ def _partner_to_shipping_data(self, partner, **kwargs):
             ),
         )
 
+        phone_number = partner.phone or partner.mobile
```
carlos-lopez-tecnativa:
```suggestion
        phone_number = partner.phone or partner.mobile or ""
```

carlos-lopez-tecnativa (changes requested):
Could you explain a bit what the goal of this PR is?

I imagine that if the partner has neither a `phone` nor a `mobile` number, this sends the string "`False`" in the `Phone` field to the UPS API. Does this cause any error? Could you paste the error returned by the UPS API?

If sending an empty string ("") instead of False solves the problem, could that be a better approach instead of sending False?

jans23:
In Germany UPS does not require recipient's phone number (while in most other countries the phone number is mandatory). This patch allows sending deliveries to German recipients which have no phone number specified.

The error message without this patch is simply: `Sending to UPS: 120209 Missing or invalid ship to phone number `

delivery_ups_oca/models/ups_request.py
KDRdev: Thank you for the suggestion. Could you please explain what's the benefit of your change in this case? With `or ""` next line `if phone_number` will not be affected. Is it to make sure we don't pass False/None further?

delivery_ups_oca/__manifest__.py
KDRdev: Done, thanks!

delivery_ups_oca/models/ups_request.py
```diff
@@ -150,6 +149,11 @@ def _partner_to_shipping_data(self, partner, **kwargs):
...
+        if phone_number:
+            result["Phone"] = dict(Number=phone_number)
+        return result
```
carlos-lopez-tecnativa: I tested this in two ways:

1. Without this change, if the partner does not have a phone or mobile number and the `Phone` field is sent as empty, I get this error message.
2. With this change, the `Phone` field is not sent, but the error is the same.

Image

So, this change does not solve the problem.

carlos-lopez-tecnativa (changes requested):

delivery_ups_oca/models/ups_request.py
KDRdev: @carlos-lopez-tecnativa thank you for the feedback. Indeed this doesn't work for international shipments. Phone number must be present for such shipments and hence error you provided. That's the expected behaviour. It turns out that UPS allows omitting phone number for shipments within the same country. 

I've updated the PR so that **empty** phone number is excluded from "ShipTo" if not provided and shipment is to the same country. This way the error your provided won't appear.

delivery_ups_oca/models/ups_request.py
```diff
@@ -180,6 +160,11 @@ def _label_data(self):
             res["LabelStockSize"] = {"Height": "6", "Width": "4"}
         return res
 
+    def _is_same_origin_dest(self, ship_from, ship_to):
+        if not ship_from.country_id or not ship_to.country_id:
+            return False
+        return ship_from.country_id.id == ship_to.country_id.id
```
carlos-lopez-tecnativa, line 211: > In Germany UPS does not require recipient's phone number (while in most other countries the phone number is mandatory). This patch allows sending deliveries to German recipients which have no phone number specified.
> 
> The error message without this patch is simply: `Sending to UPS: 120209 Missing or invalid ship to phone number `

Based on this comment, which only occurs in Germany, your code considers whether the `ship_from` and `ship_to` countries are different, which I’m not totally sure is correct.

My suggestion would be to add a new boolean field to `delivery.carrier`, and when this field is enabled (False by default), apply your code and do not send the `Phone` field in `ship_to`.

What do you think about this approach?

carlos-lopez-tecnativa (changes requested):

delivery_ups_oca/models/ups_request.py
jans23: The boolean would be just another configuration setting the user needs to understand. It would require to configure two shipping methods even though they are technically the same UPS product: One shipping method applicable to the same country and with boolean true, and the other shipping method to all other applicable countries with boolean false. Consequently this would complicate the usage of such during operations. Furthermore I doubt that the code would be much simpler thus don't see an advantage of it.

From [UPS' API description](https://developer.ups.com/tag/Shipping?loc=en_DE#operation/Shipment):

> Consignee's phone Number. Required for: UPS Next Day Air® Early service, and when Ship To country or territory is different than the ShipFrom country or territory.

From my experience I can add that this also applies to UPS Standard service so the written limitation to UPS Next Day Air Early doesn't seem to be complete.

KDRdev: If the phone is missing, we will only remove it from payload if shipping within same country. If UPS returns error, because some service requires phone, user will need to add it to contact anyway, regardless of this change. So we're extending existing functionality without adding extra limitations, hence I think extra flag on `delivery.carrier` will add only add extra compexity.

KDRdev: @carlos-lopez-tecnativa could you please have another look? Thanks!

jans23:
@carlos-lopez-tecnativa Anything else to improve?

carlos-lopez-tecnativa (approved):

jans23:
@rousseldenis I think this could be merged too.

jans23:
@carlosdauden could you review this too? That would be great.

carlosdauden (approved):

jans23:
@rousseldenis this got two approvals too.

jans23:
@pedrobaeza this got approved. Could it be merged, please?

PR #1190, opened 2026-05-20
[18.0][IMP] delivery_ups_oca: add token refresh

KDRdev:


jans23:
@carlos-lopez-tecnativa it would be great if you could review this.

delivery_ups_oca/models/ups_request.py
```diff
@@ -95,16 +109,38 @@ def _process_reply(
...
+            body_data,
+        )
+
+        res = self._send_request(url, json, data, headers, method)
+
+        if res.status_code == 401:
             self._get_new_token()
-            status = self._send_request(url, json, data, headers, method)
-        status = status.json()
+            headers["Authorization"] = f"Bearer {self.token}"
```
carlos-lopez-tecnativa, line 103: Is this change the only code that actually does something?

All the remaining code is only adding debug logging. If that is correct, then to minimize the diff and avoid logging this data, please remove all code related to logging.

carlos-lopez-tecnativa (changes requested):

delivery_ups_oca/models/ups_request.py
KDRdev: You mean it's better to create a separate PR for logging? I think it's still needed, but would gladly read your opinion.

KDRdev: @carlos-lopez-tecnativa Done

carlos-lopez-tecnativa (changes requested):
I understand that you renamed the variable `status` to `res`, but this creates a much larger diff even though the actual changes are minor. I prefer to avoid renaming variables whenever possible so that the diff highlights only the meaningful changes.

jans23:
@carlos-lopez-tecnativa I renamed it as requested.

delivery_ups_oca/models/ups_request.py
```diff
@@ -66,14 +66,21 @@ def _get_new_token(self):
         status = self._send_request(
             url, data=data, headers=headers, auth=(self.client_id, self.client_secret)
         )
-        status = status.json()
+        status_json = status.json()
```
fkantelberg: This seems to be a problem. `status` was overwritten with the json and later used in `_raise_for_status` in L77 which assumes that the first argument is a dictionary. With that change you pass the response object if I see it correctly and should get an `AttributeError`. Same goes for L78. Not sure why the renaming is even necessary and if you don't want to reuse the same variable with different content it would be easier to rename `status` in L66 to `response` and keep `status = response.json()`

delivery_ups_oca/models/ups_request.py
```diff
@@ -66,14 +66,21 @@ def _get_new_token(self):
...
+
+        debug_response = (
+            status_json.copy() if isinstance(status_json, dict) else status_json
+        )
+        if isinstance(debug_response, dict) and "access_token" in debug_response:
+            debug_response["access_token"] = "***MASKED***"
```
fkantelberg: These lines are doing nothing as `debug_response` is never used afterwards (not even logged).

delivery_ups_oca/models/ups_request.py
fkantelberg: I don't understand why everything is renamed now. `status` as a variable for the json would be feasible. You can argue if you want to name the response `res` instead of reusing `status`. It just makes the diff bigger and you run into additional problems as seen on the other comment.

fkantelberg (changes requested):
This introduces more changes than required without value or good arguments and introduces an actual problem in the code because of it.

delivery_ups_oca/models/ups_request.py
jans23: This is addressed now.

delivery_ups_oca/models/ups_request.py
jans23: Changed as requested.

delivery_ups_oca/models/ups_request.py
jans23: Removed it.

jans23:
@carlos-lopez-tecnativa @fkantelberg Your feedback should be incorporated now. Please check again and approve.

delivery_ups_oca/models/ups_request.py
rousseldenis: @KDRdev IMHO, renaming variables makes unwanted noise and avoid reviewing exact changes.

delivery_ups_oca/__manifest__.py
```diff
@@ -4,12 +4,12 @@
...
     "name": "Delivery UPS OCA",
     "summary": "Integrate UPS webservice",
-    "version": "18.0.1.2.3",
+    "version": "18.0.1.2.2",
     "development_status": "Beta",
     "category": "Delivery",
     "website": "https://github.com/OCA/delivery-carrier",
     "author": """Hunki Enterprises BV, Tecnativa, ForgeFlow,
-             Odoo Community Association (OCA)""",
+        Nitrokey, Odoo Community Association (OCA)""",
```
rousseldenis: Such minor changes don't allow to be considered as author

rousseldenis (changes requested):

jans23:
@rousseldenis Now I removed all cosmetic changes as requested.

fkantelberg (approved):

rousseldenis (approved):
LGTM. Thanks!

PR #1193, opened 2026-05-22
[18.0][IMP] delivery_ups_oca: Build UPS-compatible address lines with a maximum of 35 characters

carlos-lopez-tecnativa:
TT62689

@Tecnativa @pedrobaeza @sergio-teruel @carlosdauden could you please review this?

### DIFF
--- a/delivery_ups_oca/models/ups_request.py
+++ b/delivery_ups_oca/models/ups_request.py
@@ -132,6 +132,26 @@ class UpsRequest:
             },
         }
 
+    def _build_address_lines(self, partner):
+        """Build UPS-compatible address lines with max 35 chars and up to 3 lines."""
+        full_address = f"{partner.street or ''} {partner.street2 or ''}".strip()
+        if not full_address:
+            return [""]
+        lines = []
+        remaining = full_address
+        for _i in range(3):
+            if not remaining:
+                break
+            if len(remaining) <= 35:
+                lines.append(remaining.strip())
+                break
+            split_at = remaining.rfind(" ", 0, 36)
+            if split_at <= 0:
+                split_at = 35
+            lines.append(remaining[:split_at].strip())
+            remaining = remaining[split_at:]
+        return lines
+
     def _partner_to_shipping_data(self, partner, **kwargs):
         """Return a dict describing a partner for the shipping request"""
         return dict(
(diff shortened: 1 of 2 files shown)


delivery_ups_oca/models/ups_request.py
```diff
@@ -142,7 +158,7 @@ def _partner_to_shipping_data(self, partner, **kwargs):
             Phone=dict(Number=partner.phone or partner.mobile),
             EMailAddress=partner.email,
             Address=dict(
-                AddressLine=[partner.street, partner.street2 or ""],
+                AddressLine=self._build_ups_address_lines(partner),
```
pedrobaeza: I don't think we need to complicate the code. Just:

```suggestion
                AddressLine=((self.street or "") + (self.street2 and f" {self.street2}" or ""))[:35],
```

carlos-lopez-tecnativa: But this truncates the text to 35 characters, while UPS supports up to 3 lines of 35 characters each, for a total of 105 characters split across 3 lines.
https://developer.ups.com/tag/Shipping?loc=en_DE#operation/Shipment!path=ShipmentRequest/Shipment/Shipper/Address/AddressLine&t=request

pedrobaeza: OK, understood. You are only splitting 2 lines, not 3. Isn't it?

delivery_ups_oca/models/ups_request.py
```diff
@@ -132,6 +132,22 @@ def _quant_package_data_from_picking(self, package, picking, is_package=False):
             },
         }
 
+    def _build_ups_address_lines(self, partner):
```
pedrobaeza: Shouldn't this be `_ups_build_address_lines`?

pedrobaeza: Or wait, being inside `ups_request`, it's not needed to put the prefix `ups`, as it's obvious. I thought initially that it was in `delivery.carrier`.

delivery_ups_oca/models/ups_request.py
carlos-lopez-tecnativa: Yes, up to 2 lines. Supporting 3 lines would make the code more complex, wouldn’t it? :stuck_out_tongue_winking_eye:

pedrobaeza: But one Odoo line may have a lot of words and need 3 lines of 45 to be shown. Why not joining `street` and `street2` and then split up to 3 lines of 45?

Note that I didn't know about the 3 lines thing. I only thought it was a simple 45 len limit.

carlos-lopez-tecnativa: Done

pedrobaeza (approved):
/ocabot merge patch

PR #844, opened 2024-06-28
[14.0][IMP] delivery_state: set carrier with no updatable states

manuelregidor:
Before this IMP, the delivery_state was always set to shipping_recorded_in_carrier once the expedition was created. The _update_delivery_state method in stock.picking tries to update the delivery state for those pickings in that state. Because not all the carriers APIs or carrier modules support the state update functionality, this IMP makes it optional by setting the delivery state to no_update instead of shipping_recorded_in_carrier if the new cehckbox track_carrier_state in delivery.carrier model is disabled.

T-6303

PR #1194, opened 2026-05-26
[14.0][IMP] delivery_state: Set delivery_state to no_update on invalid Pickings

raumschmiede-joshuaL:
As mentioned here: https://github.com/OCA/delivery-carrier/pull/1152#discussion_r3020087572

There is no reason to fetch a tracking state for a picking if the picking carrier provides no implementation for it.
In our system only Schenker provides it, all others not. The cron finds all of the non-Schenker pickings each run. With this change the cron will find the picking once and sets `delivery_state` to `no_update`.

One other solution would be to add a hook `_carriers_tracking_fetch_allowed` to the module. By default it returns an empty list. `delivery_schenker` overrides this and returns `super() + ["schenker"]`. The hook is used in the overridden `send_shipping` to immediately set it to `no_update` if the carrier is not in this list.

Yes, we can set `track_carrier_state` on all carriers. But the field didn't exist in 14.0 for some time until PR https://github.com/OCA/delivery-carrier/pull/844 and for carriers without a tracking state implementation this field should not have an effect anyway. Better to automatically remove the pickings from the cron run

### DIFF
--- a/delivery_state/models/stock_picking.py
+++ b/delivery_state/models/stock_picking.py
@@ -89,18 +89,21 @@ class StockPicking(models.Model):
         defined in the model as:
             <my_provider>_tracking_state_update
         It can be triggered manually or by the cron."""
-        for picking in self.filtered("carrier_id"):
+        for picking in self:
             method = "%s_tracking_state_update" % picking.delivery_type
             carrier = picking.carrier_id
 
-            if hasattr(carrier, method):
-                try:
-                    with self.env.cr.savepoint():
-                        getattr(carrier, method)(picking)
-                except Exception as e:
-                    if not self.env.context.get("lastcall"):
-                        raise
-                    picking.pod_error = str(e)
+            if not carrier or not hasattr(carrier, method):
+                picking.delivery_state = DELIVERY_STATE_NO_UPDATE
+                continue
+
+            try:
+                with self.env.cr.savepoint():
+                    getattr(carrier, method)(picking)
+            except Exception as e:
+                if not self.env.context.get("lastcall"):
+                    raise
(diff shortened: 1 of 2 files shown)


raumschmiede-joshuaL:
@mt-software-de and @hparfr, please review

delivery_state/models/stock_picking.py
```diff
@@ -70,16 +70,22 @@ def tracking_state_update(self):
...
-                try:
-                    with self.env.cr.savepoint():
-                        getattr(picking.carrier_id, method)(picking)
-                except Exception as e:
-                    if not self.env.context.get("lastcall"):
-                        raise
-                    picking.pod_error = str(e)
+            carrier = picking.carrier_id
+
+            if not carrier or not hasattr(carrier, method):
```
mt-software-de, line 96: Could you also exclude the picking already in the domain of the cron if there is no carrier set?

mt-software-de (changes requested):

delivery_state/models/stock_picking.py
raumschmiede-joshuaL: I added this because I thought the button is still displayed in the form view. But there is a domain on the view group. I can remove the check

raumschmiede-joshuaL: There was already `filtered("carrier_id")` in line 73. So I think it is better to keep `not carrier` in line 77

mt-software-de (approved):
LGTM

raumschmiede-joshuaL:
@OCA/logistics-maintainers hopefully the last one for now. Please review

hparfr (approved):
/ocabot merge major

PR #1196, opened 2026-05-29
[18.0][IMP] delivery_postlogistics: use new API endpoints (since 2023)

guewen:
Since 2023, the endpoints for the API have changed:

Deprecated:
Authorization: https://wedec.post.ch/WEDECOAuth/token
Barcode: https://wedec.post.ch/api/barcode/v1/generateAddressLabel

 

New:
Authorization: https://api.post.ch/OAuth/token
Barcode: https://dcapi.apis.post.ch/barcode/v1/generateAddressLabel

### DIFF
--- a/delivery_postlogistics/models/delivery_carrier.py
+++ b/delivery_postlogistics/models/delivery_carrier.py
@@ -19,10 +19,10 @@ class DeliveryCarrier(models.Model):
         domain=[("package_carrier_type", "=", "postlogistics")],
         default=lambda self: self._default_postlogistics_default_package_type_id(),
     )
+    # postlogistics_endpoint_url deprecated, remove in migration to next major version
     postlogistics_endpoint_url = fields.Char(
         string="Endpoint URL",
-        default="https://wedecint.post.ch/",
-        required=True,
+        required=False,
     )
     postlogistics_client_id = fields.Char(
         string="Client ID", groups="base.group_system"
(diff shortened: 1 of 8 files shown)


delivery_postlogistics/models/delivery_carrier.py
```diff
@@ -102,19 +102,6 @@ def _default_postlogistics_default_package_type_id(self):
...
-        """
-        Auto change the end point url following the environment
-        - Test: https://wedecint.post.ch/
-        - Prod: https://wedec.post.ch/
-        """
-        for carrier in self:
-            if carrier.prod_environment:
-                carrier.postlogistics_endpoint_url = "https://wedec.post.ch/"
-            else:
-                carrier.postlogistics_endpoint_url = "https://wedecint.post.ch/"
```
guewen, line 116: No int/sandbox on the new API, instead we put printPreview in the label generation (= Specimen labels)

delivery_postlogistics/models/delivery_carrier.py
```diff
@@ -19,10 +19,10 @@ class DeliveryCarrier(models.Model):
         domain=[("package_carrier_type", "=", "postlogistics")],
         default=lambda self: self._default_postlogistics_default_package_type_id(),
     )
+    # postlogistics_endpoint_url deprecated, remove in migration to next major version
     postlogistics_endpoint_url = fields.Char(
```
guewen, line 23: The endpoint is now a constant, because even if post.ch changes the api endpoint, it is likely the API will change and we'll need to update the module anyway. We can see it here in effect as the auth path has changed too. So it is overkill and not useful to have the endpoint url configurable per delivery carrier.
I kept the field only not to break databases on upgrades.

mtriki (approved):

yvaucher (approved):
Looks legit but I don't have any project using those shipping labels. code LGTM

PR #1199, opened 2026-06-01
[19.0][FIX] delivery_multi_destination: Add uninstall_hook to avoid error when accessing Delivery Methods

victoralmau:
FWP from 18.0: https://github.com/OCA/delivery-carrier/pull/1192

Add uninstall_hook to avoid error when accessing Delivery Methods

Example use case:
- Install stock + delivery_multi_destination
- Uninstall delivery_multi_destination
- An error occurs when accessing Delivery methods because the domain and context data are still defined

Please @pedrobaeza can you review it?

@Tecnativa

### DIFF
--- a/delivery_multi_destination/README.rst
+++ b/delivery_multi_destination/README.rst
@@ -68,8 +68,8 @@ Usage
 Known issues / Roadmap
 ======================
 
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):
/ocabot merge patch

PR #1203, opened 2026-06-05
[18.0][FIX] partner_delivery_zone: Change delivery_zone_id readonly domain

Andrii9090-tecnativa:
These changes allows to update the `delivery_zone_id` field when the sales order is in the "sale" status

@sergio-teruel @CarlosRoca13 ping

@Tecnativa
TT62890

### DIFF
--- a/partner_delivery_zone/views/sale_order_view.xml
+++ b/partner_delivery_zone/views/sale_order_view.xml
@@ -8,7 +8,7 @@
         <field name="inherit_id" ref="sale.view_order_form" />
         <field name="arch" type="xml">
             <xpath expr="//field[@name='payment_term_id']" position="after">
-                <field name="delivery_zone_id" readonly="state in ['sale', 'cancel']" />
+                <field name="delivery_zone_id" readonly="state == 'cancel'" />
             </xpath>
         </field>
     </record>


sergio-teruel (changes requested):
Please!!!
Respect the original module:
https://github.com/OCA/delivery-carrier/blob/15.0/partner_delivery_zone/views/sale_order_view.xml#L13

Andrii9090-tecnativa:
This [commit](https://github.com/odoo/odoo/commit/17bece3e797913bcba8dd7e07fc8541c0a45e3f7#diff-05c263ea6063dd0045f2cd5be1888d608448ef8ff11f23371616334273c133e5l) removed the `done` state

sergio-teruel:
readonly="state == 'cancel'" <> attrs="{'readonly':[('state','in', ['done', 'cancel'])]}"

Andrii9090-tecnativa:
> readonly="state == 'cancel'" <> attrs="{'readonly':[('state','in', ['done', 'cancel'])]}"

If the `done` state does not exist in v18, we need to use only the `cancel` state for the `readonly` attribute. Now it has only one state, so we don’t need to use `in`; I use `==` directly.

sergio-teruel:
oks..

sergio-teruel (approved):

CarlosRoca13 (approved):

PR #931, opened 2025-01-15
[17.0][MIG] delivery_free_fee_removal: Migration to 17.0

juancarlosonate-tecnativa:
Migration to 17.0

TT52318 @Tecnativa

@pilarvargas-tecnativa @pedrobaeza can you check? thanks!

PR #1212, opened 2026-06-30
[18.0][FIX] delivery_free_fee_removal: Allow changing the shipping method if the order is confirmed

victoralmau:
Allow changing the shipping method if the order is confirmed

Example use case:
- Create an order
- Add a shipping method with a price of 0
- Confirm the order
- Try to change the shipping method

Similar to https://github.com/OCA/delivery-carrier/pull/931/changes/22652b989457f3ffc5639ddd3984034aad8d5191

When the migration to v18 took place https://github.com/OCA/delivery-carrier/pull/915, v17 wasn't available yet, and the approach was changed during the migration https://github.com/OCA/delivery-carrier/pull/931.

Please @pedrobaeza and @juancarlosonate-tecnativa can you review it?

@Tecnativa TT63520


pedrobaeza (commented):
Instead of this, that seems a hack over another hack, why don't you change the qty to invoice compute method for the is_free_delivery lines to return 0, and remove `action_confirm` and this new override?

pedrobaeza (commented):
Please remember to add comments on each override about what is the override doing.

pedrobaeza (approved):

juancarlosonate-tecnativa (approved):

christian-ramos-tecnativa (approved):

PR #1215, opened 2026-07-06
[18.0][FIX] delivery_ups_oca: avoid invalid UPS ShipTo tax id

sergio-teruel:
Avoid sending invalid TaxIdentificationNumber values in UPS shipment requests.
UPS Shipping API defines ShipTo.TaxIdentificationNumber as deprecated, not required, and limited to 15 characters. The module was always sending partner.vat, which can break shipment creation when the partner tax ID is longer, for example some country-specific tax identifiers.

This change:
- Sanitizes the tax ID by removing whitespace.
- Sends TaxIdentificationNumber only when it is present and within UPS’ 15-character limit.
- Omits the field otherwise, instead of blocking shipment creation.
- Adds tests for oversized tax IDs and whitespace sanitization.

cc @Tecnativa TT63497

ping @carlos-lopez-tecnativa @carlosdauden


carlosdauden (approved):

delivery_ups_oca/tests/test_delivery_ups.py
```diff
@@ -884,3 +884,23 @@ def test_ups_address_lines(self):
...
         self.assertEqual(address_lines[0], "12345 Short Street Name Suite 678")
+
+    def test_ups_partner_to_shipping_data_tax_identification_number_limit(self):
+        """Do not send TaxIdentificationNumber values rejected by UPS length limit."""
+        ups_request = UpsRequest(self.carrier)
+        self.partner.with_context(no_vat_validation=True).write({"vat": "X" * 16})
+
+        shipping_data = ups_request._partner_to_shipping_data(self.partner)
+
+        self.assertNotIn("TaxIdentificationNumber", shipping_data)
```
carlos-lopez-tecnativa, line 893: Please remove the empty lines within the function.
```suggestion
        self.partner.with_context(no_vat_validation=True).write({"vat": "X" * 16})
        shipping_data = ups_request._partner_to_shipping_data(self.partner)
        self.assertNotIn("TaxIdentificationNumber", shipping_data)
```

delivery_ups_oca/tests/test_delivery_ups.py
```diff
@@ -884,3 +884,23 @@ def test_ups_address_lines(self):
...
+    def test_ups_partner_to_shipping_data_tax_identification_number_sanitized(self):
+        """Send valid-length tax ids without whitespace."""
+        ups_request = UpsRequest(self.carrier)
+        self.partner.with_context(no_vat_validation=True).write(
+            {"vat": "IT 12345678901"}
+        )
+
+        shipping_data = ups_request._partner_to_shipping_data(self.partner)
+
+        self.assertEqual(shipping_data["TaxIdentificationNumber"], "IT12345678901")
```
carlos-lopez-tecnativa, line 902:
```suggestion
        self.partner.with_context(no_vat_validation=True).write(
            {"vat": "IT 12345678901"}
        )
        shipping_data = ups_request._partner_to_shipping_data(self.partner)
        self.assertEqual(shipping_data["TaxIdentificationNumber"], "IT12345678901")
```

delivery_ups_oca/models/ups_request.py
```diff
@@ -153,13 +154,20 @@ def _build_address_lines(self, partner):
             remaining = remaining[split_at:]
         return lines
 
+    def _get_tax_identification_number(self, partner):
+        vat = "".join((partner.vat or "").split())
```
carlos-lopez-tecnativa, line 159: I'm not sure why you need to convert this to a list.

If you call `len("string")`, it already returns the number of characters in the string. There's no need to convert it to a list first.

If the intention is to remove the spaces, a more Pythonic approach would be to use `replace()`. At least to me, it's clearer and easier to read.

Or am I missing something? Please clarify.

```suggestion
        vat = (partner.vat or "").replace(" ", ""))
```

carlos-lopez-tecnativa (changes requested):

delivery_ups_oca/models/ups_request.py
sergio-teruel: The intention here is to normalize the tax id before checking UPS’ max length, not to compute the length on a list.
split() without arguments removes all whitespace sequences, while replace(" ", "") only removes literal spaces. This avoids sending or rejecting formatted values such as IT 12345678901.

sergio-teruel:
@carlos-lopez-tecnativa Changes done!!

delivery_ups_oca/models/ups_request.py
carlos-lopez-tecnativa: Thanks for the clarification.

carlos-lopez-tecnativa (approved):
LGTM

PR #950, opened 2025-02-06
[17.0][MIG] delivery_carrier_max_weight_constraint: Migration to 17.0

BhaveshHeliconia:

PR #1217, opened 2026-07-13
[17.0][FIX] delivery_sendcloud_oca: fix test failure due to new Sendcloud API error format

BhaveshHeliconia:
This PR fixes the failure in test_10_auto_create_invoice, which incorrectly asserts Sendcloud API authentication errors.
The issue is causing the CI failures in these PRs: https://github.com/OCA/delivery-carrier/pull/959, https://github.com/OCA/delivery-carrier/pull/951, https://github.com/OCA/delivery-carrier/pull/950
### What happened:
The live Sendcloud API recently updated the error response structure when invalid credentials are used, returning a multiline message. Since the test assertions were using regex matching without the DOTALL flag (`(?s)`), they failed to match the multiline output.
### Fix:
Updated the `assertRaisesRegex` patterns in `test_delivery_sendcloud_oca.py` to use `(?s)` so they can match across newlines and support the new API error format.


rousseldenis (approved):
LGTM

hparfr (approved):
/ocabot merge patch

PR #1219, opened 2026-07-23
[18.0][FIX] delivery_ups_oca: Hide unavailable shipping methods

jans23:


jans23:
@carlos-lopez-tecnativa would be great if you could review this. Thanks.

jans23:
@carlosdauden could you review this, please?

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -121,16 +122,26 @@ def _ups_get_response_price(self, total_charges, currency, company):
...
+                "error_message": False,
+                "warning_message": False,
+            }
+        except UserError as e:
+            # During rate shopping (checkout), return failure instead of
+            # raising thus gracefully hide unavailable shipping methods.
+            return {
+                "success": False,
+                "price": 0.0,
+                "error_message": str(e),
```
KDRdev, line 146: The solution looks fine, however the error here won't be visible unless using "Get Rate" button on Sale Order in backend. Maybe it would be better to log this in case there's an error during checkout on website?

jans23: The idea of this change is to gracefully fail for shipping methods which are not available to individual destinations. Therefore a logged error would unnecessarily irritate.

KDRdev: I agree, but you shouldn't always log this and use at least _logger.debug (less irritating) in case there is some problem with shipping method unavailability. UserError is raised here not only when destination is unavailable, but for any API error. But let's wait for another reviewer's comment. Maybe it's ok as it is.

carlos-lopez-tecnativa (approved):

delivery_ups_oca/models/delivery_carrier.py
jans23: I added it to the debug log as suggested.

KDRdev: Great, thank you!

KDRdev (approved):

jans23:
@rousseldenis This has been approved. Please check if you want to merge it.

rousseldenis (approved):
LGTM

PR #1236, opened 2026-08-12
[18.0][FIX] delivery_ups_oca: guard package weight against zero packages

jans23:
When a picking is shipped without using packages from the picking (ups_use_packages_from_picking disabled) and its number_of_packages is 0, computing the per-package weight as shipping_weight / number_of_packages raised a ZeroDivisionError. Fall back to the total shipping_weight in that case so the shipment request can still be built.

### DIFF
--- a/delivery_ups_oca/models/ups_request.py
+++ b/delivery_ups_oca/models/ups_request.py
@@ -236,7 +236,7 @@ class UpsRequest:
                 self.default_packaging_id, picking, False
             )
             package_weight = round(
-                (picking.shipping_weight / picking.number_of_packages), 2
+                (picking.shipping_weight / (picking.number_of_packages or 1)), 2
             )
             for i in range(0, picking.number_of_packages):
                 package_item = package_info.copy()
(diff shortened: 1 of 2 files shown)


jans23:
@carlos-lopez-tecnativa @carlosdauden could you review this small fix, please?

delivery_ups_oca/models/ups_request.py
```diff
@@ -235,9 +235,13 @@ def _prepare_create_shipping(self, picking):
...
-            package_weight = round(
-                (picking.shipping_weight / picking.number_of_packages), 2
-            )
+            if picking.number_of_packages > 0:
+                package_weight = round(
+                    (picking.shipping_weight / picking.number_of_packages), 2
+                )
+            else:
+                package_weight = picking.shipping_weight
+
```
carlos-lopez-tecnativa: Why not simply:
```suggestion
              package_weight = round(
                  (picking.shipping_weight / (picking.number_of_packages or 1)), 2
              )
```

carlos-lopez-tecnativa (changes requested):
I confirm the error, and the PR solves it.
Image

jans23:
@carlos-lopez-tecnativa I applied your change as suggested.

carlos-lopez-tecnativa (approved):
LGTM ping @pedrobaeza

KDRdev (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1240, opened 2026-08-18
[19.0][MIG] delivery_carrier_max_weight_constraint: Migration to 19.0

prayagrk:
delivery_carrier_max_weight_constraint: Migration to 19.0


prayagrk:
Hi @BhaveshHeliconia , @LauraCForgeFlow, @mymage, Can you please review this PR?

LauraCForgeFlow (approved):
Tested on runboat LGTM!

BhaveshHeliconia (approved):
LGTM!

BhaveshHeliconia:
@prayagrk could you please squash the administrative and bot/weblate commits?

prayagrk:
> @prayagrk could you please squash the administrative and bot/weblate commits?

@BhaveshHeliconia , I’ve just updated the commit history.

BhaveshHeliconia:
@prayagrk you should only merge translation commits coming from the same author, not all the translations commits together.
https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate

delivery_carrier_max_weight_constraint/models/stock_package_type.py
```diff
@@ -0,0 +1,25 @@
...
+    is_strict_weight = fields.Boolean(
+        string="Overweight forbidden",
+        help="The maximum weight of the package can't be exceeded.",
+        compute="_compute_is_strict_weight",
+        store=True,
+        readonly=False,
+    )
+
+    @api.depends("company_id.delivery_carrier_strict_weight_package")
+    def _compute_is_strict_weight(self):
```
LINTOANTONY007, line 19: Do we need to add `@api.depends_context('company')` ? I am not sure!

mmircoli-nexapp: Regarding @LINTOANTONY007's question about @api.depends_context('company'):
The comment is not applicable as the code stands. The compute method only reads rec.company_id.delivery_carrier_strict_weight_package — it uses the company_id field stored on the record itself, not self.env.company. When company_id is falsy, it defaults to False. Since there's no self.env.company fallback, @api.depends_context('company') is not needed.
That said, one could argue that when a stock.package.type has no company_id (shared across companies), the compute should fall back to the current user's company setting, like:
rec.is_strict_weight = (
    (rec.company_id or self.env.company).delivery_carrier_strict_weight_package
)
In that case, @api.depends_context('company') would indeed be required, since the computed value would then depend on which company the user is logged into. But that would be a functional change / improvement, not a bug in the current migration.
LGTM for merge as-is. The @api.depends_context suggestion would only be relevant if we change the fallback behavior for company-less package types.

mmircoli-nexapp (approved):

PR #1250, opened 2026-08-29
[18.0][IMP] delivery_ups_oca: add debug logging

jans23:
Add debug-level logging around the UPS token exchange and API requests including token masking.


jans23:
@carlos-lopez-tecnativa this is a small but useful feature. Please have a look.

PR #1224, opened 2026-08-04
[18.0][IMP] delivery_ups_oca: negotiated rates

jans23:

PR #1257, opened 2026-09-01
[18.0][FIX] delivery_ups_oca: convert rate price from response currency, not the reverse

juancarlosonate-tecnativa:
The rate tests added in #1224 fail on any database where the order currency isn't EUR: the mocked UPS response is in EUR, so the price gets converted and the hardcoded asserts break (AssertionError: 20.96 != 16.33). OCA CI doesn't catch it because EUR is inactive on a bare install.

While debugging it we found _ups_get_response_price() also converts in the wrong direction (order to response currency instead of the reverse).

This PR fixes the conversion direction following the delivery_easypost_oca pattern, and pins the order currency in those tests so they don't depend on the database currencies.

@tecnativa @carlos-lopez-tecnativa @pedrobaeza @jans23

### DIFF
--- a/delivery_ups_oca/models/delivery_carrier.py
+++ b/delivery_ups_oca/models/delivery_carrier.py
@@ -117,13 +117,11 @@ class DeliveryCarrier(models.Model):
         """We need to convert the price if the currency is different."""
         price = float(total_charges["MonetaryValue"])
         if total_charges["CurrencyCode"] != currency.name:
-            price = currency._convert(
-                price,
-                self.env["res.currency"].search(
-                    [("name", "=", total_charges["CurrencyCode"])]
-                ),
-                company,
-                fields.Date.today(),
+            response_currency = self.env["res.currency"].search(
+                [("name", "=", total_charges["CurrencyCode"])], limit=1
+            )
+            price = response_currency._convert(
+                price, currency, company, fields.Date.context_today(self)
             )
         return price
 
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

jans23 (approved):

delivery_ups_oca/models/delivery_carrier.py
```diff
@@ -117,13 +117,11 @@ def _ups_get_response_price(self, total_charges, currency, company):
...
-                self.env["res.currency"].search(
-                    [("name", "=", total_charges["CurrencyCode"])]
-                ),
-                company,
-                fields.Date.today(),
+            response_currency = self.env["res.currency"].search(
+                [("name", "=", total_charges["CurrencyCode"])], limit=1
+            )
+            price = response_currency._convert(
+                price, currency, company, fields.Date.today()
```
carlos-lopez-tecnativa: Please use `context_today` instead of `today` to retrieve the current date in the user's timezone. In countries with negative timezones, such as those in LATAM, today can return the next day if this is executed at night.
```suggestion
                price, currency, company, fields.Date.context_today(self)
```

carlos-lopez-tecnativa (changes requested):

delivery_ups_oca/models/delivery_carrier.py
juancarlosonate-tecnativa: done, thx

carlos-lopez-tecnativa (approved):

PR #1262, opened 2026-09-16
[18.0][FIX] delivery_ups_oca: update Rate, Ship API to v2

jans23:
The current API v1 is deprecated. This PR updates to the latest v2 API.


jans23:
@carlos-lopez-tecnativa it would be great to have the API updated before v1 gets disabled (I don't know when). Please have a look.

